PR #15, opened 2016-12-01
A few API changes based on comments and design work.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/15)

### DIFF
--- a/api/v1/report.proto
+++ b/api/v1/report.proto
@@ -31,11 +31,14 @@ message ReportRequest {
   // within the current stream.
   map<string, string> facts = 2;
 
+  // The monitored resource type associated with this report.
+  string monitored_resource_descriptor_name = 3
+
   // Metric values being reported
-  repeated MetricValue metric_values = 3;
+  repeated MetricValue metric_values = 4;
 
   // Log entries being reported
-  repeated LogEntry log_entries = 4;
+  repeated LogEntry log_entries = 5;
 }
 
 // Represents a single metric value.
@@ -147,7 +150,4 @@ message ReportResponse {
 
   // Indicates whether the report was processed or not
   google.rpc.Status result = 2;
-
-  // the known facts that were used to evaluate the request
-  repeated string useful_facts = 3;
 }
--- a/api/v1/check.proto
+++ b/api/v1/check.proto
@@ -18,7 +18,7 @@ package istio.mixer.v1;
 
 import "google/rpc/status.proto";
 
-// Used to verify a preconditions before performing an action.
+// Used to verify preconditions before performing an action.
 message CheckRequest {
   // Uniquely identifies this request, used to match to responses
   int64 request_id = 1;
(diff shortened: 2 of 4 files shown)


config/v1/metric_descriptor.proto
```diff
@@ -0,0 +1,171 @@
...
+// A given metric is described by a set of labels. These labels represent the
+// different dimensions to associate with the metric. A given metric holds a
+// unique value for potentially any combination of these labels.
+//
+// Metric values can be integers, floating point numbers, strings, or booleans.
+//
+// Here's an example metric descriptor defined in a YAML configuration
+// file:
+//
+//    metrics:
```
mandarjog: I am thinking of a metric definition from the standpoint of entire istio, not just mixer.

```
- name: request_size
       elementary_measurement: request.size 
       metric_kind: DELTA
       value_type: INT64
       labels:
      - key: consumer_id
      - key: status_code
      - key: service_id
- name: bytes_transferred
       elementary_measurement: request.size + response.size
       metric_kind: DELTA
       value_type: INT64
       labels:
      - key: consumer_id
      - key: status_code
      - key: service_id
```
Disregarding proxy side aggregation, here is how this information is used jointly by mixer and the proxy.

proxy needs to know
1. The union of all labels to send across all metrics  [consumer_id, status_code, service_id]
2. The union of elementary_measurements needed by the metrics  [request.size, response.size]

Report call will contain all the above information. At this point mixer is able to assemble the actual metrics and add labels. 

.... However the same declarative metric definition can be used to fully compute the metric on proxy side.

mandarjog: + @qiwzhang this is what I was talking about this morning.

api/v1/report.proto
```diff
@@ -31,11 +31,14 @@ message ReportRequest {
   // within the current stream.
   map<string, string> facts = 2;
 
+  // The monitored resource type associated with this report.
+  string monitored_resource_descriptor_name = 3
+
   // Metric values being reported
-  repeated MetricValue metric_values = 3;
+  repeated MetricValue metric_values = 4;
```
qiwzhang: Does this imply that the proxy will populate metrics and log_entries and send to mixer?  I think a better design is: the proxy just send the primary data, such as size, duration, ip, and mixer will populate the metrics and log-entries based on its backend plugins.

config/v1/metric_descriptor.proto
qiwzhang: This looks good. 

But I prefer only mixer needs to know about this, no proxy.  The proxy just collects the facts (such as size, duration, ip) and send to mixer.

geeknoid: Mandar, what is "request.size" in your world? A fact, or a label produced by mapping facts, or something new?

In the API we have now, what I think you are trying to capture via elementary_measurement is handled imperatively by the caller. The MetricDescriptor type is really intended to describe the characteristic of a single metric value, not how these values are acquired. So the MetricDescriptor says "I want to hold integer values" and "I want to have different physical counters for each label I define". When someone calls Report, they provide the metric descriptor name, and a metric value (an integer). The mixer then derives the needed labels to properly 'route' the value, and forwards the result to the adapters.

I don't want to bind the definition of the metric with how the metric is acquired, those are separate concerns.

PR #19, opened 2016-12-02
Adapt the logging interface to work properly for our fact-based model.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/19)

### DIFF
--- a/api/v1/report.proto
+++ b/api/v1/report.proto
@@ -16,6 +16,8 @@ syntax = "proto3";
 
 package istio.mixer.v1;
 
+option go_package="istio.io/mixer/api/v1;mixer";
+
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/struct.proto";
 import "google/protobuf/any.proto";
--- a/api/v1/report.proto
+++ b/api/v1/report.proto
@@ -116,33 +116,32 @@ message LogEntry {
     EMERGENCY = 800;
   }
 
-  // Destination log for this entry
-  string log_name = 10;
+  // The name of the LogEntryDescriptor providing the schema for this log entry.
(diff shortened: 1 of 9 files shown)


api/v1/report.proto
```diff
@@ -114,33 +114,32 @@ message LogEntry {
     EMERGENCY = 800;
   }
 
+  // The name of the LogEntryDescriptor providing the schema for this log entry.
+  string descriptor_name = 1;
+
   // Destination log for this entry
```
douglas-reid: not relevant to this review, but I wonder if we need to clarify terminology here a bit. something like:

"Destination collection for this log entry".

api/v1/report.proto
```diff
@@ -114,33 +114,32 @@ message LogEntry {
     EMERGENCY = 800;
   }
 
+  // The name of the LogEntryDescriptor providing the schema for this log entry.
+  string descriptor_name = 1;
```
douglas-reid: checking: are these names fully qualified?  in other words, is this expected to be "service.example.io/logEntries/default" or just "default" (and proxy/mixer will be able to find the right descriptor)?

geeknoid: Depends on the case.

The metrics, quotas, and logs that the proxy will use will require well-known descriptor names. Those probably want to be qualified with something like proxy/<name>. For those descriptors for things that only the service itself will reference, such as for resource quotas or principals for permission checks, then those names are entirely under operator control and so don't need to be qualified.

api/v1/report.proto
geeknoid: Will do

geeknoid:
Updated PTAL.

PR #20, opened 2016-12-02
Add go_package option to protos.

douglas-reid:
This refactors the golang code generation bits for the
mixer protos by adding a package option to the protos. Generated
golang code is also now generated in the vendor/ directory,
instead of in a special go/ directory in the api/v1/ directory.

This should make importing and use of the protos a little less
cumbersome in the codebase in general.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/20)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,14 +20,15 @@ test: test_server
 ## API Targets
 
 PROTOC = bin/protoc.$(shell uname)
-API_OUTDIR_GO = api/v1/go
+VENDOR = vendor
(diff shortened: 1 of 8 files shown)


geeknoid:
I don't think putting build outputs into the vendor directory is the right approach. This breaks the expectations for what vendor is for.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/20#-:-KXzi-yKFditrFLGQ85g:bhcb082)*

mandarjog:
The alternative is to put it in mixer/API/v1 itself, and forgo the go directory.

PR #28, opened 2016-12-05
Suggest to use singular names for directory naming

wora:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/28)


wora:
I assume someone who has write access needs to merge this PR. Thanks.

PR #30, opened 2016-12-06
Change request_id to index

smawson:
request_id sounds like a unique id for a request, and it isn't, it is the index within the stream of that object. Attempting to clarify.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/30)

### DIFF
--- a/api/v1/check.proto
+++ b/api/v1/check.proto
@@ -22,8 +22,9 @@ import "google/rpc/status.proto";
 
 // Used to verify preconditions before performing an action.
 message CheckRequest {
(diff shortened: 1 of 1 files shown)


geeknoid:
I'm having trouble submitting my feedback, Github keeps giving me errors. So apologies if this ends up showing up multiple times.

- How about request_index for additional clarity about what index this is about?

- Can you also please change the other protos?

Thanks.

smawson:
Let me drop this and create one to rename to request_index in all the places, and do it from my fork rather than a branch of the master.

PR #34, opened 2016-12-06
Basic proposal for well-known and strongly typed facts.

smawson:
- Renames fact to attribute, and makes attributes typed (using metric value types)
- Adds well known types to an Attributes object
- Replaces map<string, string> facts with Attributes object
- Removes metric values and log entries from report.
- Removes aggregation, I'd like to add aggregation generically for attributes, rather than just for metric values.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/34)

### DIFF
--- a/api/v1/attributes.proto
+++ b/api/v1/attributes.proto
@@ -18,45 +18,24 @@ package istio.mixer.v1;
 
 option go_package="istio.io/mixer/api/v1;mixerpb";
 
-// Contains a set of attributes related to an event.
-message Attributes {
-  // Attributes related to the source of an event.
+// A set of attributes reported over a period of time.
+// An instant can be represented by only reporting an end
+// time and no start time.
+message AttributeSpan {
+  // The start of the span being reported.
+  google.protobuf.Timestamp start_time = 1;
 
-  uint32 source_ip = 1;
-  uint32 source_port = 2;
-  string source_name = 3;
-  string source_instance = 4;
-  string source_namespace = 5;
-  string source_user = 6;
-  // ...
+  // The end of the span being reported.
+  google.protobuf.Timestamp end_time = 2;
 
-  // Attributes related to the target of an event.
-  uint32 target_ip = 101;
-  uint32 target_port = 102;
-  string target_service = 103;
-  string target_name = 104;
-  string target_instance = 105;
-  string target_namespace = 106;
-  string target_user = 107;
-  // ...
-
-  // Attributes related to the origin of an event,
-  // if the source of the event is proxying it.
-  uint32 origin_ip = 201;
-  uint32 origin_port = 202;
-  string origin_name = 203;
-  string origin_instance = 204;
-  string origin_namespace = 205;
-  string origin_user = 206;
-  // ...
-
-  // Non predefined attributes go here.
-  repeated Attribute attributes = 1000;
+  // The named values being reported in the span.
+  repeated Attribute attributes = 3;
 }
 
 // An individual attribute, which has a name and a value.
 message Attribute {
-  // The name of the attribute.
+  // The name of the attribute, which is dot-delimited
+  // ascii string, for example "source.device_id"
   string name = 1;
 
   // The value of the attribute.
--- a/api/v1/attributes.proto
+++ b/api/v1/attributes.proto
@@ -1,0 +1,69 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.mixer.v1;
+
+option go_package="istio.io/mixer/api/v1;mixerpb";
+
+// Contains a set of attributes related to an event.
+message Attributes {
+  // Attributes related to the source of an event.
+
+  uint32 source_ip = 1;
+  uint32 source_port = 2;
+  string source_name = 3;
+  string source_instance = 4;
+  string source_namespace = 5;
+  string source_user = 6;
+  // ...
+
+  // Attributes related to the target of an event.
+  uint32 target_ip = 101;
+  uint32 target_port = 102;
+  string target_service = 103;
+  string target_name = 104;
+  string target_instance = 105;
+  string target_namespace = 106;
+  string target_user = 107;
+  // ...
+
+  // Attributes related to the origin of an event,
+  // if the source of the event is proxying it.
+  uint32 origin_ip = 201;
+  uint32 origin_port = 202;
+  string origin_name = 203;
+  string origin_instance = 204;
+  string origin_namespace = 205;
+  string origin_user = 206;
+  // ...
+
+  // Non predefined attributes go here.
+  repeated Attribute attributes = 1000;
+}
+
+// An individual attribute, which has a name and a value.
+message Attribute {
+  // The name of the attribute.
+  string name = 1;
+
+  // The value of the attribute.
+  oneof value {
+    string string_value = 2;
+    bool bool_value = 3;
+    double double_value = 4;
+    int64 int64_value = 5;
+  }
+}
--- a/api/v1/report.proto
+++ b/api/v1/report.proto
@@ -26,10 +26,9 @@ message ReportRequest {
   // Uniquely identifies this request, used to match to responses
   int64 request_id = 1;
 
-  // The set of attributes associated with this request. This value represents
-  // a delta relative to the attributes currently known to the mixer within
-  // the current stream.
-  Attributes attributes = 2;
+  // Spans of attributes being reported to the mixer. Only attributes that
+  // have changed in value will be reported within a span.
+  repeated AttributeSpan spans = 2;
 }
 
 message ReportResponse {
--- a/api/v1/report.proto
+++ b/api/v1/report.proto
@@ -18,9 +18,7 @@ package istio.mixer.v1;
 
 option go_package="istio.io/mixer/api/v1;mixerpb";
 
-import "google/protobuf/timestamp.proto";
-import "google/protobuf/struct.proto";
-import "google/protobuf/any.proto";
+import "attributes.proto";
 import "google/rpc/status.proto";
 
 // Used to report telemetry after performing an action.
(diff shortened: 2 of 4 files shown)


api/v1/attributes.proto
```diff
@@ -0,0 +1,69 @@
...
+
+package istio.mixer.v1;
+
+option go_package="istio.io/mixer/api/v1;mixerpb";
+
+// Contains a set of attributes related to an event.
+message Attributes {
+  // Attributes related to the source of an event.
+
+  uint32 source_ip = 1;
```
louiscryan: Add TODO to cover ipv6

api/v1/report.proto
```diff
@@ -18,131 +18,18 @@ package istio.mixer.v1;
...
-
-    // The log entry payload, represented as a Unicode string (UTF-8).
-    string text_payload = 6;
-
-    // The log entry payload, represented as a structure that
-    // is expressed as a JSON object.
-    google.protobuf.Struct struct_payload = 7;
-  }
+  // The set of attributes associated with this request. This value represents
+  // a delta relative to the attributes currently known to the mixer within
```
mandarjog: Are we still insisting that this be delta?

api/v1/attributes.proto
```diff
@@ -0,0 +1,69 @@
...
+  string origin_user = 206;
+  // ...
+
+  // Non predefined attributes go here.
+  repeated Attribute attributes = 1000;
+}
+
+// An individual attribute, which has a name and a value.
+message Attribute {
+  // The name of the attribute.
```
louiscryan: Clarify naming conv

api/v1/attributes.proto
```diff
@@ -0,0 +1,69 @@
...
+  // if the source of the event is proxying it.
+  uint32 origin_ip = 201;
+  uint32 origin_port = 202;
+  string origin_name = 203;
+  string origin_instance = 204;
+  string origin_namespace = 205;
+  string origin_user = 206;
+  // ...
+
+  // Non predefined attributes go here.
```
mandarjog: Should we define standard http headers as attributes here?
perhaps a proto already exists that defines all standard http headers.

api/v1/attributes.proto
```diff
@@ -0,0 +1,69 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.mixer.v1;
+
+option go_package="istio.io/mixer/api/v1;mixerpb";
+
+// Contains a set of attributes related to an event.
+message Attributes {
```
geeknoid: I believe this is a premature and undesirable optimization. It will cause the implementation of the mixer to have a whole bunch of conditional code to deal with the fields as well as the map in a generic way. We should not be doing this.

api/v1/report.proto
```diff
@@ -18,131 +18,18 @@ package istio.mixer.v1;
...
-// Represents a single metric value.
-message MetricValue {
-  // Name of the MetricDescriptor providing the schema for the metric.
-  string descriptor_name = 1;
-
-  // The start of the time period over which this metric value's measurement
-  // applies. The time period has different semantics for different metric
-  // kinds (cumulative, delta, and gauge). See the documentation for the
-  // MetricDescriptor for more information.
-  google.protobuf.Timestamp start_time = 3;
```
louiscryan: Interval needs to be retained

api/v1/report.proto
```diff
@@ -18,131 +18,18 @@ package istio.mixer.v1;
...
-import "google/protobuf/any.proto";
+import "attributes.proto";
 import "google/rpc/status.proto";
 
 // Used to report telemetry after performing an action.
 message ReportRequest {
   // Uniquely identifies this request, used to match to responses
   int64 request_id = 1;
 
-  // The set of discovered facts associated with this request. This value
```
geeknoid: I don't believe this is the right direction at all. This completely any pretense that the report method can be used by services, it becomes only useful for the proxy which I think we would regret. Likely that we would end up implementing a separate NonProxyReport call in the future which would be sad.

api/v1/attributes.proto
louiscryan: Martin - what do you think about solving the generic handling issue with annotations? E.g.

option (istio.attr) = { name: "source.ip", description: "IPv4 address of the source of the traffic"}
uint32 source_ip = 1;

We're almost certainly going to want some extended metadata for well-known attributes anyway.

I agree this is an early optimization but I strongly suspect well end up having to optimize to eliminate the overhead of the 'name' string field on Attribute pretty quickly. I am fine deferring this until we have some numbers...

api/v1/attributes.proto
louiscryan: repeated string would be fine for headers where name and value are interleaved.

api/v1/attributes.proto
louiscryan: The alternative way to handle this would be to define an enum for well-known types have a oneof (name | well-known-enum value)

api/v1/report.proto
smawson: I'm fine either way, but wasn't changing that part of it, since I didn't remember if we came to a consensus. Others thoughts on whether to keep the delta or discard for now?

api/v1/attributes.proto
smawson: I think the mixer code shouldn't be too ugly for this, it can even use the protobuf descriptor and do it fully generically, if the field names follow a standard pattern. Basically take field name, replace '_' with '.', and stick those all in a map, then all the rest of the code just uses that. No need for annotations, unless we need the description in code somehow.

This is meant to serve two purposes:
1. optimization
1. provide a set of well-known, well-typed attributes that proxies and mixers can agree on.

I don't think it premature for the second, but agree it may be for the first, which should make us careful to only add fields we think will live for 5+ years.

api/v1/report.proto
smawson: We aren't supporting aggregation for now, so interval isn't needed. Let's add it back later in a generic way across all attributes?

Although well known attributes that have aggregation intervals would be troublesome...

An alternative to that would be to define:
StringAttribute
BoolAttribute
DoubleAttribute
IntAttribute

Or something akin to that, add the intervals there, and then we can define the well known attributes based on those types, rather than using the raw fields...

Or we go to the suggestion of using enum or string for name, and just have a repeated thing.

Thoughts?

api/v1/attributes.proto
```diff
@@ -0,0 +1,69 @@
...
+  // Attributes related to the origin of an event,
+  // if the source of the event is proxying it.
+  uint32 origin_ip = 201;
+  uint32 origin_port = 202;
+  string origin_name = 203;
+  string origin_instance = 204;
+  string origin_namespace = 205;
+  string origin_user = 206;
+  // ...
+
```
smawson: Eventually absolutely, I intentionally did not go whole hog so we could add them piece at a time :)

We can always start using them ahead of time without this since it is just an optimization.

api/v1/report.proto
smawson: Why would we need to do that?

Can you explain a use case where you would need to define a separate method for non-proxy callers?

The direction we're going with the api and config is that the caller should not need a lot of custom logic, and instead mixer should do the heavy lifting, so I'm curious why a more complex API is necessary.

api/v1/attributes.proto
geeknoid: I don't actually believe the premise underlying this optimization for several reasons:

- We have an existence proof with ESF & Chemist within GCP (for the non-Googlers, ESF is the logical equivalent to the Istio proxy and Chemist is the logical equivalent to the Istio mixer). In the ESF/Chemist world, we use long-form labels with every request. Check calls typically have 2-3 labels, Report calls have ~10 labels. Labels have pretty long scoped names, and often long values too. There is no delta functionality provided so labels are sent anew for every single request. This demonstrates that the overall model is definitely usable at scale in its current form.

- Communication with the various backends we interact with for logging and monitoring use long-form labels. Even if we optimize the proxy->mixer protocol, the mixer->backend communication will still be on the slow fully-expanded plan.

- With delta facts, only a minimal set of facts need to be transmitted from proxy to mixer. I can imagine that perhaps the 'fact context' model that Mandar suggested could further improve the efficiency of the deltas.

- Proxy-side aggregation, batching, and caching means that QPS between the proxy and the mixer is typically considerably lower than the inbound QPS to the proxy. So the communication overhead is amortized over many incoming calls to the proxy.

So I definitely would rather we stay on the current plan until we demonstrate there is value in something different.

api/v1/report.proto
geeknoid: I don't think we can tolerate any design that doesn't allow aggregation. Our existing systems within Google fall apart when high-traffic is delivered to the proxy and something accidentally defeats aggregation. We jump through hoops to ensure aggregation works as advertised since it delivers considerable traffic reduction and hence reduces operational costs.

api/v1/attributes.proto
wora: If we define real schema, let's make sure every field is documented, add example, and define the official way to provide the value. If a developer looks at this spec, they doesn't know what value to actually use to make the call.

api/v1/attributes.proto
wora: Can we add a sub-message for these fields, so we don't have to repeat every field?

api/v1/report.proto
wora: A couple of general thoughts.

1. For Check, we should only check once per minute for the same set of attributes. The server should return the set of actually used attributes. This would give us an infinitely scalable design, and also very easy to understand.

2. For Report, different backend requires different schema and has different restrictions. It is certain we need to report to different parties at the same time, billing, logging, monitoring, auditing, producer, consumer, auditor, network, health checker, load balancer, and more. How to design a usable and scalable solution is a bit unknown. We should study the existing systems.

We can model it as a set of attributes, and let mixer decide which attribute goes where. It would work well except the efficiency. We can start with this.

api/v1/attributes.proto
smawson: Added a TODO

api/v1/attributes.proto
smawson: Attempted to clarify.

api/v1/attributes.proto
smawson: I agree with Doug, let's not add anything until validated it is needed. I've removed the well-known attributes completely but left a place to put them in the future if needed. PTAL

api/v1/report.proto
smawson: I've added back intervals, using an approach similar to the stackdriver agent API of grouping values by span, rather than having a span on each value.

api/v1/attributes.proto
smawson: Removed all these for now, let's wait until we have an agreed set before putting anything in the API anywhere.

api/v1/report.proto
smawson: I've reworked the proposal a bit, added back aggregation with timespans, and removed the well known attributes.

At this point we're almost back to your proposal, minus logs, and with metrics grouped in with attributes.

We _could_ break metrics out, but I don't really see the need, it complicates the api for no real benefit. This API is really just a "report a set of attributes over some time periods" API, and not a general purpose monitoring API.

smawson:
PTAL at the latest PR, it simplifies things a bit while also adding back in aggregation and batching.

api/v1/attributes.proto
```diff
@@ -0,0 +1,48 @@
...
+syntax = "proto3";
+
+package istio.mixer.v1;
+
+option go_package="istio.io/mixer/api/v1;mixerpb";
+
+// A set of attributes reported over a period of time.
+// An instant can be represented by only reporting an end
+// time and no start time.
```
wora: In this case, I think Operation is a better term than AttributeSpan, as it represents something very concrete.

api/v1/attributes.proto
```diff
@@ -0,0 +1,48 @@
...
+
+  // The named values being reported in the span.
+  repeated Attribute attributes = 3;
+}
+
+// An individual attribute, which has a name and a value.
+message Attribute {
+  // The name of the attribute, which is dot-delimited
+  // ascii string, for example "source.device_id"
+  string name = 1;
```
wora: We need a type here. For example, string and ip address are different type, the comparison algorithm may change. Also string vs timestamp string.

api/v1/report.proto
mandarjog: We can check in a yaml of labelDescriptors / Attribute descriptors.

geeknoid:
I'm a bit confused in how this API is expected to work. Can you connect the dots for me?

From a practical perspective, our adapters need something basically like MetricValue and LogEntry in order to do their work. They need strongly-typed values and they need to know these types such that they can encode the information properly for their respective backends.

So what's the process to go from a call to Report to a set of MetricValue and LogEntry records being generated and delivered to adapters? And in the way you've defined the API right now, do you expect the Report method to be usable by services or is it only intended for the proxy?

Check is already only intended to be used by the proxy, and if we make report this way too, then I would like to change the API names: BeforeRequest and AfterRequest. This is even more generic in nature and would give us more flexibility in the implementation without needing to compromise semantics. By that I mean that BeforeRequest could readily allocate quota if we'd like, whereas having Check allocate quota is a bad idea because it is doing a mutating operation as part of an inherently/implicitly read-only thing.

api/v1/attributes.proto
smawson: That sounds reasonable, should I add that to the comment?

api/v1/attributes.proto
smawson: You want a type beyond just the type of the value?  Something like a "kind" like we have in swagger schemas?

smawson:
re: connecting the docs.

I think the config doc lays out the strategy pretty well, but it needs some work to provide cleaner examples and more background.

Basically the model for e.g. metrics is:

config says:
metric "foo" takes value from attribute metrics.request_size, is of type gauge, is labelled by attributes source.user, source.name
etc. to define more metrics.

metric adapter will take that config, then when a request comes in with some attributes, it will pull out the attributes it needs, combine them with the template defined in the config, and then send those to the backend service. So if the backend service takes a MetricValue proto, for example, the adapter should fill in that proto based on its config.

Logging is similar, logging will have a template which is filled in by the attributes, and then adapted to whatever backend system needs to be called.

So I don't expect the internal representation within mixer to be LogEntry or MetricValue, but instead to just be attributes, and specific adapters do the conversion to LogEntry if they use LogEntry.

As to whether this would be usable by services, it seems like it would be pretty easy to use, we should walk through a use case to make sure though. But for v1 the goal is proxy -> mixer, I don't think service -> mixer is in scope.

re: naming, I've actually been thinking that a generic way to do this is something similar to that, where you have basically hooks in the proxy that are based on a particular execution point, and the call to the mixer just says "i'm at this point, here are the attributes, do your thing."

on the other hand check/report are nice names for this.

I think we should go through the exercise of assuming this api is for proxy only, and a separate api is for services, and compare the two under some use cases to see how it looks, both from config and from code.

geeknoid:
I can see this working. Although implementation-wise, I want the mixer core
to be the one dealing with attributes and not the adapters. I'd like the
adapters to get structured data and leave the interaction with the
attribute system to us. This is why I want to use a MetricValue struct and
LogEntry struct internally (just Go structs, not protos). Mixer looks at
the config, extracts the necessary attributes and populates the structures
which are delivered to the adapters. Adapters can't cheat (like reading
attributes not specified in config), can't get it wrong, etc.

Do you object to switching to BeforeRequest/AfterRequest? In the
service->mixer case, then we'll have Report, Quota, CheckPermission,
CheckWhitelist, and other CheckXXX calls.

An alternate would be to just have a single method called ProxyEvent()
which takes an enum {BEFORE_REQUEST, AFTER_REQUEST}. This would let us
cleanly introduce new phases in the future if we want. And it makes the
semantics of the call totally open-ended which eliminates incorrect
preconceptions. I also like the name which makes it clear what the intended
use case is.

smawson:
ProxyEvent is close to what Louis had put somewhere, that we should have a single stream, so we can share attributes within a stream and not copy them between streams.  I think it is reasonable although maybe don't like the name that much (maybe ProcessEvent? HandleEvent?  EventStream? I kind of like EventStream, or OperationStream).

Do you want to sketch up what that would look like?  We could put it in a different .proto file within the directory.  We could then rename the AttributeSpan span thing to be an Event or an Operation or something (although it is still a span).  We do need to figure out the batching/aggregation model still though, maybe it doesn't make sense to report a batch of aggregations, batching is replaced by streaming, and aggregation is the main thing? I don't know.

Totally fine with internal structure of mixer copying into well-defined types, I think that is what Mandar had sketched up in the config doc too. API gets attributes, mixer turns those into the well defined types, and passes those into the adapter.

api/v1/attributes.proto
```diff
@@ -0,0 +1,48 @@
...
+// time and no start time.
+message AttributeSpan {
+  // The start of the span being reported.
+  google.protobuf.Timestamp start_time = 1;
+
+  // The end of the span being reported.
+  google.protobuf.Timestamp end_time = 2;
+
+  // The named values being reported in the span.
+  repeated Attribute attributes = 3;
```
smawson: So the idea was that you could have repeated attributes, but if we have a time span we kind of only have one per span, so it could be a map? Maybe leave repeated for now so we can have repeated and worry about that later (map is under the covers implemented this way anyway so no wire difference really).

mandarjog:
If we think of mixer as a filter chain extension of the proxy then the API surface would simply describe the filter chain name. 

- right after reading headers
- right after making routing decisions
- after the response is sent back 
And so on

In that case proxyFilterXXX is a more appropriate name.

Single stream for everything seems excessive,  at that point the API surface looses all its descriptive value.

api/v1/attributes.proto
douglas-reid: Please. I think it is worth documenting.

louiscryan: Operation seems like too strong a name, it implies that an instance of this type may be correlated with an operation performed by the service. I prefer the XXXEvent variants suggested elsewhere

douglas-reid:
Should we close this PR (reason: OBE) and open a new issue for adding Operations/AttributeSpans and for potential method renaming?

geeknoid:
This stuff has evolved quickly and is being addressed in different commits.

PR #41, opened 2016-12-07
Adding configManager and implementing the higher level Check logic.

ozevren:
This CL introduces a new component, ConfigManager, which is responsible for managing the configuration.

In addition, the higher-level of the check call is implemented: When a request comes in, the Check logic extracts well known serverName and peerId facts, finds the associated ListChecker configurations, obtains adapter instances with the appropriate fact checkers and applies them in sequence.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/41)

### DIFF
--- a/server/adapterManager.go
+++ b/server/adapterManager.go
@@ -17,8 +17,8 @@ package main
 import (
 	"errors"
 
+	"istio.io/mixer"
 	"istio.io/mixer/adapters"
-
 	"istio.io/mixer/adapters/denyChecker"
 	"istio.io/mixer/adapters/factMapper"
 	"istio.io/mixer/adapters/genericListChecker"
@@ -52,8 +52,14 @@ type AdapterManager struct {
 	ListCheckers map[string]adapters.Builder
 }
 
-// TODO: this implementation needs to be driven from external config instead of being
-// hardcoded
+// GetListCheckerAdapter returns a matching adapter for the given dispatchKey. If there is no existing adapter,
+// it instantiates one, based on the provided adapter config.
+func (mgr *AdapterManager) GetListCheckerAdapter(dispatchKey mixer.DispatchKey, config *adapters.AdapterConfig) (adapters.ListChecker, error) {
+	// TODO: instantiation & caching of the adapters.
+	return nil, errors.New("NYI")
+}
+
+// TODO: this implementation needs to be driven from external config instead of being hardcoded
 func prepBuilders(l []adapters.Builder) (map[string]adapters.Builder, error) {
 	m := make(map[string]adapters.Builder, len(l))
 
--- a/server/apiHandlers.go
+++ b/server/apiHandlers.go
@@ -17,9 +17,12 @@ package main
 // This is what implements the per-request logic for each API method.
 
 import (
+	"github.com/golang/glog"
+
 	"google.golang.org/genproto/googleapis/rpc/code"
 	"google.golang.org/genproto/googleapis/rpc/status"
 
+	"istio.io/mixer"
 	"istio.io/mixer/adapters"
 
 	mixerpb "istio.io/mixer/api/v1"
@@ -45,6 +48,8 @@ type APIHandlers interface {
 }
 
 type apiHandlers struct {
+	adapterManager *AdapterManager
+	configManager  *ConfigManager
 }
 
 func newStatus(c code.Code) *status.Status {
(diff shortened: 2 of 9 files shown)


config/listCheckerConfig.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+package config
+
+import (
+	"istio.io/mixer/adapters"
+	"errors"
+)
+
+type ListCheckerConfig struct {
```
geeknoid: Shouldn't this have the name of the fact to use for the symbol lookup?

config/listCheckerConfig.go
```diff
@@ -0,0 +1,31 @@
...
+	AdapterConfig *adapters.AdapterConfig
```
geeknoid: I would have expected three structs overall in this space:

- One struct which is the adapter-specific config block, defined by the adapter implementation. This is used to pass adapter-specific options.

- One struct which is the type-specific config block (one for all ListChecker adapters), which is used to determine fact->argument assignment for that type of adapter. I think the struct being defined here is of this variety.

- Some combo that includes the above two together, representing the total configuration space for a single adapter.

Given the above, I don't understand why this has a pointer to AdapterConfig.

server/configManager.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2016 Google Inc.
```
douglas-reid: should configManager live in server?  or should it be in the config package?

server/adapterManager.go
```diff
@@ -52,8 +52,14 @@ type AdapterManager struct {
 	ListCheckers map[string]adapters.Builder
 }
 
-// TODO: this implementation needs to be driven from external config instead of being
-// hardcoded
+// GetListCheckerAdapter returns a matching adapter for the serviceName and peerID. If there is no existing adapter,
+// it instantiates one, based on the provided adapter config.
+func (mgr *AdapterManager) GetListCheckerAdapter(serviceName string, peerID string, config *adapters.AdapterConfig) (adapters.ListChecker, error) {
```
geeknoid: That's a good idea. Even if we don't ever use more facts as a key, it would be clearer if we create a type specifically for the key.

server/apiHandlers.go
```diff
@@ -17,13 +17,25 @@ package main
...
+const (
+	// ServiceName is a well-known name of a fact that is passed by the caller to specify
+	// the particular config that should be used when handling a request.
+	ServiceName = "serviceName"
+
+	// PeerID is a well-known name of a fact that is passed by the caller to specify
+	// the id of the client.
+	PeerID = "peerId"
+)
+
```
geeknoid: I had been thinking about these to facts as constants in the mixer, just something we assume. But in reality, we could make this configurable at the global level in the mixer. Is that a useful feature for the operator?

server/apiHandlers.go
```diff
@@ -55,24 +69,79 @@ func newQuotaError(c code.Code) *mixerpb.QuotaResponse_Error {
...
 	tracker.UpdateFacts(request.GetFacts())
-	response.RequestId = request.RequestId
-	response.Result = newStatus(code.Code_UNIMPLEMENTED)
+
+	serviceName := tracker.GetLabels()[ServiceName]
+	peerID := tracker.GetLabels()[PeerID]
+
+	allowed, err := h.checkLists(tracker, serviceName, peerID)
+	if err != nil {
+		glog.Warning("Unexpected check error: %v", err)
```
geeknoid: We should explicitly record the key information (serviceName, peerId)

server/apiHandlers.go
```diff
@@ -55,24 +69,79 @@ func newQuotaError(c code.Code) *mixerpb.QuotaResponse_Error {
...
+		if err != nil {
+			return false, err
+		}
+
+		symbol, err := listCheckerConfig.EvaluateSymbolValue(tracker.GetLabels())
+		if err != nil {
+			return false, err
+		}
+
+		// TODO: This logic assumes that the result of CheckList will always have the same semantics
```
geeknoid: I just assumed we'd be handling polarity through config on a per-adapter basis.

server/apiHandlers.go
```diff
@@ -55,24 +69,79 @@ func newQuotaError(c code.Code) *mixerpb.QuotaResponse_Error {
...
+		// as a whitelist lookup. We need to clearly define the blacklist semantics and plumb it at some point.
+		inList, err := listCheckerAdapter.CheckList(symbol)
+		if err != nil {
+			return false, err
```
geeknoid: Might be nice to log, at the info level the adapter that denied the check.

server/configManager.go
geeknoid: I was thinking that "config" would strictly be for proto definitions representing the configuration state for the mixer.

douglas-reid: I don't know that config needs to be exclusive for the protos, especially since the protos live in version dirs a level down.

In my head, server would be limited to strictly the bits required to get the server itself up (main.go, bits required for handling grpc, some middleware perhaps); the other bits would live elsewhere (maybe in root 'mixer' package when appropriate, etc.).

So, one approach could be:
- `mixer.server.AdapterManager` -> `mixer.adapter.Manager`
- `mixer.server.ConfigManager` -> `mixer.config.Manager`

Anyways, we can resolve this later. Please don't let this hold up the review.

config/listCheckerConfig.go
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"istio.io/mixer/adapters"
```
ozevren: Done.

config/listCheckerConfig.go
ozevren: As discussed earlier, we can update the code in such a way i.e.
----------
type ListCheckerEvaluationConfig struct {
// Only config for evaluation purposes
}

// Given a config and facts, evaluate the symbol value to use.
func EvaluateListcheckerConfig(labels map[string]string, ListCheckerEvaluationConfig) (string, error)

In the apiHandler:
adapterConfig, evaluationConfig, err := configManager.getConfig(peerID, serviceName)
adapter, err := adapterManager.getAdapter(peerID, serviceName, adapterConfig)

symbolValue := EvaluateListCheckerConfig(labelValues, evaluationConfig)

adapter.CheckList(symbolValue)

----------
However, I'd expect the machinery inside the "EvaluateListCheckerConfig" to be amenable to caching, especially if the config language is complicated and has expression support (which seems to be the case). As such, my inclination was to make the ConfigManager a caching component, similar to AdapterManager, and hold onto evaluators. This makes them effectively interfaces, similar to the current design here. (Except my choice of names are a bit off. ListCheckerConfig should probably be ListCheckerConfigEvaluator and instead of having an attached adapter config, it would return the adapter config as one of the result values. (Still adjusting to Go model of doing things).

If you buy into this view (i.e. cached evaluators) then I can update the code to be a bit more coherent?

geeknoid: Yeah, I think that sounds good.

server/apiHandlers.go
```diff
@@ -17,13 +17,25 @@ package main
...
+	"github.com/golang/glog"
+
 	"google.golang.org/genproto/googleapis/rpc/code"
 	"google.golang.org/genproto/googleapis/rpc/status"
 
 	"istio.io/mixer/adapters"
 	"istio.io/mixer/api/v1"
 )
 
+const (
```
ozevren: I created and put them in DispatchKey.go. Would that work? (Will send the update soon).

server/apiHandlers.go
ozevren: Used Warningf and added "DispatchKey" information.

ozevren:
PTAL.

ozevren:
Updated the wrong comment and rebased to HEAD. PTAL (and merge? :)

PR #42, opened 2016-12-07
Improve unit tests

geeknoid:
And fix a build break due to yesterday's symbol renaming.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/42)

### DIFF
--- a/adapters/denyChecker/builder_test.go
+++ b/adapters/denyChecker/builder_test.go
@@ -32,19 +32,20 @@ func TestAll(t *testing.T) {
 
 	a, err := b.NewAdapter(b.DefaultAdapterConfig())
 	if err != nil {
-		t.Error("Unable to create adapter")
+		t.Errorf("Unable to create adapter: %v", err)
 	}
 	listChecker := a.(adapters.ListChecker)
 
-	var ok bool
-	ok, err = listChecker.CheckList("")
-	if ok || err != nil {
-		t.Error("Expecting to always get false")
-	}
+	cases := []string{"", "ABC", "ABC/DEF", "AAA", "123"}
+	for _, c := range cases {
+		ok, err := listChecker.CheckList(c)
+		if err != nil {
+			t.Errorf("CheckList(%s) failed with error %v", c, err)
+		}
 
-	ok, err = listChecker.CheckList("ABC")
-	if ok || err != nil {
-		t.Error("Expecting to always get false")
+		if ok {
+			t.Errorf("CheckList(%s) => got 'true'; want 'false'", c)
+		}
 	}
 
 	listChecker.Close()
--- a/adapters/factMapper/builder_test.go
+++ b/adapters/factMapper/builder_test.go
@@ -26,19 +26,21 @@ func TestBuilderInvariants(t *testing.T) {
 	testutil.TestBuilderInvariants(b, t)
 }
 
-func TestNoRules(t *testing.T) {
+func setup(t *testing.T, bc BuilderConfig, ac AdapterConfig) adapters.FactTracker {
 	b := NewBuilder()
-	b.Configure(b.DefaultBuilderConfig())
+	b.Configure(&bc)
 
-	rules := make(map[string]string)
-	var aa adapters.Adapter
-	var err error
-	if aa, err = b.NewAdapter(&AdapterConfig{Rules: rules}); err != nil {
+	a, err := b.NewAdapter(&ac)
+	if err != nil {
 		t.Error("Expected to successfully create a mapper")
 	}
-	a := aa.(adapters.FactConverter)
+	fc := a.(adapters.FactConverter)
+	return fc.NewTracker()
+}
 
-	tracker := a.NewTracker()
+func TestNoRules(t *testing.T) {
+	rules := make(map[string]string)
+	tracker := setup(t, BuilderConfig{}, AdapterConfig{Rules: rules})
 
 	labels := tracker.GetLabels()
 	if len(labels) != 0 {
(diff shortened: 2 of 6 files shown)


adapters/denyChecker/builder_test.go
```diff
@@ -36,15 +36,13 @@ func TestAll(t *testing.T) {
...
-
-	ok, err = listChecker.CheckList("ABC")
-	if ok || err != nil {
-		t.Error("Expecting to always get false")
+	cases := []string{"", "ABC", "ABC/DEF", "AAA", "123"}
+	for _, c := range cases {
+		var ok bool
+		ok, err = listChecker.CheckList(c)
+		if ok || err != nil {
+			t.Error("Expecting to always get false, got success with " + c)
```
douglas-reid: While we are here, can we make the error message a bit more informative?

Something like:

`t.Errorf("CheckList(%s) => got 'true'; want 'false'", c)`

Same comment for other locations in the tests.

adapters/denyChecker/builder_test.go
```diff
@@ -36,15 +36,13 @@ func TestAll(t *testing.T) {
...
-	}
-
-	ok, err = listChecker.CheckList("ABC")
-	if ok || err != nil {
-		t.Error("Expecting to always get false")
+	cases := []string{"", "ABC", "ABC/DEF", "AAA", "123"}
+	for _, c := range cases {
+		var ok bool
+		ok, err = listChecker.CheckList(c)
+		if ok || err != nil {
```
douglas-reid: can we just make this:

`if ok, err := listChecker.CheckList(c); ok || err != nil {`

?

adapters/factMapper/builder_test.go
```diff
@@ -26,17 +26,20 @@ func TestBuilderInvariants(t *testing.T) {
 	testutil.TestBuilderInvariants(b, t)
 }
 
-func TestNoRules(t *testing.T) {
+func setup(t *testing.T, bc BuilderConfig, ac AdapterConfig) adapters.FactConverter {
```
douglas-reid: what about having setup return `(adapters.FactConverter, error)` and then leaving the error handling and logging in the actual tests? That would avoid passing in `*testing.T` and keep the test bits in the body of the test methods?

adapters/factMapper/builder_test.go
```diff
@@ -26,17 +26,20 @@ func TestBuilderInvariants(t *testing.T) {
...
+	if err != nil {
 		t.Error("Expected to successfully create a mapper")
 	}
-	a := aa.(adapters.FactConverter)
+	return a.(adapters.FactConverter)
+}
+
+func TestNoRules(t *testing.T) {
+	rules := make(map[string]string)
+	a := setup(t, BuilderConfig{}, AdapterConfig{Rules: rules})
```
douglas-reid: Perhaps for another PR, but there is a lot of repetition in this TestNoRules() code.  Can we just have a cases or tests struct with a rules map and loop through with the same check and error message?

adapters/genericListChecker/builder_test.go
```diff
@@ -26,119 +26,56 @@ func TestBuilderInvariants(t *testing.T) {
 	testutil.TestBuilderInvariants(b, t)
 }
 
-func TestWhitelistWithOverrides(t *testing.T) {
+func runTests(t *testing.T, bc BuilderConfig, ac AdapterConfig, goodValues []string, badValues []string) {
```
douglas-reid: Why not just have one TestCheckList that loops over a struct that includes a name, BuilderConfig, AdapterConfig, goodValues, and badValues?

The struct could live outside the TestCheckList method.

I think that would be more idiomatic and a bit easier to read.

adapters/genericListChecker/builder_test.go
```diff
@@ -26,119 +26,56 @@ func TestBuilderInvariants(t *testing.T) {
...
 
-	goodValues := []string{"One", "Two", "Three"}
 	for _, value := range goodValues {
-		var ok bool
-		ok, err = a.CheckList(value)
-		if !ok {
-			t.Error("Expecting check to pass")
+		ok, err := a.CheckList(value)
+		if !ok || err != nil {
+			t.Error("Expecting check to pass for value " + value)
```
douglas-reid: as elsewhere, i think the error message could be slightly improved to be:

`t.Errorf("CheckList(%s): got 'false'; want 'true', value)`

server/apiHandlers.go
```diff
@@ -61,18 +61,18 @@ func NewAPIHandlers() APIHandlers {
 
 func (h *apiHandlers) Check(tracker adapters.FactTracker, request *mixerpb.CheckRequest, response *mixerpb.CheckResponse) {
 	tracker.UpdateFacts(request.GetFacts())
-	response.RequestId = request.RequestId
+	response.RequestIndex = request.RequestIndex
```
douglas-reid: we need presubmits pronto.

adapters/denyChecker/builder_test.go
geeknoid: Improved this code for better output.

adapters/factMapper/builder_test.go
geeknoid: I actually added more stuff into setup. I like having the redundant bits all in one place, especially since these bits are not really the object of the individual tests

adapters/factMapper/builder_test.go
geeknoid: Used more tables in various tests where it makes sense.

adapters/genericListChecker/builder_test.go
geeknoid: Done

adapters/genericListChecker/builder_test.go
geeknoid: Done

geeknoid:
PTAL.

PR #45, opened 2016-12-09
Add a simple client for the mixer.

geeknoid:
The client is a fully functional shell. As things mature,
I'll add options to make it possible to send any request
with any parameter to the server.

Also fixes the default command-line options for mixer
proper. You can now start the server with no args, and
start the client with no args, and a check request will
be issued and handled end-to-end.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/45)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 mixer.bin
+client.bin
 *.swp
 # Compiled Object files, Static and Dynamic libs (Shared Objects)
 *.o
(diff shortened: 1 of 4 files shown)


douglas-reid:
For future reference, we might consider [google/subcommands](https://github.com/google/subcommands) for building CLIs. It is what we've used in the past to build servicemanagement CLIs, etc.

kyessenov:
Or we can choose https://github.com/spf13/cobra since 1) it's what kubectl uses 2) spf13 is working on Go at Google.

Issue #32, opened 2016-12-06
Establish bazel.build support


douglas-reid:
Tracking issue for adding [bazel.build](https://bazel.build/) support for istio mixer.

douglas-reid:
There is now full bazel.build coverage for the code in the current repo.  Marking issue as resolved.

PR #56, opened 2016-12-12
Add initial travis-CI build config for bazel.build.

douglas-reid:
This does not attempt to provide any coverage information. It does attempt to use caching of directories to improve test performance.

There is a lot of room for improvement here, but it should establish the basic ability to run tests with our bazel.build and see the status of those tests via travis-ci. Hopefully, this can be improved with subsequent updates.

This is meant to address issues #2 and #32.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/56)

### DIFF
--- a/.bazelrc.travis
+++ b/.bazelrc.travis
@@ -1,0 +1,19 @@
+# This is from Bazel's former travis setup, to avoid blowing up the RAM usage.
+startup --host_jvm_args=-Xmx2500m
+startup --host_jvm_args=-Xms2500m
+startup --batch
+test --ram_utilization_factor=10
+
+# This is so we understand failures better
+build --verbose_failures
+
+# This is so we don't use sandboxed execution. Sandboxed execution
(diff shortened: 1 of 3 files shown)


.bazelrc.travis
```diff
@@ -0,0 +1,19 @@
...
+test --ram_utilization_factor=10
+
+# This is so we understand failures better
+build --verbose_failures
+
+# This is so we don't use sandboxed execution. Sandboxed execution
+# runs stuff in a container, and since Travis already runs its script
+# in a container (unless you require sudo in your .travis.yml) this
+# fails to run tests.
+build --spawn_strategy=standalone --genrule_strategy=standalone
```
sebastienvas: You could actually remove that part since you have a sudo call in your travis file. Sandboxing should work properly

douglas-reid: Please feel free to fix any issues here. I was just trying to get something working enough to get basic build break detection work asap.

PR #62, opened 2016-12-13
Add metricReporter interface to adapters package.

douglas-reid:
Proposal for monitoring interface for adapters that will expose metrics
to various backends. This borrows heavily from the config/v1 descriptor
schema.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/62)

### DIFF
--- a/adapters/metricsReporter.go
+++ b/adapters/metricsReporter.go
@@ -1,0 +1,63 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adapters
+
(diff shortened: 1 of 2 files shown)


adapters/metricsReporter.go
```diff
@@ -0,0 +1,86 @@
...
+		ReportMetrics([]MetricValue) error
+	}
+
+	// MetricKind represents the category of metric that is recorded in a metric
+	// value.
+	MetricKind int
+
+	// MetricValue holds the value of a metric, along with dimensional and
+	// other metadata.
+	MetricValue struct {
```
geeknoid: I think we want to separate schema from values, otherwise it is too repetitive. A metric value should say "this value corresponds to schema A", and the details of schema X indicate the description, display name, unit, etc. We can deliver all the descriptors as an map as part of aspect-level config state, and adapters can hold on to the pointer.

douglas-reid:
PTAL. I've removed the bits we discussed at the end of the meetings today.

adapters/metricsReporter.go
```diff
@@ -0,0 +1,63 @@
...
+		// time and no start time.
+		StartTime time.Time
+
+		// EndTime identifies the end time of value collection.
+		EndTime time.Time
+
+		// Attributes are a set of key-value pairs for dimensional data.
+		// These correspond directly to the attributes specified in the
+		// schema for this metric value.
+		Attributes map[string]string
```
geeknoid: I think we'll be using strongly-typed values for attributes, but using string here is fine until the issue is settled at  a higher level.

PR #66, opened 2016-12-14
Introduce Istio Attibute system. It defines a standard way for

wora:
different Istio components communicate with each other. For example,
the proxy can generate attributes, which can be logged, audited,
monitored, billed, rate limited.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/66)

### DIFF
--- a/attribute/v1/attribute.proto
+++ b/attribute/v1/attribute.proto
@@ -1,0 +1,80 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.attribute.v1;
+
+// An `Attribute` represents a small piece of metadata about an activity on a
+// network service, such as `origin.ip` of an API request.
+//
+// The Istio system will use the attributes to communicate across different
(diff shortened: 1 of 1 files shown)


douglas-reid:
I think we should consolidate this to #67.  @wora sound ok?

wora:
The other change is still mixer only. We should make it a top-level concept, so it can be widely adopted by many systems.

douglas-reid:
@geeknoid @wora would moving the attributes stuff from #67 into a top-level spot (maybe in an istio/api rep even) unblock us here?  how do we move forward with the two proposals for attributes?

ZackButcher:
+1 to moving to a top-level repo: our intent is that it's "easy" to have multiple implementations of the same interface; ours are just the canonical implementations. To me, this implies the APIs should live outside of our implementations. I think all of the interfaces (today just the proto service definitions) we're writing against should be moved up to a new istio/api repo. So, for example, istio/mixer/api/v1 moves to istio/api/mixer/v1. This should entirely remove cross-repo dependencies for our implementations as well.

geeknoid:
We've handled this in parallel in a different PR.

geeknoid:
All right, so we'll have Attribute and AttributeDescriptor as two common
types. Hong was suggesting these be in the istio.attribute.v1 package. So
where should these two files live?

In the Istio repo? In a distinct repo?

I see Mandar went ahead and created an istioapi repo. I presume that's his
vote on this topic.

douglas-reid:
I like the top-level repo, mainly because I think it allows nice separations of concerns, etc.

I think @wora is right that attributes should be a top-level concept.

I think we still have the question of whether or not we need Attributes as distinct from Metrics. If we do decide that it is reasonable to do that, we will obviously need to update the Mixer API surface to allow passing of both in Check()/Report() or whatever shape those end up being.  Should we open an issue for that discussion?

geeknoid:
All righty, Once I finish my current refactoring from facts->attributes,
I'll move attributes into the istioapi repo.
​
As for support metrics as distinct from attributes, I don't believe this is
actually necessary given the current design. We can certainly have an issue
to track further discussions.

PR #67, opened 2016-12-14
Attributes galore.

geeknoid:
Switch away from labels/facts to attributes. This also sets up
all the descriptors to properly function for the simplified
proxy-centric mixer API.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/67)


geeknoid:
Updates as discussed. PTAL.

wora:
I strongly recommend attribute should be the top-level concept, at the same level of proxy, mixer, and others. It will be the foundation of Istio, not just an aspect of mixer.

config/v1/attribute_descriptor.proto
```diff
@@ -0,0 +1,65 @@
...
+  // The type of data carried by attributes
+  ValueType value_type = 2;
+
+  // An optional human-readable description of the attribute's purpose.
+  string description = 3;
+
+  // Types of supported attribute values.
+  enum ValueType {
+    // A 64-bit signed integer.
+    INT64 = 0;
```
louiscryan: Are we trying to represent semantic types (IPv4) or data types (INT64) here?

Is the goal of this enum to represent the semantics of of values that are both lables and metrics or just labels.
We should clarify the semantic distinction between labels and metrics.

metrics -> any value where the primary intent is to be able to aggregate it or apply a numerical threshold to it
labels -> any value where the primary intent is to use it as a dimension for aggregation of a metric or as a key for a rule

Generally speaking if there is confusion about a value being a metric or a label we should err on the side of using labels even though specialized arithmetic may be done on the value to from another label dimension. Examples
- ip address ranges to determine internal vs external IP
- http status code ranges to classify error classes

PR #71, opened 2016-12-14
Disable the Bazel sandbox in our .bazelrc

ZackButcher:
There is currently a race condition in the bazel build as it imports sources in parallel. By disabling the sandbox we remove this race, making the build work consistently and deterministically.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/71)


ZackButcher:
Ping for CLA evaluation.

PR #73, opened 2016-12-15
Add cc_proto_library for api proto.

qiwzhang:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/73)

### DIFF
--- a/BUILD.protobuf
+++ b/BUILD.protobuf
@@ -1,0 +1,748 @@
+# Bazel (http://bazel.io/) BUILD file for Protobuf.
+
+licenses(["notice"])
+
+exports_files(["LICENSE"])
+
+################################################################################
+# Protobuf Runtime Library
+################################################################################
+
+COPTS = [
+    "-DHAVE_PTHREAD",
+    "-Wall",
+    "-Wwrite-strings",
+    "-Woverloaded-virtual",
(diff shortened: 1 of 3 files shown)


qiwzhang:
With this change, in my local, I can build

bazel build //api/v1:cc_proto

but I could not build

bazel build //api/v1:all

....
external/com_github_google_protobuf/src: warning: directory does not exist.
google/protobuf/any.proto: File not found.
google/rpc/status.proto: Import "google/protobuf/any.proto" was not found or had errors.
google/rpc/status.proto:89:12: "google.protobuf.Any" is not defined.
api/v1/check.proto: Import "google/rpc/status.proto" was not found or had errors.
api/v1/check.proto:37:3: "google.rpc.Status" is not defined.

douglas-reid:
This PR, to me, is evidence that we should move the API surface out of the mixer repo into a separate repo.  It adds a ton of scaffolding to support c++, of which there is no code within the repo itself.

douglas-reid:
@qiwzhang this appears to be because we now have multiple bits in the build pulling in the protobuf library in different ways.  I'll keep digging to find a workaround.

douglas-reid:
Please take a look at #75. This should add the functionality needed.

PR #76, opened 2016-12-16
Move BUILD from api/v1 to the top.

qiwzhang:
This is just to show an idea.  It will break the build.

Bazel external depencies are so different between mixer,  proxy, and with ESP.  It is better to break them.  My idea is  mixer is using top level BUILD to build go_porto files.   when proxy/ESP use "new_git_reposity" to get mixer it,  it will use its BUILD to override the top level mixer BUILD.  Its BUILD will build anything it wants with its own dependencies.  This ideal only works if there is not BUILD file in the api/v1 folder.

If we go with this idea, we can revert back the last cc_option PR.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/76)


qiwzhang:
@douglas-reid  if you are ok with this idea,  could you fix mixer BUILD?

douglas-reid:
This breaks the current Mixer build, so we will need to, at the very least, fix the libs/binaries that depend on the proto libs inside of Mixer if we go with this approach.

Instead of relying on an overlay, could we just also include a filegroup() for the proto files?  Then the ESP or whatever builds could just use that bit of the Mixer build to retrieve and build protos however it wants? Would that work?

qiwzhang:
It will be great if we can do filegroup.  But I am afraid we could not.  MIxer needs a BUILD file to define it. But ESP could not use that BUILD file.  Otherwise we could not break the dependency.

mandarjog:
The mixer client code should now move to
https://github.com/istio/mixerclient
All istio wide protos  (including mixer, config, etc) are available at
https://github.com/istio/istioapi

douglas-reid:
@qiwzhang what if we close this PR and work quickly on an approach that will work for both repos in istio/api ?

qiwzhang:
sure.

PR #87, opened 2016-12-20
Eliminate fact-related code.

geeknoid:
Also, update the example client to support passing attributes into the new API.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/87)


example/client/cmd/check.go
```diff
@@ -23,55 +23,52 @@ import (
...
-		if err != nil {
-			errorf("Unable to establish connection to %s", MixerAddress)
-			return
-		}
-		defer deleteAPIClient(cs)
+func checkCmd(rootArgs *rootArgs) *cobra.Command {
+	return &cobra.Command{
+		Use:   "check",
+		Short: "Invokes the mixer's Check API.",
+		Run: func(cmd *cobra.Command, args []string) {
```
douglas-reid: should we consider moving this out to remove all the indents, etc.?

example/client/cmd/report.go
```diff
@@ -24,72 +24,69 @@ import (
...
-				now, _ := ptypes.TimestampProto(time.Now())
-				request.LogEntries[i] = &mixerpb.LogEntry{
-					Severity:  mixerpb.LogEntry_DEFAULT,
-					Timestamp: now,
-					Payload: &mixerpb.LogEntry_TextPayload{
-						TextPayload: arg,
-					},
+			// send the request
+			request := mixerpb.ReportRequest{RequestIndex: 0, AttributeUpdate: attrs}
+			/*
```
douglas-reid: we are saving the commented out stuff here to re-establish functionality as soon as the switch is done?

example/client/cmd/report.go
```diff
@@ -24,72 +24,69 @@ import (
...
-	Run: func(cmd *cobra.Command, args []string) {
-		if len(args) == 0 {
-			errorf("Message is missing.")
-			return
-		}
+func reportCmd(rootArgs *rootArgs) *cobra.Command {
+	return &cobra.Command{
+		Use:   "report <message>...",
+		Short: "Invokes the mixer's Report API.",
+		Run: func(cmd *cobra.Command, args []string) {
```
douglas-reid: same question as above re: moving out the func.

example/client/cmd/util.go
```diff
@@ -50,23 +54,120 @@ func deleteAPIClient(cs *clientState) {
...
+type assignFunc func(index int32, value string) error
+
+func decomposeAttributes(attrs string, dict map[int32]string, assign assignFunc) error {
+	if len(attrs) > 0 {
+		for _, a := range strings.Split(attrs, ",") {
 			i := strings.Index(a, "=")
 			if i < 0 {
-				return nil, fmt.Errorf("Attribute value %v does not include an = sign", a)
+				return fmt.Errorf("Attribute value %v does not include an = sign", a)
 			} else if i == 0 {
```
douglas-reid: unnecessary else?

example/client/cmd/util.go
```diff
@@ -50,23 +54,120 @@ func deleteAPIClient(cs *clientState) {
...
+	return nil
+}
+
+func parseAttributes(rootArgs *rootArgs) (*mixerpb.Attributes, error) {
+	attrs := mixerpb.Attributes{}
+	attrs.Dictionary = make(map[int32]string)
+
+	// once again, the following boilerplate would be more succinct with generics...
+
+	if err := decomposeAttributes(rootArgs.stringAttributes, attrs.Dictionary, func(index int32, value string) error {
```
douglas-reid: whoa. can we move the func() up a line or something (maybe pull it out entirely) to make this bit easier to follow?

(same question for the rest of these).

geeknoid:
*[example/client/cmd/report.go, line 60 at r1](https://reviewable.io:443/reviews/istio/mixer/87#-KZSZ81thMyV_YpwNESO-r1-60:-KZSlp4kJzThkw2wqPo7:b-mlh7un) ([raw file](https://github.com/istio/mixer/blob/155cc993acad39b225c172965384bd722b3f9d16/example/client/cmd/report.go#L60)):*

we are saving the commented out stuff here to re-establish functionality as soon as the switch is done?
</blockquote></details>

Yep, added a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/87)*

geeknoid:
Review status: 0 of 27 files reviewed at latest revision, 5 unresolved discussions.

---

*[example/client/cmd/check.go, line 30 at r1](https://reviewable.io:443/reviews/istio/mixer/87#-KZSZ81thMyV_YpwNESM-r1-30:-KZSlj2pQ7PokEeOJ683:b-896fix) ([raw file](https://github.com/istio/mixer/blob/155cc993acad39b225c172965384bd722b3f9d16/example/client/cmd/check.go#L30)):*

should we consider moving this out to remove all the indents, etc.?
</blockquote></details>

Done.

---

*[example/client/cmd/report.go, line 31 at r1](https://reviewable.io:443/reviews/istio/mixer/87#-KZSZ81thMyV_YpwNESO-r1-31:-KZSln4A1CZlXCvyEaQJ:b-896fix) ([raw file](https://github.com/istio/mixer/blob/155cc993acad39b225c172965384bd722b3f9d16/example/client/cmd/report.go#L31)):*

same question as above re: moving out the func.
</blockquote></details>

Done.

---

*[example/client/cmd/util.go, line 65 at r1](https://reviewable.io:443/reviews/istio/mixer/87#-KZSZ81thMyV_YpwNESQ-r1-65:-KZSlw3uxUQepvOStDyC:b-896fix) ([raw file](https://github.com/istio/mixer/blob/155cc993acad39b225c172965384bd722b3f9d16/example/client/cmd/util.go#L65)):*

unnecessary else?
</blockquote></details>

Done.

---

*[example/client/cmd/util.go, line 89 at r1](https://reviewable.io:443/reviews/istio/mixer/87#-KZSZ81thMyV_YpwNESQ-r1-89:-KZSlxdlq20d3nCasNbo:b-896fix) ([raw file](https://github.com/istio/mixer/blob/155cc993acad39b225c172965384bd722b3f9d16/example/client/cmd/util.go#L89)):*

whoa. can we move the func() up a line or something (maybe pull it out entirely) to make this bit easier to follow?

(same question for the rest of these).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/87)*

geeknoid:
PTAL

example/client/cmd/check.go
```diff
@@ -23,55 +23,55 @@ import (
...
+	var attrs *mixerpb.Attributes
+	var err error
 
-		var attrs map[string]string
-		if attrs, err = parseAttributes(Attributes); err != nil {
-			errorf(err.Error())
-			return
-		}
+	if attrs, err = parseAttributes(rootArgs); err != nil {
+		errorf(err.Error())
```
douglas-reid: errorf() is logging to std err?  why not use the glog.Errorf() here?

example/client/cmd/util.go
```diff
@@ -50,23 +54,112 @@ func deleteAPIClient(cs *clientState) {
...
-				return nil, fmt.Errorf("Attribute value %v does not include an = sign", a)
-			} else if i == 0 {
-				return nil, fmt.Errorf("Attribute value %v does not contain a valid name", a)
+func parseAttributes(rootArgs *rootArgs) (*mixerpb.Attributes, error) {
+	attrs := mixerpb.Attributes{}
+	attrs.Dictionary = make(map[int32]string)
+
+	// once again, the following boilerplate would be more succinct with generics...
+
+	for i := 0; i < 6; i++ {
```
douglas-reid: can we do this with an unexported enum, so the cases are more readable?

and/or don't loop and just do each section separately (the shared bits are actually quite small -- the parsing of the args list, right?).

geeknoid:
Review status: 0 of 27 files reviewed at latest revision, 7 unresolved discussions.

---

*[example/client/cmd/util.go, line 63 at r2](https://reviewable.io:443/reviews/istio/mixer/87#-KZSZ81thMyV_YpwNESQ-r2-63:-KZTFEUzPVPU7qgtO2uA:b-896fix) ([raw file](https://github.com/istio/mixer/blob/daf3384c7b016c123bbdd190e541d8fe9b2bd391/example/client/cmd/util.go#L63)):*

can we do this with an unexported enum, so the cases are more readable?

and/or don't loop and just do each section separately (the shared bits are actually quite small -- the parsing of the args list, right?).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/87)*

example/client/cmd/check.go
geeknoid: Done

geeknoid:
PTAL

douglas-reid:
I don't see the latest changes?  Did you push up the latest?

example/client/cmd/util.go
```diff
@@ -50,25 +54,124 @@ func deleteAPIClient(cs *clientState) {
...
+				if eq == 0 {
+					return nil, fmt.Errorf("Attribute value %v does not contain a valid name", a)
+				}
+				name := a[0:eq]
+				value := a[eq+1:]
+
+				index := int32(len(attrs.Dictionary))
+				attrs.Dictionary[index] = name
+
+				switch i {
```
douglas-reid: can we eliminate this ugliness with something more broken down into components and using code like:

```

func attrMap(s string) (map[string]string, error) {
	m := make(map[string]string)
	for _, attrStr := range strings.Split(s, ",") {
		keyVal := strings.Split(attrStr, "=")
		if len(keyVal) != 2 || keyVal[0] == "" || keyVal[1] == "" {
			return nil, errors.New("doh!")
		}
		m[keyVal[0]] = keyVal[1]
	}
	return m, nil
}

type convertFn func(string) (interface{}, error)

var (
	int64Fn = func(s string) (interface{}, error) { return strconv.ParseInt(s, 10, 64) }
	boolFn = func(s string) (interface{}, error) { return strconv.ParseBool(s) }
)

func convert(m map[string]string, f convertFn) (map[string]interface{}, error) {
	n := make(map[string]interface{}, len(m))
	for k, v := range m {
		nv, err := f(v)
		if err != nil {
			return nil, err
		}
		n[k] = nv
	}
	return n, nil
}

func toBoolMap(m map[string]interface{}) (map[string]bool, error) {
	out := make(map[string]bool, len(m))
	for k, v := range m {
		out[k] = v.(bool)
	}
	return out, nil
}


func main() {

	var good = "t=false,s=3453,z=testing"
	fmt.Println(attrMap(good))

	var bad = "t=134,s=234234,=234,w="
	fmt.Println(attrMap(bad))
	
	i, _ := attrMap("t=false,s=true")
	fmt.Println(convert(i, boolFn))
	
	c, _ := convert(i, boolFn)
	fmt.Println(toBoolMap(c))
}
```

The advantages would be smaller, more legible (debateable perhaps) code.

geeknoid:
PTAL.

PR #94, opened 2016-12-21
Addition of Aspect / Adapter / Aspect Manager

mandarjog:
This PR adds Aspect, Adapter and Aspect Manager.

1. It adds the base types and interfaces
2. designates the aspectmanager as the uniform interface to aspect.
 
Aspect managers
1. build a specific aspects
2. Execute an aspect instance by mapping attributes --> input and other aspect specific business logic.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/94)


geeknoid:
*[BUILD.api, line 10 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXy9frxNNrB8dB3PrQ:-KZXy9fsYomYatWpp-eo:b-124kdj) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L10)):*
> ```
> 
> go_proto_library(
>     # use go_default_library here to prevent need to append lib name on imports
> ```

Comment is incorrect.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[BUILD.api, line 48 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXyIcax8sgDccFV6Xg:-KZXyIcax8sgDccFV6Xh:b-124kdj) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L48)):*
> ```
> 
> go_proto_library(
>     # use go_default_library here to prevent need to append lib name on imports
> ```

Comment is incorrect.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[BUILD.api, line 51 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXyNGf8aTCkqosGfz3:-KZXyNGf8aTCkqosGfz4:b41mhyp) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L51)):*
> ```
>     name = "istio/config/v1",
>     imports = [
>         "../../external/com_github_google_protobuf/src"
> ```

Definitely not the right way to do this. I think the importmap feature is what you need here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[Makefile, line 16 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXzFRJZysE7UxX0MlG:-KZXzFRJZysE7UxX0MlH:bkr40y9) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/Makefile#L16)):*
> ```Makefile
> 
> build:
> 	@bazel build //$(T):go_default_library --define istio=local  --verbose_failures
> ```

Presumably, this is temporary?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[BUILD, line 9 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-5jKRVg_-0b5W7wq:-KZY-5jKRVg_-0b5W7wr:b-uu37w6) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD#L9)):*
> ```Python
> go_prefix("istio.io/mixer")
> 
> config_setting(
> ```

Why is this stuff needed in here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[cmd/client/check.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-H2GARdXD4yuDGzl:-KZY-H2HtmB2iTpAnlzF:b-n1oxtc) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/cmd/client/check.go#L23)):*
> ```Go
> 
> 	"github.com/spf13/cobra"
> 	mixerpb "istio.io/api/mixer/api/v1"
> ```

Do you want to move all the apis around? I thought we settled on the current location?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[pkg/api/BUILD, line 5 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-m5wSmxiIb3lWznQ:-KZY-m5xTQ-EHXE2KmN0:bnykfvz) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/api/BUILD#L5)):*
> ```Python
> load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
> 
> config_setting(
> ```

@douglas-reid is there a way we could have a centralized definition of config_setting and ISTIO_DEPTS so we don't repeat this boilerplate all over the place?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[pkg/attribute/BUILD, line 9 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY0VRYrHT4RTlvhI4M:-KZY0VRYrHT4RTlvhI4N:bwtkulj) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/attribute/BUILD#L9)):*
> ```Python
>     values = { "define": "istio=local" }
> )
> ISTIO_DEPS = [
> ```

ISTIO_DEPS, and config_settings are not necessary in this file. The code here is independent of the mixer API.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[pkg/attribute/expressionEvaluator.go, line 17 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY0lkt9QDt9hsCgMDe:-KZY0lkt9QDt9hsCgMDf:bpom2op) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/attribute/expressionEvaluator.go#L17)):*
> ```Go
> package attribute
> 
> type (
> ```

I don't think this type belongs in the attribute package. This is a higher level components that operates on attributes, but is not itself part of the 'attribute infrastructure'

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/api/BUILD, line 5 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-m5wSmxiIb3lWznQ:-KZY8Cy2C0hjvpmqwPRk:b-mykei6) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/api/BUILD#L5)):*

@douglas-reid is there a way we could have a centralized definition of config_setting and ISTIO_DEPTS so we don't repeat this boilerplate all over the place?
</blockquote></details>

I think there is, I have made several attempts at it, but so for unsuccessful. 
general idea is that create a bazel rule which selects a repo on the fly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[Makefile, line 16 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXzFRJZysE7UxX0MlG:-KZXzMfN_DvfeBRNrkhi:buwrnox) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/Makefile#L16)):*

Presumably, this is temporary?
</blockquote></details>

yes, it is convenient for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[BUILD, line 9 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-5jKRVg_-0b5W7wq:-KZYBP7HqGADkdPjaXl1:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD#L9)):*

Why is this stuff needed in here?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
Reviewed 1 of 28 files at r1.
Review status: 1 of 28 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[pkg/aspect/types.go, line 30 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYBoi8pBuWq-_6hxC1:-KZYBoi8pBuWq-_6hxC2:b-cwqws2) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L30)):*
> ```Go
> 
> 	// AdapterConfig provides common configuration params
> 	// Impl config protobufs are requested to support these
> ```

s/requested/required/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[pkg/aspect/types.go, line 27 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYCe05SEwRe4CQnUyr:-KZYCe05SEwRe4CQnUys:bvxhuzi) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L27)):*
> ```Go
> )
> 
> type (
> ```

I'm definitely not a fan of the "types.go" pattern. I would much rather have the types defined in files with meaningful names, close to the implementation code when that makes sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
Review status: 1 of 28 files reviewed at latest revision, 18 unresolved discussions.

---

*[pkg/aspect/types.go, line 50 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYCBhCphdvxYr-eiNA:-KZYCBhCphdvxYr-eiNB:ba06351) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L50)):*
> ```Go
> 		ValidateConfig(config proto.Message) error
> 
> 		io.Closer
> ```

I think it's customary to put this (io.Closer) at the top of the definition.

---

*[pkg/aspect/listChecker/types.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEKxPfr4Iin8gtAaD:-KZYEKxPfr4Iin8gtAaE:bu7coch) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/types.go#L23)):*
> ```Go
> 
> type (
> 	// AdapterConfig -- Developer visible typed config for
> ```

Who is the developer in this context? Do you mean operator?

---

*[pkg/aspect/registry/manager.go, line 15 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYFVl0gJGvzZSWLnQ7:-KZYFVl2WiEE_SpiPGVm:bjlxt22) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L15)):*
> ```Go
> // limitations under the License.
> 
> package registry
> ```

Since this code is not actually an aspect, maybe this could go instead in pkg/aspectregistry?

---

*[pkg/aspect/registry/manager.go, line 45 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEtXiA8B8C3pgx0f7:-KZYEtXiA8B8C3pgx0f8:bkbh155) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L45)):*
> ```Go
> func (m *Manager) Execute(cfg *aspect.Config, ctx attribute.Context, mapper attribute.ExprEvaluator) (*aspect.Output, error) {
> 	mgr, found := m.mreg[cfg.Aspect.Kind]
> 	if !found {
> ```

Combine with previous line

---

*[pkg/aspect/registry/manager.go, line 50 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEqgwr4LoTH9H_8gA:-KZYEqgwr4LoTH9H_8gB:bkbh155) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L50)):*
> ```Go
> 
> 	adapter, found := m.areg.ByImpl(cfg.Adapter.Impl)
> 	if !found {
> ```

Combine with previous line

---

*[pkg/aspect/registry/manager.go, line 55 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYFH9gCT9asMlcIYMv:-KZYFH9gCT9asMlcIYMw:bkbh155) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L55)):*
> ```Go
> 
> 	asp, err := m.CacheGet(cfg, mgr, adapter)
> 	if err != nil {
> ```

Combine with previous line

---

*[pkg/aspect/registry/manager.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYF110dB6CM_Qe6QZo:-KZYF110dB6CM_Qe6QZp:bit4qe8) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L58)):*
> ```Go
> 		return nil, err
> 	}
> 	// TODO act on aspect.Output and mutate attribute.Context
> ```

Never mutate attribute.Context!

attribute.Context reflects the shared state between the proxy and the mixer. If the mixer mutates this state, it'll get out of sync with what the proxy knows about, and then the universe will explode shortly thereafter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/registry/manager.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYF110dB6CM_Qe6QZo:-KZYFxxvkRdrrJ_J4GxI:bubwnlz) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L58)):*

Never mutate attribute.Context!

attribute.Context reflects the shared state between the proxy and the mixer. If the mixer mutates this state, it'll get out of sync with what the proxy knows about, and then the universe will explode shortly thereafter.
</blockquote></details>

When an aspect injects new attributes into the system, what do we call this action?
I was referring to it as mutate. Is that to be done in a different place / data struct ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/attribute/BUILD, line 9 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY0VRYrHT4RTlvhI4M:-KZYFXvcaKO1QfkRALYh:b-2c1e2e) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/attribute/BUILD#L9)):*

ISTIO_DEPS, and config_settings are not necessary in this file. The code here is independent of the mixer API.
</blockquote></details>

context.go #checkPreconditions() needs attrs *mixerpb.Attributes 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/listChecker/types.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEKxPfr4Iin8gtAaD:-KZYGNUMoKMAHHfV3aZu:b-milwgs) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/types.go#L23)):*

Who is the developer in this context? Do you mean operator?
</blockquote></details>

I mean the developer who will write an adapter implementation. 
I want to draw distinction between what a "User" sees and should care about, vs what a developer sees in terms of interfaces.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/registry/manager.go, line 15 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYFVl0gJGvzZSWLnQ7:-KZYGbZ8Jio0LP-Cii8d:b-w6i2n1) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L15)):*

Since this code is not actually an aspect, maybe this could go instead in pkg/aspectregistry?
</blockquote></details>

makes sense. I will move it over. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[cmd/client/check.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-H2GARdXD4yuDGzl:-KZYHW056IPCu8SboB32:b-nquikt) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/cmd/client/check.go#L23)):*

Do you want to move all the apis around? I thought we settled on the current location?
</blockquote></details>

maintaining its correct name per repo. I must have missed the fact that we agreed on a different name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/types.go, line 50 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYCBhCphdvxYr-eiNA:-KZYKti5JzX1zvL2jaKe:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L50)):*

I think it's customary to put this (io.Closer) at the top of the definition.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
*[pkg/aspect/listChecker/types.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEKxPfr4Iin8gtAaD:-KZYS3i1NROYbJWEHv9g:b-9p8511) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/types.go#L23)):*

I mean the developer who will write an adapter implementation. 
I want to draw distinction between what a "User" sees and should care about, vs what a developer sees in terms of interfaces.
</blockquote></details>

Let's say "adapter author" then for clarity.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
Review status: 1 of 28 files reviewed at latest revision, 17 unresolved discussions.

---

*[pkg/aspect/registry/manager.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYF110dB6CM_Qe6QZo:-KZYSE5W96rDocakf6-z:b8t1hmy) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L58)):*

When an aspect injects new attributes into the system, what do we call this action?
I was referring to it as mutate. Is that to be done in a different place / data struct ?


</blockquote></details>

I think it means we want to introduce a wrapper around attribute.context which holds locally mutated attributes on top of the 'protocol' attributes which are immutable.

I'll take care of adding this functionality.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

douglas-reid:
Reviewed 28 of 28 files at r1.
Review status: all files reviewed at latest revision, 26 unresolved discussions.

---

*[BUILD.api, line 5 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYTAqT_K2cMQ5574p1:-KZYTAqT_K2cMQ5574p2:b-ys0tu9) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L5)):*
> ```
> load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
> 
> go_prefix("istio.io/api")
> ```

I think we need to think about @wora's comments in [istio/api #3](https://github.com/istio/api/issue/3).  This now means we have 'api' showing up in multiple places in the path.

At some point, we should rationalize the paths. I actually like the suggestions of istio/mixer/v1, istio/config/v1, etc. We don't have to decide now, but let's get to it soon.

---

*[BUILD.api, line 51 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXyNGf8aTCkqosGfz3:-KZYTj1NlW-eZu_r2G1D:b-xwuup8) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L51)):*

Definitely not the right way to do this. I think the importmap feature is what you need here.
</blockquote></details>

importmap isn't needed for "well-known" protos.  but you do need to specify where to find the sources for imports.

---

*[cmd/client/check.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-H2GARdXD4yuDGzl:-KZYUrOln6L0GrvcnoZA:b-jzib74) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/cmd/client/check.go#L23)):*

maintaining its correct name per repo. I must have missed the fact that we agreed on a different name.
</blockquote></details>

Please see my other comments.  I think we should consider @wora's proposal for the istio/api repo structure.

---

*[pkg/api/BUILD, line 5 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-m5wSmxiIb3lWznQ:-KZYV6c334qxpQaoB9bd:b3vnjuk) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/api/BUILD#L5)):*

I think there is, I have made several attempts at it, but so for unsuccessful. 
general idea is that create a bazel rule which selects a repo on the fly.
</blockquote></details>

Agree that there probably is (we could write our own rules, etc.). But I'd have to experiment for a few hours/days to know for sure.  Maybe duplication is OK for now; clean up later (should we file an issue to track?)

---

*[pkg/aspect/BUILD, line 22 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYVU3f2Q6cNVA9aaCK:-KZYVU3f2Q6cNVA9aaCL:b-9frigr) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/BUILD#L22)):*
> ```Python
>     "@com_github_golang_protobuf//proto:go_default_library",
>     "@com_github_google_go_genproto//googleapis/rpc/code:go_default_library",
>     "//pkg/attribute:go_default_library"] +select ({
> ```

need a format pass here?

---

*[pkg/aspect/types.go, line 27 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYCe05SEwRe4CQnUyr:-KZYVa-foO2Z6DMVjKAs:b32v57) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L27)):*

I'm definitely not a fan of the "types.go" pattern. I would much rather have the types defined in files with meaningful names, close to the implementation code when that makes sense.
</blockquote></details>

+1

---

*[pkg/aspect/types.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYVsI_w2rHj9L2Pvyx:-KZYVsI_w2rHj9L2Pvyy:b-v6a17z) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L58)):*
> ```Go
> 		Name() string
> 		// Close - ability to close aspect
> 		io.Closer
> ```

if we put the other at top, might as well do it here.

---

*[pkg/aspect/types.go, line 79 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYWRuvNIFrC5l_Ps_k:-KZYWRuvNIFrC5l_Ps_l:b-lmiq4a) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L79)):*
> ```Go
> 	}
> 	// Config combines all configuration related to an aspect
> 	Config struct {
> ```

So we have: Cfg, Config, AdapterCfg, and AdapterConfig.  I find this a bit hard to follow.  Are there different names, etc., that we could use to distinguish?

Maybe Config, CombinedConfig, AdapterConfig, and AdapterImpConfig or something?

---

*[pkg/aspect/types.go, line 83 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXRwZfydkRF4LqCW9:-KZYXRw_vVFb51Z-rryv:b-eh9sfm) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L83)):*
> ```Go
> 		Adapter *AdapterCfg
> 	}
> 	// AdapterCfgReg registry maps from adapter "kind" -->
> ```

missing the other half?

---

*[pkg/aspect/listChecker/BUILD, line 22 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXV_cahQIJAr4y1MW:-KZYXV_cahQIJAr4y1MX:b4qx3rp) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/BUILD#L22)):*
> ```Python
>     "//pkg/aspect:go_default_library",
>     "@com_github_golang_protobuf//proto:go_default_library",
>     "@com_github_google_go_genproto//googleapis/rpc/code:go_default_library",
> ```

formatting?

---

*[pkg/aspect/listChecker/manager.go, line 22 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXZDfuOiklaXeTQMX:-KZYXZDfuOiklaXeTQMY:b-4k6q0t) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/manager.go#L22)):*
> ```Go
> 
> 	"google.golang.org/genproto/googleapis/rpc/code"
> 	listcheckerpb "istio.io/api/istio/config/v1/aspect/listChecker"
> ```

can we move this named import to its own section of the import block?

---

*[pkg/aspect/listChecker/manager.go, line 41 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXu--PJcRuG9B8ejn:-KZYXu--PJcRuG9B8ejo:b-93gitk) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/manager.go#L41)):*
> ```Go
> 
> // NewAspect creates a listChecker aspect
> func (m *manager) NewAspect(cfg *aspect.Config, _aa aspect.Adapter) (aspect.Aspect, error) {
> ```

can we avoid leading/trailing underscores in names?

---

*[pkg/aspect/registry/BUILD, line 21 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYYEstxfDtI705badO:-KZYYEstxfDtI705badP:b4qx3rp) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/BUILD#L21)):*
> ```Python
> DEPS = ["//pkg/aspect:go_default_library",
>     "//pkg/aspect/listChecker:go_default_library",
>     "//pkg/attribute:go_default_library"] +select ({
> ```

formatting?

---

*[pkg/aspect/registry/manager.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYF110dB6CM_Qe6QZo:-KZYYMKl4UELRhS97E6Q:b-4k9j4g) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L58)):*

I think it means we want to introduce a wrapper around attribute.context which holds locally mutated attributes on top of the 'protocol' attributes which are immutable.

I'll take care of adding this functionality.

</blockquote></details>

should we just use the [context](https://golang.org/pkg/context/) for request-scoped information? we will likely want to pass those around in some future state anyways (to cover deadlines and pass traces, etc.).x

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/types.go, line 79 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYWRuvNIFrC5l_Ps_k:-KZYbyLwemNKPDiynYoM:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L79)):*

So we have: Cfg, Config, AdapterCfg, and AdapterConfig.  I find this a bit hard to follow.  Are there different names, etc., that we could use to distinguish?

Maybe Config, CombinedConfig, AdapterConfig, and AdapterImpConfig or something?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/listChecker/manager.go, line 22 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXZDfuOiklaXeTQMX:-KZYcZw8jz_NR5tpAjhg:bb6un1j) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/manager.go#L22)):*

can we move this named import to its own section of the import block?
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/listChecker/BUILD, line 22 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXV_cahQIJAr4y1MW:-KZYcn3isO5Nx12nI5-l:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/BUILD#L22)):*

formatting?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/listChecker/manager.go, line 41 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXu--PJcRuG9B8ejn:-KZYgHDl7vFwDdImH7gi:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/manager.go#L41)):*

can we avoid leading/trailing underscores in names?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/listChecker/types.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEKxPfr4Iin8gtAaD:-KZYhhONDXPKGZWs4CT2:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/listChecker/types.go#L23)):*

Let's say "adapter author" then for clarity.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/registry/manager.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYF110dB6CM_Qe6QZo:-KZYkTE6ySwQQw59RXhD:bnwveme) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L58)):*

should we just use the [context](https://golang.org/pkg/context/) for request-scoped information? we will likely want to pass those around in some future state anyways (to cover deadlines and pass traces, etc.).x
</blockquote></details>

attribute.Context leads me to believe that they are attributes available in the current request scope (here request==message).  These are the attributes that we expect to see. 
If there is a special set of immutable attributes underlying this set, that we should name it something else. It perhaps does not need to be exported.
attribute.Context    attribute.baseContext. @geeknoid  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
Addressed all comments apart from splitting types.go into multiple parts.
@geeknoid @douglas-reid 
PTAL

mandarjog:
*[pkg/expr/types.go, line 17 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY0lkt9QDt9hsCgMDe:-KZZkzSpEOVj12E6_Ovv:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/expr/types.go#L17)):*

I don't think this type belongs in the attribute package. This is a higher level components that operates on attributes, but is not itself part of the 'attribute infrastructure'
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

douglas-reid:
This looks pretty good. I've closed most of my comments, as I think bits like the api import paths and BUILD file formatting can be addressed later.

One thing: instead of pkg/aspectregistry, why not just put the aspect registry stuff in pkg/aspect ?  We could have an expanded types.go and a manager.go in that package.  That seems somewhat cleaner to me.  Thoughts?

---

Reviewed 23 of 23 files at r2.
Review status: 30 of 31 files reviewed at latest revision, 19 unresolved discussions.

---

*[cmd/client/check.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-H2GARdXD4yuDGzl:-KZZmJw_2lUqNSB7XEC5:bhfpvq8) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/cmd/client/check.go#L23)):*

Please see my other comments.  I think we should consider @wora's proposal for the istio/api repo structure.
</blockquote></details>

As above, let's deal with this in a subsequent PR.

---

*[pkg/aspect/types.go, line 27 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYCe05SEwRe4CQnUyr:-KZZlA3sidD4ZUONNi4j:b-jwx13p) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L27)):*

+1
</blockquote></details>

@mandarjog and I looked at k8s codebase and discovered a ton of types.go in packages. So, in general, it seems at least somewhat accepted to use types.go in packages. However, it appears that this is the only file in the aspect pkg. So, in this case, it seems like maybe just aspect.go is  the better choice. Thoughts?

---

*[pkg/aspect/types.go, line 32 at r2](https://reviewable.io:443/reviews/istio/mixer/94#-KZZlxOVbpKN592DZQe0:-KZZlxOVbpKN592DZQe1:b-313srj) ([raw file](https://github.com/istio/mixer/blob/eac590a7c8aaaa35a91d34dc30febe4afd07020f/pkg/aspect/types.go#L32)):*
> ```Go
> 	// This interface is extended by specific aspects
> 	Adapter interface {
> 		// Close - ability to close Adapter
> ```

extreme nit: probably don't need to comment embedded interfaces.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94#-:-KZZmxPEoIZYJjAhRXrk:bvpdsh0)*

mandarjog:
*[pkg/aspectregistry/BUILD, line 21 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYYEstxfDtI705badO:-KZ_CQhxt6kiEXuAYJhL:bmypplu) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspectregistry/BUILD#L21)):*

formatting?
</blockquote></details>

 Have formatted all BUILD files with standard Python pep8 formayter

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
All specific aspects depend on aspect

So aspect/aspect1 --> aspect 
And 
Aspectregistry --> aspect/aspect1
So putting registry with aspect will make a circular dep

Re build files, I have run standard Python pep8 formatted on them.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94#-:-KZ_CxO9w7Mo6Hh7suej:b-yejh2m)*

douglas-reid:
Ok. But couldn't we organize the code differently to eliminate the cycle and have Registry be in pkg/aspect?  It seems like if we moved the various Aspect interfaces up to pkg/aspect too, the Registry code wouldn't need to look into the various aspect packages.  So, type listchecker.Aspect would instead become aspect.ListChecker (and be more idiomatic, as it exposes the single method of CheckList()).

This seems, at first glance, to be reasonable. The top-level aspect package defines all of the aspects. Then the specific aspect packages define the manager behavior, etc. for their individual aspect types. I'd actually prefer that I think -- as right now the code is overrun with Aspect types named *.Aspect.

Other issues: a few more minor nits around comments and had some questions about the Manager and CheckList signature.

---

Reviewed 1 of 1 files at r3.
Review status: 31 of 32 files reviewed at latest revision, 29 unresolved discussions.

---

*[pkg/aspect/types.go, line 22 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_jtHHOtA4F2vMahjd:-KZ_jtHHOtA4F2vMahje:b-18570h) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L22)):*
> ```Go
> 	"github.com/golang/protobuf/proto"
> 	"google.golang.org/genproto/googleapis/rpc/code"
> 	istiopb "istio.io/api/istio/config/v1"
> ```

nit: let's move this down to the bottom below the other istio.io imports (with a newline sep).

---

*[pkg/aspect/types.go, line 47 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_a3752wWhJddlZhds:-KZ_a3752wWhJddlZhdt:b-7pyk4r) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L47)):*
> ```Go
> 
> 	// Aspect -- User visible cross cutting concern
> 	Aspect interface {
> ```

this does leave us with aspect.Aspect.  Should we move this up a level?  Or, and I already regret suggesting this, use a different name (Instance/Builder/???)?

---

*[pkg/aspect/types.go, line 57 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ__SIRVcwpiklWUelp:-KZ__SIRVcwpiklWUelq:bt94pdf) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L57)):*
> ```Go
> 	// at present it only has an additional field that stores the converted
> 	// and typed proto
> 	Cfg struct {
> ```

final nits here:  what about just expanding the type names here: Cfg -> Config, Reg -> Registry ?  I think that might match the rest of the code better.

---

*[pkg/aspect/types.go, line 104 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_bUfotLABxg9aNrJ7:-KZ_bUfotLABxg9aNrJ8:b-3cvdyr) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L104)):*
> ```Go
> 	Manager interface {
> 		// Execute dispatch to the given aspect using aspect and adapter configs
> 		// A cached instance of Aspect is provided that was previsouly obtained by
> ```

s/previsouly/previously/

---

*[pkg/aspect/listChecker/manager.go, line 77 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_ekBD43OcmIOLjF8k:-KZ_ekBD43OcmIOLjF8l:btt4xj7) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/listChecker/manager.go#L77)):*
> ```Go
> 	// check if mapping is available
> 
> 	if symbolExpr, ok = cfg.Aspect.Inputs["Symbol"]; !ok {
> ```

I'm a bit confused looking at this again. If we are going to hardcode `"Symbol"` in here, what is the advantage of then shoving it back in the Input pb in order to call `CheckList()`?  Why not just have `CheckList(string)` at this point? 

It seems like that if we want to use the proto schema for validation, we could use it to extract the field names, validate that mappings for those fields exist, and then use an internal struct that includes the information (or skip the struct entirely).

Or, if we insist on passing in the proto, just do it directly (skip the Arg struct entirely) -> `CheckList(l *listcheckerpb)`.

What am I missing?

---

*[pkg/aspect/listChecker/types.go, line 34 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_htajEVdOcITEHln4:-KZ_htajEVdOcITEHln5:b-rvmr97) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/listChecker/types.go#L34)):*
> ```Go
> 	// Arg -- Developer visible input passed into the perform fn
> 	// of the aspect
> 	Arg struct {
> ```

Is this used anywhere else? Does it need to be exported?

---

*[pkg/aspect/listChecker/types.go, line 42 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_feo06palpK545idq:-KZ_feo06palpK545idr:b-etqkp4) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/listChecker/types.go#L42)):*
> ```Go
> 		aspect.Aspect
> 		// CheckList verifies whether the given symbol is on the list.
> 		CheckList(Arg *Arg) (bool, error)
> ```

Why not just have this be `CheckList(symbol string) (bool, error)`? What advantage does having the pb indirection give?  If we want flexibility, we could still use the Args struct, but I would think we could have the args struct look something like:

```
type Args struct {
   Symbol string
}
```

Or as suggested elsewhere, get rid of Arg altogether and just have `CheckList(l *listcheckerpb) (bool, error)`?

---

*[pkg/aspectregistry/manager.go, line 26 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_mLmdvO_SHYqa448y:-KZ_mLmdvO_SHYqa448z:bk8hcmv) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspectregistry/manager.go#L26)):*
> ```Go
> 
> // NewManager Creates a new Uber Aspect manager
> // provides an Execute function to act on a aspect/adapter
> ```

nit: clean up comments (consistent caps, full sentences).

---

*[pkg/aspectregistry/types.go, line 30 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_lsZe-I4OP-7e0UcX:-KZ_lsZe-I4OP-7e0UcY:bxsfq7i) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspectregistry/types.go#L30)):*
> ```Go
> 
> // Registry -- Interface used by adapters to register themselves
> type Registry interface {
> ```

Here, again, we have stutter: aspectregistry.Registry.  

---

*[pkg/expr/types.go, line 25 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_kjJnuiMHANcvvrlH:-KZ_kjJnuiMHANcvvrlI:b-upq7wg) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/expr/types.go#L25)):*
> ```Go
> 		Eval(mapExpression string, ctx attribute.Context) (interface{}, error)
> 
> 		// EvalPredicate Evaluates given predicate using the attribute context
> ```

nit: s/Evaluates/evaluates/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94#-:-KZ__3v4aGMpTztE3K3N:b-pdmdcz)*

mandarjog:
I started with a single directory and everything dumped into aspect.

import aspect/listChecker

listChecker.Aspect  and listChecker.Adapter listChecker.Config listChecker.AdapterConfig 
listChecker.Manager   ...

or
import aspect
aspect.ListChecker aspect.ListCheckerAdapter  aspect.ListCheckerConfig aspect.ListCheckerAdapterConfig
aspect.ListCheckerManager

While implementing an aspect, I think the former is better, it gives them all interfaces and structs in the same namespace.
There is uniformity between aspect. There are identically named data structures performing same function.

---

Review status: 31 of 32 files reviewed at latest revision, 30 unresolved discussions, some commit checks failed.

---

*[BUILD.api, line 10 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXy9frxNNrB8dB3PrQ:-KZYBSv7HOdaOwPAtfuQ:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L10)):*

Comment is incorrect.
</blockquote></details>

Done.

---

*[BUILD.api, line 48 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXyIcax8sgDccFV6Xg:-KZYBTlBRIxmivQUfORJ:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L48)):*

Comment is incorrect.
</blockquote></details>

Done.

---

*[pkg/aspect/types.go, line 83 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYXRwZfydkRF4LqCW9:-KZagbyTp6s83FeGank3:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L83)):*

missing the other half?
</blockquote></details>

Done.

---

*[pkg/aspect/types.go, line 32 at r2](https://reviewable.io:443/reviews/istio/mixer/94#-KZZlxOVbpKN592DZQe0:-KZadHF22yMvJhHunwLH:b-896fix) ([raw file](https://github.com/istio/mixer/blob/eac590a7c8aaaa35a91d34dc30febe4afd07020f/pkg/aspect/types.go#L32)):*

extreme nit: probably don't need to comment embedded interfaces.
</blockquote></details>

Done.

---

*[pkg/aspect/types.go, line 22 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_jtHHOtA4F2vMahjd:-KZad-vQauVtaRLcGJzv:b-896fix) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L22)):*

nit: let's move this down to the bottom below the other istio.io imports (with a newline sep).
</blockquote></details>

Done.

---

*[pkg/aspect/types.go, line 47 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_a3752wWhJddlZhds:-KZaeMiq4x7PU2ShRqVY:bvdv0fl) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L47)):*

this does leave us with aspect.Aspect.  Should we move this up a level?  Or, and I already regret suggesting this, use a different name (Instance/Builder/???)?
</blockquote></details>

up a level would be 
import .../pkg

pkg.Aspect and pkg.Adapter. That does solve the cyclic dependency problem. 
though pkg dir in k8s is bare. it has no direct files.

---

*[pkg/aspect/types.go, line 104 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_bUfotLABxg9aNrJ7:-KZahH5PLuvzDiVLd4Z_:b-896fix) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L104)):*

s/previsouly/previously/
</blockquote></details>

Done.

---

*[pkg/aspect/listChecker/manager.go, line 77 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_ekBD43OcmIOLjF8k:-KZahITv17oOolwf90I9:bl7mekr) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/listChecker/manager.go#L77)):*

I'm a bit confused looking at this again. If we are going to hardcode `"Symbol"` in here, what is the advantage of then shoving it back in the Input pb in order to call `CheckList()`?  Why not just have `CheckList(string)` at this point? 

It seems like that if we want to use the proto schema for validation, we could use it to extract the field names, validate that mappings for those fields exist, and then use an internal struct that includes the information (or skip the struct entirely).

Or, if we insist on passing in the proto, just do it directly (skip the Arg struct entirely) -> `CheckList(l *listcheckerpb)`.

What am I missing?
</blockquote></details>

listcheckerpb is the user mappable input.
But that is not the only input that an aspectImpl can get during a call.


accessLog.Args {
      accessLogpb    //  { this contains the mappable stream name }
      AccessLogEntry   // This is defined by the accessLog aspect  *not* as a proto 
                                    // and it is filled in via config
}

aspectX.Args puts these uniformly together.     

Options are 
1. Respective aspectManager makes the choice that is easiest to the impl, 
     so use `CheckList(string)`
2. Continue with uniformity that `aspectX.Args` contains everything you need.
3. define accessLog.Arg  with embedded proto and AccessLogEntry instead of fields.
    embedding proto ensures that we are passing along schema that we have published as
     the mappable input proto.


I like (3). It looks like 2 while writing aspect manager, but the adapterImpl author gets an
unrolled datastructure. 

 

---

*[pkg/aspect/listChecker/types.go, line 34 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_htajEVdOcITEHln4:-KZakTYTdZofiAIGysRo:b-7gnsf2) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/listChecker/types.go#L34)):*

Is this used anywhere else? Does it need to be exported?
</blockquote></details>

Yes. the adapter impl will write CheckList fn that uses *Arg

---

*[pkg/aspect/listChecker/types.go, line 42 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_feo06palpK545idq:-KZakf_Vhkb-FiaqvHi8:bmofl8w) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/listChecker/types.go#L42)):*

Why not just have this be `CheckList(symbol string) (bool, error)`? What advantage does having the pb indirection give?  If we want flexibility, we could still use the Args struct, but I would think we could have the args struct look something like:

```
type Args struct {
   Symbol string
}
```

Or as suggested elsewhere, get rid of Arg altogether and just have `CheckList(l *listcheckerpb) (bool, error)`?
</blockquote></details>

related to earlier comment.

---

*[pkg/aspectregistry/types.go, line 30 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_lsZe-I4OP-7e0UcX:-KZaudrUrIM0m3_ttdVi:b-kt2tkf) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspectregistry/types.go#L30)):*

Here, again, we have stutter: aspectregistry.Registry.  
</blockquote></details>

aspectsupport.Registry  ?
aspectsupport.Manager ?

---

*[pkg/expr/types.go, line 25 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ_kjJnuiMHANcvvrlH:-KZaqgxGwkttmWWJF3Dc:b-896fix) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/expr/types.go#L25)):*

nit: s/Evaluates/evaluates/
</blockquote></details>

Done.

---

*[pkg/aspect/registry/manager.go, line 45 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEtXiA8B8C3pgx0f7:-KZYjEL_zIIbzZu6SKA8:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L45)):*

Combine with previous line
</blockquote></details>

Done.

---

*[pkg/aspect/registry/manager.go, line 50 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYEqgwr4LoTH9H_8gA:-KZYjFY-LfdE8ObBitz5:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L50)):*

Combine with previous line
</blockquote></details>

Done.

---

*[pkg/aspect/registry/manager.go, line 55 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYFH9gCT9asMlcIYMv:-KZYjG3d6tHKZ1tqAiK0:b-896fix) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/registry/manager.go#L55)):*

Combine with previous line
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94#-:-KZaX8V0jZxk-nghKJHx:btjio0g)*

geeknoid:
*[pkg/api/BUILD, line 5 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-m5wSmxiIb3lWznQ:-KZbZYAyLjeuHuAiaTWq:b-y3op3h) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/api/BUILD#L5)):*

Agree that there probably is (we could write our own rules, etc.). But I'd have to experiment for a few hours/days to know for sure.  Maybe duplication is OK for now; clean up later (should we file an issue to track?)
</blockquote></details>

Issue #99, assigned to Doug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

geeknoid:
Review status: 31 of 32 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[BUILD.api, line 51 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXyNGf8aTCkqosGfz3:-KZbYNBgp4P8JMHynslC:bwwyjsy) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L51)):*

importmap isn't needed for "well-known" protos.  but you do need to specify where to find the sources for imports.
</blockquote></details>

This can't be the right way to find these files. This is assuming the structure of the blaze output directories.

---

*[cmd/client/check.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZY-H2GARdXD4yuDGzl:-KZbYtVzUXCgSRAfwQpu:btglpje) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/cmd/client/check.go#L23)):*

As above, let's deal with this in a subsequent PR.
</blockquote></details>

The current path should be istio.io/mixer/api/v1 though, so this change and all other related changes in this PR should be undone.

---

*[pkg/aspect/types.go, line 27 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYCe05SEwRe4CQnUyr:-KZbZwp2lDzsYOJrp0io:bnn2zzu) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspect/types.go#L27)):*

@mandarjog and I looked at k8s codebase and discovered a ton of types.go in packages. So, in general, it seems at least somewhat accepted to use types.go in packages. However, it appears that this is the only file in the aspect pkg. So, in this case, it seems like maybe just aspect.go is  the better choice. Thoughts?
</blockquote></details>

I'm not sure I care too much about how k8s organizes at that level. I just don't think it is good practice. Makes it harder to find/discover things, puts unrelated things together, put related things far apart, etc.

---

*[pkg/aspect/types.go, line 57 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ__SIRVcwpiklWUelp:-KZb_k_B3ol3pBieLH-r:b32v57) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L57)):*

final nits here:  what about just expanding the type names here: Cfg -> Config, Reg -> Registry ?  I think that might match the rest of the code better.
</blockquote></details>

+1

---

*[pkg/aspectregistry/manager.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZYF110dB6CM_Qe6QZo:-KZbaVbVyv_pQ5voE0_B:b-y7bo4q) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/pkg/aspectregistry/manager.go#L58)):*

attribute.Context leads me to believe that they are attributes available in the current request scope (here request==message).  These are the attributes that we expect to see. 
If there is a special set of immutable attributes underlying this set, that we should name it something else. It perhaps does not need to be exported.
attribute.Context    attribute.baseContext. @geeknoid  
</blockquote></details>

I sent out a PR that introduces an Overrider feature that lets you do local mutations on top of the immutable context.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[pkg/aspect/types.go, line 57 at r3](https://reviewable.io:443/reviews/istio/mixer/94#-KZ__SIRVcwpiklWUelp:-KZbiGxt3q095vEBmEE8:b-896fix) ([raw file](https://github.com/istio/mixer/blob/f3b94517fd5cbea41728a53bc8b1c82940deedb2/pkg/aspect/types.go#L57)):*

+1
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

mandarjog:
*[BUILD.api, line 51 at r1](https://reviewable.io:443/reviews/istio/mixer/94#-KZXyNGf8aTCkqosGfz3:-KZblMwQcW6ebpSNMAWX:b-y5i7w2) ([raw file](https://github.com/istio/mixer/blob/9416633b39dad573279d9bc8e3cef9b1d4a1af7c/BUILD.api#L51)):*

This can't be the right way to find these files. This is assuming the structure of the blaze output directories.
</blockquote></details>

created https://github.com/istio/mixer/issues/100

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/94)*

douglas-reid:
Talked through things with @mandarjog in person. It sounded like he had a way to simplify some of the config passing and a plan for the code organization inline with discussion on #mixer.

I'm going to be OOO until Jan. 4, so I'm removing myself from the reviewers list. @geeknoid you have conn.

mandarjog:
superseded by #112

PR #96, opened 2016-12-22
Generic string interning package to help memory & perf.

geeknoid:
We can leverage this in a number of places to keep our memory
footprint in check.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/96)

### DIFF
--- a/pkg/intern/intern.go
+++ b/pkg/intern/intern.go
@@ -1,0 +1,77 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package intern
+
+import (
+	"sync"
+)
+
+// Pool is a general purpose string interning package to reduce memory
+// consumption and improve processor cache efficiency.
+type Pool struct {
+	sync.RWMutex
+	strings     map[string]string
+	currentSize int
+	maxSize     int
+}
+
+const (
+	// just a guess of course...
+	averageStringLength = 10
+)
+
+// NewPool allocates a new interning pool ready for use.
+//
+// Go doesn't currently have sophisticated GC primitives, such as weak pointers.
+// As a result, a simple string interning solution can easily become subject to
+// memory bloating. Strings are only ever added and never removed, leading to
+// an eventual OOM condition. Can easily be leveraged to DDoS a server for example.
+//
+// The code here uses a simple approach to work around this problem. If the table
+// holding the interned strings ever holds more than than maxSize's worth of strings,
+// the table is completely dropped on the floor and a new table is allocated. This
+// allows any stale strings pointed to by the old table to be reclaimed by the GC.
+// This effectively puts a cap on the memory held by any single pool. The cost of
+// this approach of course is that interning will be less efficient.
+func NewPool(maxSize int) *Pool {
+	return &Pool{strings: make(map[string]string, maxSize/averageStringLength), maxSize: maxSize}
+}
+
+// Intern returns a sharable version of the string, allowing the
+// parameter's storage to be garbage collected.
+func (p *Pool) Intern(s string) string {
+	// quick try if its already in the table
+	p.RLock()
+	result, found := p.strings[s]
+	p.RUnlock()
+
+	if !found {
+		// look again under a serializing r/w lock
+		p.Lock()
+		if result, found = p.strings[s]; !found {
+			if len(s) > p.maxSize-p.currentSize {
+				p.strings = make(map[string]string, p.maxSize/averageStringLength)
+				p.currentSize = 0
+			}
+
+			p.strings[s] = s
+			p.currentSize += len(s)
+			result = s
+		}
+		p.Unlock()
+	}
+
+	return result
+}
--- a/pkg/intern/intern_test.go
+++ b/pkg/intern/intern_test.go
@@ -1,0 +1,45 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package intern
+
+import (
+	"testing"
+)
+
+func TestIntern(t *testing.T) {
+	p := NewPool(4)
+
+	strings := []string{
+		"a",
+		"b",
+		"c",
+		"d",
+		"e",
+		"f",
+		"g",
+		"h",
+	}
+
+	// We're only testing the semantics here, we're not actually
+	// verifying that interning has taken place. That's too hard to
+	// do in Go...
+
+	for _, s := range strings {
+		r := p.Intern(s)
+		if r != s {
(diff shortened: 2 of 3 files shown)


douglas-reid:
On mobile, will look in more detail later. Wanted to mention: https://github.com/golang/go/issues/5160 and ask if you had looked at the solution mentioned in that thread (and how this compares).

pkg/intern/intern.go
```diff
@@ -0,0 +1,72 @@
...
+
+// A general purpose string interning package to reduce memory
+// consumption and improve processor cache efficiency.
+package intern
+
+import (
+	"sync"
+)
+
+type Pool struct {
```
douglas-reid: this seems fine.  I guess we could also use sync.Pool (a la the example from the other issue), and not worry about the mutex'ing here. Thoughts?

douglas-reid: We should probably document Pool, as it is exported.

pkg/intern/intern.go
```diff
@@ -0,0 +1,72 @@
...
+	sync.RWMutex
+	strings  map[string]string
+	maxCount int
```
douglas-reid: maxCount is one way to limit the pool. would we ever consider a size-defined limit?

pkg/intern/intern.go
geeknoid: The solution mentioned has the unfortunate effect of zapping the table when the GC decides it needs more memory, rather than on a specific size limit. This means the table can potentially get very big indeed. When multiple components use a similar technique in a process, it means we don't get to decide what feature needs more memory compared to another. 

So I like my approach better...

geeknoid: Done.

We really need to get lint checking working in the build...

pkg/intern/intern.go
geeknoid: Excellent idea, switched to maxSize instead.

geeknoid:
PTAL

pkg/intern/intern_test.go
```diff
@@ -0,0 +1,45 @@
...
+		"d",
+		"e",
+		"f",
+		"g",
+		"h",
+	}
+
+	// We're only testing the semantics here, we're not actually
+	// verifying that interning has taken place. That's too hard to
+	// do in Go...
```
mandarjog: You can actually write a test that checks for the purpose.
Here is how I have done it in the prototype.

mixologist/cp/whitelist/whitelist_test.go#67

```go
import (
   g "github.com/onsi/gomega"
)
func TestWhiteListUnload(t *testing.T) {
	g.RegisterTestingT(t)
	wl, err := build("")
	g.Expect(err).To(g.BeNil())
	var finalized atomic.Value
	finalized.Store(false)
	// check adapter is eventually unloaded
	// This test ensures that after unload is called
	// the adapter is removed from memory
	runtime.SetFinalizer(wl, func(ff interface{}) {
		finalized.Store(true)
	})
	wl.Unload()
	runtime.GC()
	g.Eventually(func() bool {
		runtime.GC()
		return finalized.Load().(bool)
	}).Should(g.BeTrue(), "Adapter was not fully unloaded")
}
```

geeknoid: I played with that stuff and it unfortunately doesn't work for strings.

The point here would be to know whether a string gets collected or not. Unfortunately, calling SetFinalize on a string gives a runtime error because a string is a value types and not a pointer. I'd really need to be passing the underlying uintptr which I don't have access to.

Anyway, the library code is so simple, I'm pretty sure it's behaving as it should.

douglas-reid:
:LGTM: This LGTM. There may be some clean up on the comments, but otherwise, this seems ok.

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/intern/intern.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/96#-KZZB1jCV0S50riv7HHI-r1-23:-KZbu60QctCmaoEj8Jxw:bb2ywv1) ([raw file](https://github.com/istio/mixer/blob/0d5ee74e40aaf84641924b3f0a1a1772a138dfe9/pkg/intern/intern.go#L23)):*

Done.

We really need to get lint checking working in the build...

</blockquote></details>

Yeah, it is strange that golint didn't generate any output for this in the build for this PR.  Did running it yourself generate any issues?

---

*[pkg/intern/intern.go, line 33 at r2](https://reviewable.io:443/reviews/istio/mixer/96#-KZbwvPyN1ZxMt1wPf4p:-KZbwvPyN1ZxMt1wPf4q:bosynbh) ([raw file](https://github.com/istio/mixer/blob/bc8a11062d5176f1f514216c29835215e4e56bac/pkg/intern/intern.go#L33)):*
> ```Go
> const (
> 	// just a guess of course...
> 	averageStringLength = 10
> ```

nit: elsewhere, a comment reads maxSize/16 where the code is maxSize/averageStringLength. Should this be 16, or should that comment be updated?

---

*[pkg/intern/intern.go, line 50 at r2](https://reviewable.io:443/reviews/istio/mixer/96#-KZbxHYsccetGcHLwA2x:-KZbxHYti_HBYJqMRN8r:bbd244u) ([raw file](https://github.com/istio/mixer/blob/bc8a11062d5176f1f514216c29835215e4e56bac/pkg/intern/intern.go#L50)):*
> ```Go
> // this approach of course is that interning will be less efficient.
> func NewPool(maxSize int) *Pool {
> 	// We use maxSize/16 to preallocate space in the map which assumes an average string length of 8 characters
> ```

i think this comment needs updating based on the const "averageStringLength"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/96#-:-KZbxWFh7xF9OD-m4i_Q:b-m275iv)*

geeknoid:
Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/intern/intern.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/96#-KZZB1jCV0S50riv7HHI-r1-23:-KZc1Bh5Y-Uwz0gfkQTC:b-896fix) ([raw file](https://github.com/istio/mixer/blob/0d5ee74e40aaf84641924b3f0a1a1772a138dfe9/pkg/intern/intern.go#L23)):*

Yeah, it is strange that golint didn't generate any output for this in the build for this PR.  Did running it yourself generate any issues?
</blockquote></details>

Done.

---

*[pkg/intern/intern.go, line 26 at r1](https://reviewable.io:443/reviews/istio/mixer/96#-KZZB1jCV0S50riv7HHI-r1-26:-KZc1FDP2xPKhF-rL-ij:b-896fix) ([raw file](https://github.com/istio/mixer/blob/0d5ee74e40aaf84641924b3f0a1a1772a138dfe9/pkg/intern/intern.go#L26)):*

Excellent idea, switched to maxSize instead.
</blockquote></details>

Done.

---

*[pkg/intern/intern.go, line 33 at r2](https://reviewable.io:443/reviews/istio/mixer/96#-KZbwvPyN1ZxMt1wPf4p:-KZc1H8qKFuPdyh3GluI:b-896fix) ([raw file](https://github.com/istio/mixer/blob/bc8a11062d5176f1f514216c29835215e4e56bac/pkg/intern/intern.go#L33)):*

nit: elsewhere, a comment reads maxSize/16 where the code is maxSize/averageStringLength. Should this be 16, or should that comment be updated?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/96)*

douglas-reid:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/96)*

geeknoid:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/intern/intern.go, line 50 at r2](https://reviewable.io:443/reviews/istio/mixer/96#-KZbxHYsccetGcHLwA2x:-KZc2K9JdXaSaBVLZQtX:b-896fix) ([raw file](https://github.com/istio/mixer/blob/bc8a11062d5176f1f514216c29835215e4e56bac/pkg/intern/intern.go#L50)):*

i think this comment needs updating based on the const "averageStringLength"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/96)*

mandarjog:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/96#-:-KZc2LAkU0j__h1l4Hg1:bnfp4nl)*

PR #98, opened 2016-12-22
Add attribute.Bag to allow mixer-based mutation of attributes.

geeknoid:
---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/98)

### DIFF
--- a/pkg/attribute/BUILD
+++ b/pkg/attribute/BUILD
@@ -20,6 +20,7 @@ DEPS = [
 go_library(
     name = "go_default_library",
     srcs = [
+        "bag.go",
         "context.go",
         "dictionaries.go",
         "manager.go",
@@ -42,6 +43,14 @@ go_test(
     name = "manager_test",
     size = "small",
     srcs = ["manager_test.go"],
+    library = ":go_default_library",
+    deps = DEPS,
+)
+
+go_test(
+    name = "bag_test",
+    size = "small",
+    srcs = ["bag_test.go"],
     library = ":go_default_library",
     deps = DEPS,
 )
--- a/pkg/attribute/bag.go
+++ b/pkg/attribute/bag.go
@@ -1,0 +1,167 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attribute
+
+import (
+	"time"
+)
+
+// Bag is a generic mechanism to track a set of attributes.
+type Bag interface {
+	// String returns the named attribute if it exists.
+	String(name string) (string, bool)
+
+	// SetString sets an override for a named attribute.
+	SetString(name string, value string)
+
+	// Int64 returns the named attribute if it exists.
+	Int64(name string) (int64, bool)
+
+	// SetInt64 sets an override for a named attribute.
+	SetInt64(name string, value int64)
+
+	// Float64 returns the named attribute if it exists.
+	Float64(name string) (float64, bool)
+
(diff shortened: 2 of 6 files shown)


mandarjog:
*[pkg/attribute/overrider.go, line 23 at r1](https://reviewable.io:443/reviews/istio/mixer/98#-KZbZxb64a1fXIhdxJT1:-KZbZxb64a1fXIhdxJT2:bmhodvc) ([raw file](https://github.com/istio/mixer/blob/9f86be873ef8dd5b322f9eaa877b863881f3bea8/pkg/attribute/overrider.go#L23)):*
> ```Go
> // Overrider is a generic mechanism to locally override attribute
> // valuesvfrom a context, without mutating the underlying context.
> type Overrider interface {
> ```

I think mutable is a more appropriate name.
```go
type MutableContext interface {
SetXX fns

Context
}
```

In my view there is chain of Contexts, the last one of which is mutable.
There is a 
- global attribute context
- per input stream attribute context
- per message attribute context

At this point request processing starts.

Here we *want* mutable context because aspectImpls may be ordered and one may depend on certain attributes to be present that were injected earlier.

The adapter managers should be passed in mutableContext. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/98)*

geeknoid:
Updated the model. The immutable Context interface is gone, replaced with the mutable Bag interface. The context is still maintained internally, it represents the exact set of attributes exchanged from the proxy to the mixer. The Bag instead represents the mixer's ongoing mutation to those attributes as they travel down different pipelines.

mandarjog:
*[pkg/attribute/bag.go, line 63 at r2](https://reviewable.io:443/reviews/istio/mixer/98#-KZbpCwCN1h93ajUQUt_:-KZbpCwCN1h93ajUQUta:b-900xxh) ([raw file](https://github.com/istio/mixer/blob/184517a8a28ab48c77cc039b388d84fa44385da2/pkg/attribute/bag.go#L63)):*
> ```Go
> }
> 
> type bag struct {
> ```

 
1. At present there is 2 step process of chained contexts.
    But there is at least 1 more.  Why not model it as a chain.

   a. attributes established during connection phase at don't change. (like the identity of the caller)   {we are missing this }
    b. The exact attribute snapshot that was sent by proxy to mixer   { context }
    c.  attributes added during request processing by adapters  { bag }
2. why the word 'bag' instead of 'context'.      I think the earlier name conveyed the concept better.

```go
Context interface {
    Mutable
    Immutable
    
}

context struct {
   Mutable
   Immutable
   parent Immutable
}

func (c *context) String(name string) (string, bool) {
	var r string
	var b bool
     
	if r, b = c.strings[name]; !b {
                if c.parent != nil {                            
		   r, b = c.parent.String(name)
                }
	}
	return r, b
}

func newContext(parent Immutable) {
     return &context{ parent: parent}
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/98)*

geeknoid:
Our current API/protocol has no notion of global attributes and I don't
think it's necessary. Every gRPC stream is orthogonal and even invariants
such as client name are sent as part of the normal communication exchange
between the proxy and server. I therefore don't see any case where we
actually would benefit from a more complex chain solution. I don't want to
build special purpose support for these global attributes since there is no
tangible benefit (more moving pieces, no real perf impact)

As for the name, it would like I can rename Context<->Bag if you'd rather
have the external package interface be called a Context. I don't see the
benefit of one over the other personally.

geeknoid:
Regarding "context" vs. "bag", I forgot to mention about the API surface.

The API surface defines the notion of attribute context. The proxy can pick different contexts to interact with when it talks to the mixer. These contexts are only updated as directed by the proxy, the mixer needs to consider those immutable in order for the proxy->mixer protocol to work.

So within the mixer, I want to use the word "context" to match with the protocol/API surface. 

Now, instead of attribute.Bag, we could go with attribute.MutableContext but that's just longer and doesn't seem particularly more mnemonic to me.

douglas-reid:
I still feel like we should be using the standard context constructs here. The WithValue() stuff should provide the appropriate overriding, etc. as needed.  What do we get out of providing our re-implementation of it (other than the myriad typed methods)?

douglas-reid:
Removing myself from reviewers list, as I'll be OOO for a little while.

It sounds like we will be passing Contexts around now in the code (which should help with cancels and tracing) -- and that we may attach Bag or whatever to that mechanism, which seems fine to me.

@mandarjog you have the conn.

PR #101, opened 2016-12-22
Add repositories.bzl to remove boilerplate in BUILD files

### LINKED ISSUE #99: Fix boilerplate goo in BUILD files for local istio_api usage

lizan:
Closes #99

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/101)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -120,16 +120,7 @@ new_go_repository(
     importpath = "github.com/spf13/pflag",
 )
 
-new_git_repository(
-    name = "com_github_istio_api",
-    build_file = "BUILD.api",
-    commit = "1d9417f607be5503eee95fdb109c0d906fe6b5f5",
-    remote = "https://github.com/istio/api.git",
-)
+load("//:repositories.bzl", "istio_api_repositories")
 
-# use with --define istio=local invocation
-new_local_repository(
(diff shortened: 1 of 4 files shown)


mandarjog:
@lizan
1. This makes you update the workspace file locally; which I was trying to avoid.
2. I don't like the literal string as buildfile;  I have an alternative that is equivalent from a user's standpoint, but does not do the literal string.

lizan:
1. It is the minimal change to avoid changing each BUILD file, another
approach is add a load statement to each BUILD. go_library doesn't work
without any source, so a dummy go_library target didn't work.

2. What is the equivalent? It is the easiest way to avoid copying BUILD
files around, especially when we have a repo (likely istio/istio) depends
on mixer, you don't have to copy BUILD for mixer dependencies to that repo.

kyessenov:
So is the issue here you want to avoid pushing changes to api repo before trying them out in mixer?

Let's use gazella to automate BUILD file generations. Then if you use go tool, you can directly edit $GOROOT/.../api repo. If you use bazel, then there is a repo in /external/ that you can edit and Bazel accepts. What am I missing?

douglas-reid:
iiuc, the issue is not wanting to edit the repo in /external/.

PR #103, opened 2016-12-23
Rejigger the attribute code to support mixer-side attribute mutation.

geeknoid:
The attribute.Context interface is replaced with attribute.MutableBag and
attribute.Bag. These interfaces provide a chaining/overriding mechanism
that the mixer uses to manipulate attributes wihtin its pipeline.

Start plumbing through the use of a Go Context through the layers. It's now
only at the topmost layer but will need to be driven through all the layers.
The bag of attributes for a request is attached to the context so it travels
with the request down to the lower layers.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/103)


geeknoid:
PTAL

pkg/api/grpcServer.go
```diff
@@ -177,7 +183,7 @@ func (s *GRPCServer) Report(stream mixerpb.Mixer_ReportServer) error {
 		new(mixerpb.ReportRequest),
 		new(mixerpb.ReportResponse),
 		func(tracker attribute.Tracker, request proto.Message, response proto.Message) {
-			s.handlers.Report(tracker, request.(*mixerpb.ReportRequest), response.(*mixerpb.ReportResponse))
+			s.handlers.Report(context.Background(), tracker, request.(*mixerpb.ReportRequest), response.(*mixerpb.ReportResponse))
```
geeknoid: Yep.

PR #107, opened 2016-12-24
Optimize the attribute package.

geeknoid:
- Create an emptyBag type to use at the root of the chain
of bags to avoid conditionals throughout the code.

- Instead of clearing maps to nil on reset, drain them instead
so that we avoid generating a lot of GC garbage.

- Use a pool to recycle bags across requests. This works with the
above map improvement to effectively bring to 0 the GC impact of
steady state usage of the attribute package.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/107)


mandarjog:
@jba I think the assertion is that we are not adding more work to the GC. If other generated code is allocating something in a loop or downstream library code is doing the same, we cannot help that.

That said, it remains to be benchmarked. @geeknoid  ?

geeknoid:
Indeed, I only imply that the attribute package itself isn't generating more garbage. The fact our libraries do is a whole other issue. As we push the limits of perf, I'm sure we'll eventually want to optimize our dependencies, including the generated gRPC code. But we're nowhere near doing that level of optimization.

As for a benchmark, I think that's a great idea. I'll open an issue for this.

jba:
Next question: why the emphasis on GC? Does that come from experience with kubernetes or other projects? Have you tried the new low-latency collector in Go 1.8?

Since the Attributes proto will be garbage anyway, you could define a bag to be triple of Attributes proto, dictionary, and parent, and do the int-to-string key lookup lazily. That would simplify and speed up the StartRequest/EndRequest implementation and clearly not generate any additional garbage. (You'd null out the Attributes field when you put the bag back in the pool, but that proto was garbage anyway.)

geeknoid:
The emphasis on GC comes from several years of experience working on a high-performance garbage collected system. At the end of the day, in an I/O intensive system, you're almost always going to eventually hit a GC bottleneck.

Concurrent garbage collection is nice to reduce GC pauses and that's certainly a bonus. However, it doesn't reduce the actual CPU cost incurred. Some cycles somewhere are still being spent running the collection. In fact, doing it concurrently typically ends up costing you more CPU time than if you paused.

Not only do you pay cycles for the collection process itself. But generating a lot of garbage means the code is chewing through memory at a good clip. This inherently means poorer CPU cache and TLB locality. So not only do you incur the work of the collector, your other code ends up running slower too.

If a system is designed with the belief that GC is free, it can be very painful to fix at a later point. You don't necessarily have the right control points in the system to implement a reuse policy for example. For our situation, if our interface to the adapters are allocation-heavy, we'll have a hard time to fix that over time.

As for your suggestion about improving the code, I'll take a look. But from the face of it, I don't think this sounds like a real improvement. In reality, the dictionary is mostly going to be constant for the life of the tracker, and moreover is likely to be a singleton for whole mixer. In that context, it is preferable to front-load any processing on the dictionary such that the mainline code can run unencumbered with any conditional checks. 

Tnanks.

geeknoid:
Deferring the int->string conversion would probably end up as a next negative.

As it is, the bag maintains maps of string->data. This makes local access and mutation of the bag natural with no fuss (since the APIs like String() and SetString() deal in terms of strings). If I were to defer the int->string mapping, that would imply I internally keep int->data maps instead. This would lead to mapping occurring in the String() and SetString()-like methods. Since we will likely be reading/writing a lot of these attributes within the mixer, this ends up not being a good trade-offs relative to just converting upfront as the data arrives.

jba:
Thanks for the detailed explanation. I'll defer to your experience on the importance of minimizing GC.

As you say, if you don't design with allocation in mind, it can be hard to fix later on. So let me make another suggestion: the Attributes message is inherently allocation-heavy, so you might think about changing that, since it will be hard to change later, when the proxy - mixer API is locked down. (Maybe there are hooks in gRPC or protoc that let you introduce memory optimizations; if so, ignore the following.)

If you changed from maps to pairs of repeated fields:
```
repeated int32 string_keys
repeated string string_vals
```
then the wire format would be the same but you'd save the allocation of the maps (which is probably not significant, since you keep all the values in the bag so only the map data structure itself is garbage). More radically, you could represent the string map as
```
repeated int32 string_keys
repeated int32 starts
repeated int32 ends
string string_data
```
where the i'th string is `data[starts[i]: ends[i]]`. That would involve only four allocations.

geeknoid:
Excellent ideas, thank you. This would at least eliminate the map objects and the hash tables, and all the overhead of having and map insertion. First thing we do on the mixer side is to iterate through this data and insert it into our own map, so we're actually not using the fact the data is being sent as a map.

I'd like to wait for this until we actually have an idea of what the proxy-side code for this protocol will look look and have some end-to-end benchmarks for us to play with. I've opened issue #111 to track this.

In the broader context, I'm hoping that as gRPC matures further, provisions will be introduced for object reuse. Instead or reading from a stream and getting some object, I'd like to pass in the object and have it filled in.

Thanks.

geeknoid:
PTAL.

The code typically had 3 nodes in a bag chain: the mutable bag, the bag carrying the protocol-level attributes, and an empty sentinel bag. I've now collapsed the last two into a single bag which eliminates one conditional in the case of reading data present in the protocol level bag, and one function call in the case of lookups for attributes not in the bag chain at all.

I also refactored things a bit such that the interface definitions are separated from the two implementations.

pkg/attribute/rootBag.go
```diff
@@ -0,0 +1,224 @@
...
+// attribute called 'source.ip' which is of type IP_ADDRESS could be listed as
+// a string or an int, and we wouldn't catch it here.
+func checkPreconditions(dictionary dictionary, attrs *mixerpb.Attributes) error {
+	for k := range attrs.StringAttributes {
+		if _, present := dictionary[k]; !present {
+			return fmt.Errorf("attribute index %d is not defined in the current dictionary", k)
+		}
+	}
+
+	for k := range attrs.Int64Attributes {
```
mandarjog: could we return https://github.com/hashicorp/go-multierror  here?
Would be good to see all errors at one time.

pkg/api/methodHandlers.go
```diff
@@ -231,12 +233,13 @@ func buildLogEntries(entries []*mixerpb.LogEntry) []adapters.LogEntry {
 */
 
 func (h *methodHandlers) Quota(ctx context.Context, tracker attribute.Tracker, request *mixerpb.QuotaRequest, response *mixerpb.QuotaResponse) {
-	ab, err := tracker.Update(request.AttributeUpdate)
+	ab, err := tracker.StartRequest(request.AttributeUpdate)
```
geeknoid: Done

pkg/api/grpcServer.go
```diff
@@ -143,7 +143,9 @@ func (s *GRPCServer) Stop() {
 type handlerFunc func(tracker attribute.Tracker, request proto.Message, response proto.Message)
 
 func (s *GRPCServer) streamLoop(stream grpc.ServerStream, request proto.Message, response proto.Message, handler handlerFunc) error {
-	tracker := s.attrMgr.NewTracker()
+	tracker := s.attrMgr.GetTracker()
```
jba: The usual Go style is to avoid the word "Get" if it doesn't add value. I'd just call the function "Tracker".

pkg/attribute/bag.go
```diff
@@ -46,6 +40,9 @@ type Bag interface {
 
 	// Bytes returns the named attribute if it exists.
 	Bytes(name string) ([]uint8, bool)
+
+	// Close indicates the bag can be reclaimed.
+	Close()
```
jba: A method named Close should return error, by strong convention. I would suggest another name, like Done.

pkg/attribute/rootBag.go
```diff
@@ -0,0 +1,224 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/golang/protobuf/ptypes"
+
+	mixerpb "istio.io/api/mixer/api/v1"
+)
+
+// The rootBag is always the base of a chain of bags. It's state is updated
```
jba: Its

pkg/attribute/rootBag.go
```diff
@@ -0,0 +1,224 @@
...
+// only via an Attributes proto. This code is not protected by locks because
+// we know it is always mutated in a single-threaded context.
+type rootBag struct {
```
jba: This seems like a lot of code for a performance improvement that may not be significant or necessary. A null check is negligible compared to a map lookup, so I wouldn't think testing for ab.parent == nil would be noticeable.

pkg/attribute/rootBag.go
geeknoid: Done

pkg/attribute/tracker.go
```diff
@@ -24,51 +26,93 @@ import (
...
-	Update(attrs *mixerpb.Attributes) (MutableBag, error)
+	// supplied Attributes proto. When this happens, none of the tracked attribute
+	// state will have been affected.
+	StartRequest(attrs *mixerpb.Attributes) (MutableBag, error)
+
+	// EndRequest performs post-request cleanup
+	EndRequest()
+
+	// Close indicates the tracker can be reclaimed.
+	Close()
```
jba: See my earlier comment on Close.

geeknoid:
PTAL

pkg/attribute/rootBag.go
geeknoid: This is actually not a significant amount of extra code compared to the previous iteration, if you discount the copyright message and import statements. I mostly moved code that was in the mutablebag.go source to rootBag.go. My main motivation for the change was overall functional clarity, the micro perf boost is just a bonus.

pkg/attribute/bag.go
geeknoid: Renamed throughout.

pkg/api/methodHandlers.go
```diff
@@ -70,41 +72,46 @@ func NewMethodHandlers(adapterManager *server.AdapterManager, configManager *ser
...
-	allowed, err := h.checkLists(dispatchKey, tracker)
-	if err != nil {
-		glog.Warningf("Unexpected check error. dispatchKey: '%v', error: '%v'", dispatchKey, err)
-		response.Result = newStatus(code.Code_INTERNAL)
-		return
-	}
+func (h *methodHandlers) Check(ctx context.Context, tracker attribute.Tracker, request *mixerpb.CheckRequest, response *mixerpb.CheckResponse) {
+	response.RequestIndex = request.RequestIndex
+	response.Result = wrapper(ctx, tracker, request.AttributeUpdate,
+		func(ctx context.Context, ab attribute.MutableBag) *status.Status {
```
geeknoid: I don't understand what you're suggesting. As it is, the callback for the wrapper function is provided inline. I could move the code to a separate function but that would just add lines with no real value.

As written, we don't currently close over any state from the parent.

geeknoid: I pulled out the logic into separate functions.

PR #112, opened 2016-12-28
Addition of Aspect / Adapter / Aspect Manager

mandarjog:
This PR adds Aspect, Adapter and Aspect Manager.

- It adds the base types and interfaces
- designates the aspectmanager as the uniform interface to aspect.

Aspect managers
- build a specific aspects
- Execute an aspect instance by mapping attributes --> input and other aspect specific business logic.

continuation of abandoned pr #94 

adapter/ipListChecker is converted to new style aspects.

TODO:
1. `proto` definitions for `ipListChecker` adapter is checked in along with `.pb.go` file.
     Change build not potentially not include .pb.go.
2. Connections from server to Aspect Manager has not yet been established.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/112)


adapter/ipListChecker/aspect.go
```diff
@@ -27,62 +27,21 @@ import (
...
-	a.setList([]*net.IPNet{})
-
-	// load up the list synchronously so we're ready to accept traffic immediately
-	a.refreshList()
-
-	// crank up the async list refresher
-	go a.listRefresher()
-
-	return &a, nil
+func (a *aspectState) Name() string {
```
geeknoid: Given the adapter has a name, why does the aspect need a name method as well?

adapter/ipListChecker/aspect.go
```diff
@@ -99,9 +58,9 @@ func (a *aspectState) CheckList(symbol string) (bool, error) {
 
 	// get an atomic snapshot of the current list
 	l := a.getList()
-	if l == nil {
+	if l == nil || len(l) == 0 {
 		// TODO: would be nice to return the last I/O error received from the provider...
```
geeknoid: The TODO is no longer relevant, right?

adapter/ipListChecker/cfg.proto
```diff
@@ -0,0 +1,32 @@
+// Copyright 2016 Google Inc.
```
geeknoid: So why do we need to have these things expressed in protos? Seems awfully painful. Can't we just have 'em all defined in Go and use reflection?

adapter/ipListChecker/cfg.proto
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package ipListChecker;
+
+message Config {
+  // ProviderURL is the URL where to find the list to check against
```
geeknoid: Should be using the names as expressed in the proto, here and elsewhere

adapter/ipListChecker/cfg.proto
```diff
@@ -0,0 +1,32 @@
...
+	string provider_url = 1;
+
+	// RefreshInterval determines how often the provider is polled for
+	// an updated list in seconds
+	int64 refresh_interval = 2;
```
geeknoid: Using the Duration type?

At the very least, indicate the unit (seconds, nanos?)

adapter/ipListChecker/cfg.proto
```diff
@@ -0,0 +1,32 @@
...
+
+	// TimeToLive indicates how long to keep a list before discarding it.
+	// Typically, the TTL value should be set to noticeably longer (> 2x) than the
+	// refresh interval to ensure continued operation in the face of transient
+	// server outages.
+	int64 ttl = 3;
```
geeknoid: Same comment as for the RefreshInterval above.

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	return &manager{}
+}
+
+// NewAspect creates a listChecker aspect
+func (m *manager) NewAspect(cfg *aspect.CombinedConfig, ga aspect.Adapter) (aspect.Aspect, error) {
+	aa, ok := ga.(Adapter)
+	if !ok {
+		return nil, fmt.Errorf("Adapter of incorrect type. Expected listChecker.Adapter got %#v %T", ga, ga)
+	}
+	_, ok = cfg.Aspect.TypedParams.(*listcheckerpb.Config)
```
geeknoid: How can this happen? Is this an operator-level error?

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := aa.ValidateConfig(implcfg); err != nil {
+		return nil, err
+	}
+	return aa.NewAspect(
+		&AdapterConfig{
+			aspect.ImplConfig{Message: implcfg},
+		})
+}
+
+// Execute performs the aspect function based on given Cfg and AdapterCfg and attributes
```
geeknoid: Comment needs updating.

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	var acfg *listcheckerpb.Config
+
+	// check if mapping is available
+	if acfg, found = cfg.Aspect.TypedParams.(*listcheckerpb.Config); !found {
+		return nil, fmt.Errorf("Unexpected type %#v", cfg.Aspect.TypedParams)
+	}
+
+	// CheckAttribute should be processed and sent to input
+	if symbolExpr, found = cfg.Aspect.Inputs[acfg.CheckAttribute]; !found {
+		return nil, fmt.Errorf("Mapping for %s not found", symbolExpr)
```
geeknoid: symbolExpr -> acfg.CheckAttribute

pkg/aspect/types.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/expr"
+
+	istiopb "istio.io/api/istio/config/v1"
+)
+
+type (
+	// Adapter represents a factory to create an adapterImpl that provides a specific aspect
+	// This interface is extended by specific aspects
```
geeknoid: specific aspects, or specific adapters?

pkg/aspect/types.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2016 Google Inc.
```
geeknoid: Can we split this up? types.go is an anti-pattern IMO.

pkg/aspect/types.go
```diff
@@ -0,0 +1,104 @@
...
+	Aspect interface {
+		io.Closer
+		// Name returns the official name of the aspect
+		Name() string
+	}
+
+	// Config is the internal struct used for the Aspect proto
+	// at present it only has an additional field that stores the converted
+	// and typed proto
+	Config struct {
```
geeknoid: WHy is this "Config" while the other type is "AspectConfig". Can we also use AdapterConfig for consistency?

pkg/aspect/types.go
```diff
@@ -0,0 +1,104 @@
...
+		istiopb.Aspect
+		// TypedParams points to the proto after google_protobuf.Struct is converted
+		TypedParams proto.Message
+	}
+
+	// AdapterConfig is the internal struct used for the Adapter proto
+	// at present it only has an additional field that stores the converted
+	// and typed Args
+	AdapterConfig struct {
```
geeknoid: This is awful confusing.

There's an Adapter type defined here. But that's different that istiopb.Adapter. And AdapterConfig is logical extension to istiopb.Adapter. I think we need renames...

pkg/aspect/types.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	// CombinedConfig combines all configuration related to an aspect
+	CombinedConfig struct {
+		Aspect  *Config
+		Adapter *AdapterConfig
+	}
+
+	// ImplConfig provides common configuration params
+	// Impl config protobufs are required to support these
+	ImplConfig struct {
```
geeknoid: Why isn't this just a "base class" for AdapterConfig?

pkg/aspectsupport/manager.go
```diff
@@ -0,0 +1,88 @@
...
+	"istio.io/mixer/pkg/aspect"
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/expr"
+)
+
+// NewManager Creates a new Uber Aspect manager
+// provides an Execute function to act on a aspect/adapter
+// in a given context
+func NewManager(areg Registry) *Manager {
+	// Add all manager here as new aspects are added
```
geeknoid: manager -> managers

pkg/aspectsupport/manager.go
```diff
@@ -0,0 +1,88 @@
...
+		mreg[mgr.Kind()] = mgr
+	}
+	return &Manager{
+		mreg:        mreg,
+		aspectCache: make(map[CacheKey]aspect.Aspect),
+		areg:        areg,
+	}
+}
+
+// Execute performs the aspect function based on given Cfg and AdapterCfg and attributes
```
geeknoid: Comment out of date

pkg/aspectsupport/manager.go
```diff
@@ -0,0 +1,88 @@
...
+// returns aspect output or error if the operation could not be performed
+func (m *Manager) Execute(cfg *aspect.CombinedConfig, ctx attribute.Bag, mapper expr.Evaluator) (*aspect.Output, error) {
+	var mgr aspect.Manager
+	var found bool
+
+	if mgr, found = m.mreg[cfg.Aspect.Kind]; !found {
+		return nil, fmt.Errorf("Could not find Mgr %s", cfg.Aspect.Kind)
```
geeknoid: Mgr -> manager

pkg/aspectsupport/types.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2016 Google Inc.
```
geeknoid: Icky types.go again...

pkg/expr/types.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2016 Google Inc.
```
geeknoid: Icky types.go

pkg/expr/types.go
```diff
@@ -0,0 +1,31 @@
...
+import "istio.io/mixer/pkg/attribute"
+
+type (
+	// Evaluator maps Given an expression Language
+	Evaluator interface {
+		// Eval evaluates given expression using the attribute bag
+		Eval(mapExpression string, ctx attribute.Bag) (interface{}, error)
+
+		// Eval evaluates given expression using the attribute bag to a string
+		EvalString(mapExpression string, ctx attribute.Bag) (string, error)
```
geeknoid: Here and elsewhere, I think we should be using some other name instead of "ctx". We'll eventually be plumbing the context.Context stuff through and that should be called "ctx" by convention. So perhaps, the bags should be called "bag" or "attrs".

adapter/denyChecker/adapter_test.go
```diff
@@ -18,14 +18,8 @@ import (
 	"testing"
 
 	"istio.io/mixer/pkg/adapter"
-	"istio.io/mixer/pkg/adaptertesting"
 )
 
-func TestBuilderInvariants(t *testing.T) {
```
mandarjog: Yes, I will add it back.
I wanted to get the structure in.
 
I also think that this is something we can do across the board, so every adapter need not implement it.

adapter/ipListChecker/adapter.go
```diff
@@ -16,73 +16,96 @@ package ipListChecker
...
-	return &adapterState{}
-}
-
 func (a *adapterState) Name() string {
-	return "IPListChecker"
+	return ImplName
 }
 
 func (a *adapterState) Description() string {
 	return "Checks whether an IP address is present in an IP address list."
```
mandarjog: Yes, I think it makes sense.
I am a bit conflicted on imposing so much cognitive overhead on adapter authors though.

adapter/ipListChecker/adapter.go
```diff
@@ -16,73 +16,96 @@ package ipListChecker
...
+	c, ok := cfg.(*Config)
+	if !ok {
+		return fmt.Errorf("Invalid message type %#v", cfg)
+	}
 	var u *url.URL
 
-	if u, err = url.Parse(c.ProviderURL); err == nil {
+	if u, err = url.Parse(c.ProviderUrl); err == nil {
 		if u.Scheme == "" || u.Host == "" {
 			err = errors.New("Scheme and Host cannot be nil")
```
mandarjog: Done

adapter/ipListChecker/adapter.go
```diff
@@ -16,73 +16,96 @@ package ipListChecker
...
+		closing:         make(chan bool),
+		refreshInterval: time.Second * time.Duration(config.RefreshInterval),
+		ttl:             time.Second * time.Duration(config.Ttl),
+	}
+	aa.client = http.Client{Timeout: aa.ttl}
+
+	// install an empty list
+	aa.setList([]*net.IPNet{})
+
+	// load up the list synchronously so we're ready to accept traffic immediately
```
mandarjog: Removed it. 
Perhaps we should add a   Ready() bool ? so synchronous nature is not forced?

adapter/ipListChecker/aspect.go
mandarjog: Done

adapter/ipListChecker/aspect.go
mandarjog: In the aspectCache, we have fully constructed aspect structs. these structs do not hold a reference to the Adapter  (Builder) that they came from. ImplName is that link, which I think is useful in logging.

adapter/ipListChecker/cfg.proto
mandarjog: That is how we had it in the prototype. 
1. Since we (+ adapter authors) are fully bought into defining everything thru protobufs, it forces uniformity.
2. It restricts the types of data structures that can be used for config. proto structs are a subset of what can be expressed in go.

I agree that it is a pain though; if we cannot come up with a very easy proto compile story; then we will revert to go reflection only.

adapter/ipListChecker/cfg.proto
mandarjog: done.

adapter/ipListChecker/cfg.proto
mandarjog: done.

adapter/ipListChecker/cfg.proto
mandarjog: done

pkg/aspect/listChecker/manager.go
mandarjog: This is an internal error.
Mark it as internal; or come up with a new error interface ?

pkg/aspect/listChecker/manager.go
mandarjog: done.

pkg/aspect/listChecker/manager.go
mandarjog: done

pkg/aspect/types.go
mandarjog: done.

pkg/aspect/types.go
mandarjog: At the point of use this is

```
aspect.Config
aspect.AdapterConfig
```

pkg/aspect/types.go
mandarjog: `aspect.AdapterConfig` is the internal version of
`istiopb.Adapter`

and
`aspect.Config` is the internal version of
`istiopb.Aspect`

will renaming `istiopb.Aspect --> istiopb.AspectConfig` and
`istiopb.Adapter --> istiopb.AdapterConfig` help ?

They are already in the `istio.config.v1` package.

or alternatively 
change
`istiopb "istio.io/api/istio/config/v1"`
to
`istioconfig "istio.io/api/istio/config/v1"`

```
	// Config (aspect.Config) is the internal struct used for the Aspect proto
	// at present it only has an additional field that stores the converted
	// and typed proto
	Config struct {
		istioconfig.Aspect
		// TypedParams points to the proto after google_protobuf.Struct is converted
		TypedParams proto.Message
	}
```

pkg/aspect/types.go
mandarjog: I removed it. I will add the standard params as "adapter invariant" checks

pkg/aspectsupport/manager.go
mandarjog: Done.

pkg/aspectsupport/manager.go
mandarjog: Done

pkg/aspectsupport/manager.go
mandarjog: Done

adapter/denyChecker/adapter_test.go
mandarjog: added issue #113

pkg/aspect/listChecker/manager.go
mandarjog: opened issue #114

pkg/expr/types.go
mandarjog: done.

mandarjog:
@geeknoid addressed comments. 
I have not broken out types.go into parts yet. PTAL.

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+type (
+	manager struct{}
+)
+
+// Manager returns "this" aspect Manager
+func Manager() aspect.Manager {
+	return &manager{}
+}
+
+// NewAspect implements aspect.Manager#NewAspect() Creates a listChecker aspect
```
geeknoid: I think you can remove the first sentence of this comment.

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+func (m *manager) NewAspect(cfg *aspect.CombinedConfig, ga aspect.Adapter) (aspect.Aspect, error) {
```
geeknoid: What does the "g" in "ga" stand for?

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	aa, ok := ga.(Adapter)
+	if !ok {
+		return nil, fmt.Errorf("Adapter of incorrect type. Expected listChecker.Adapter got %#v %T", ga, ga)
+	}
+	_, ok = cfg.Aspect.TypedParams.(*listcheckerpb.Config)
```
geeknoid: We have cfg.Adapter.TypedArgs and cfg.Aspect.TypedParams. Should both of these fields have the same named?

pkg/aspect/listChecker/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := aa.ValidateConfig(cfg.Adapter.TypedArgs); err != nil {
+		return nil, err
+	}
+	return aa.NewAspect(
+		&AdapterConfig{
+			ImplConfig: cfg.Adapter.TypedArgs,
+		})
+}
+
+// Execute implements aspect.Manager#Execute()
```
geeknoid: I don't think the golang standard wants comments like these...

pkg/aspect/listChecker/types.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"istio.io/mixer/pkg/aspect"
+	"github.com/golang/protobuf/proto"
+)
+
+type (
+	// AdapterConfig -- Adapter Author visible typed config for
+	// creating new listChecker aspect
+	// This may or may not have listcheckerpb.Config
```
geeknoid: I don't understand this "may or may not" sentence, and I don't see where this config would be in the first place.

pkg/aspect/listChecker/types.go
```diff
@@ -0,0 +1,47 @@
...
+		aspect.Aspect
+		// CheckList verifies whether the given symbol is on the list.
+		CheckList(Symbol string) (bool, error)
+	}
+
+	// Adapter builds the ListChecker Aspect
+	Adapter interface {
+		aspect.Adapter
+		// NewAspect returns a new ListChecker
+		NewAspect(cfg *AdapterConfig) (Aspect, error)
```
geeknoid: This is a bit strange isn't it? I would have expected that to build an Aspect, you must pass an AspectConfig. Here, we're building an Aspect and passing an AdapterConfig.

pkg/aspect/types.go
```diff
@@ -0,0 +1,98 @@
...
+	"google.golang.org/genproto/googleapis/rpc/code"
+
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/expr"
+
+	istioconfig "istio.io/api/istio/config/v1"
+)
+
+type (
+	// Adapter represents a factory to create an adapterImpl that provides a specific aspect
```
geeknoid: I don't think this sentence is quite right: The developer creates an adapter implementation, not a factory method. Perhaps:

Adapter represents the factory responsible for returning instantiations of particular aspects.

pkg/aspect/types.go
```diff
@@ -0,0 +1,98 @@
...
+	// This interface is extended by adapter implementations
```
geeknoid: Is this comment right?

At the very least, I think you mean that "this interface is extended for individual aspects".

pkg/aspect/types.go
```diff
@@ -0,0 +1,98 @@
...
+		Description() string
+		// DefaultConfig returns a default configuration struct for this
+		// adapter Implmentation. This will be used by the configuration system to establish
+		// the shape of the block of configuration state passed to the NewAspect method.
+		DefaultConfig() (implConfig proto.Message)
+		// ValidateConfig determines whether the given configuration meets all correctness requirements.
+		ValidateConfig(implConfig proto.Message) error
+	}
+
+	// Aspect -- User visible cross cutting concern
```
geeknoid: Perhaps a beefier comment is in order here? Worth mentioning this is extended for individual aspects.

pkg/aspect/types.go
```diff
@@ -0,0 +1,98 @@
...
+	Aspect interface {
+		io.Closer
+		// Name returns the official name of the aspect
+		Name() string
```
geeknoid: Let's call this AdapterName to indicate it's actually the name of the adapter to which this aspect is tied. Otherwise, folks will be inclined to create unique names for each aspect instance.

pkg/aspect/types.go
```diff
@@ -0,0 +1,98 @@
...
+		TypedParams proto.Message
+	}
+
+	// AdapterConfig (aspect.AdapterConfig) is the internal struct used for the Adapter proto
+	// at present it only has an additional field that stores the converted
+	// and typed Args
+	AdapterConfig struct {
+		istioconfig.Adapter
+		// TypedParams points to the proto after google_protobuf.Struct is converted
+		TypedArgs proto.Message
```
geeknoid: Rename to TypedParams.

pkg/aspect/listChecker/types.go
```diff
@@ -0,0 +1,47 @@
...
+
+type (
+	// AdapterConfig -- Adapter Author visible typed config for
+	// creating new listChecker aspect
+	// This may or may not have listcheckerpb.Config
+	// Aspect manager handles as much of this config
+	// as it makes sense. Rest of the struct is packaged
+	// and sent down to NewAspect()
+	// It will always have ImplConfig if one is present
+	AdapterConfig struct {
```
geeknoid: I'm still confused relative to aspect.Config and this AdapterConfig struct.

Too many config types!

adapter/ipListChecker/adapter.go
mandarjog: yes we should, until someone complains.

adapter/ipListChecker/cfg.proto
```diff
@@ -0,0 +1,33 @@
...
+
+    // refresh_interval determines how often the provider is polled for
+    // an updated list unit: seconds
+    int64 refresh_interval = 2;
+
+    // TimeToLive indicates how long to keep a list before discarding it.
+    // Typically, the TTL value should be set to noticeably longer (> 2x) than the
+    // refresh interval to ensure continued operation in the face of transient
+    // server outages.
+    // unit: seconds
```
mandarjog: I don't think we need nanos here. It is more natural to express time as seconds (or another unit) in a json or yaml file.

pkg/aspect/types.go
mandarjog: updated.

pkg/aspect/types.go
mandarjog: renamed to TypedArgs

pkg/aspect/types.go
mandarjog: done.

pkg/aspect/types.go
mandarjog: called it ImplName.

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
-	"istio.io/mixer/pkg/adapter"
+	"github.com/golang/protobuf/proto"
+
+	"istio.io/mixer/pkg/aspect/listChecker"
+	"istio.io/mixer/pkg/aspectsupport"
+)
+
+const (
+	// ImplName is the canonical name of this implementation
+	ImplName = "istio/IPListChecker"
```
geeknoid: AdapterName seems better overall that ImplName.

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
+		return nil, fmt.Errorf("Config not found %#v", c.ImplConfig)
+	}
 	var u *url.URL
-
-	if u, err = url.Parse(c.ProviderURL); err == nil {
-		if u.Scheme == "" || u.Host == "" {
-			err = errors.New("Scheme and Host cannot be nil")
-		}
+	var err error
+	if u, err = url.Parse(config.ProviderUrl); err != nil {
```
geeknoid: This error is handled during ValidateConfig, right?

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
+func (a *adapterState) NewAspect(c *listChecker.Config) (listChecker.Aspect, error) {
+	if err := a.ValidateConfig(c.ImplConfig); err != nil {
+		return nil, err
 	}
-}
 
-func (a *adapterState) ValidateAspectConfig(config adapter.AspectConfig) error {
-	c := config.(*AspectConfig)
-	var err error
+	config, found := c.ImplConfig.(*Config)
```
geeknoid: This error should be caught during ValidateConfig, right?

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
+
+const (
+	// ImplName is the canonical name of this implementation
+	ImplName = "istio/IPListChecker"
 )
 
-// AdapterConfig is used to configure a adapter.
-type AdapterConfig struct {
+// Register registration entry point
+func Register(r aspectsupport.Registry) {
```
geeknoid: A bit odd that the *adapter* registry is part of aspectsupport...

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
+		// bogus URL format
+		return nil, err
 	}
-	return err
-}
 
-func (a *adapterState) NewAspect(config adapter.AspectConfig) (adapter.Aspect, error) {
-	if err := a.ValidateAspectConfig(config); err != nil {
-		return nil, err
+	aa := aspectState{
```
geeknoid: Can we put the rest of the code from this method into aspect.go in a newAspect method? I'd rather keep related code together, and the code from this point on is really part of the aspect constructor.

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
 package ipListChecker
```
geeknoid: Giving this another look over, I now realize that the new design completely foregoes adapter-level config, all we have is an aspect level configuration.

I really believe there is a need for global adapter-level config, distinct from aspect-level config. As I mentioned previously, we use this kind of stuff all the time in Google. You can supply command-line options to a server when it starts up which is global to the server, independent of any configuration you may give to the server for individual tenants.

Since adapters don't have command-line options, adapter-level config would be the logical equivalent.

pkg/aspect/types.go
```diff
@@ -0,0 +1,107 @@
...
+		ImplName() string
+	}
+
+	// Config (aspect.Config) is the internal struct used for the Aspect proto
+	// at present it only has an additional field that stores the converted
+	// and typed proto
+	Config struct {
+		istioconfig.Aspect
+		// TypedParams points to the proto after google_protobuf.Struct is converted
+		TypedParams proto.Message
```
geeknoid: I don't understand why this is called TypedParams, but below we have AdapterConfig.TypedArgs. Why don't these two fields have the same name?

pkg/aspect/types.go
```diff
@@ -0,0 +1,107 @@
...
+		io.Closer
+		// ImplName returns the official name of adapter implementation ex. "istio/statsd"
+		ImplName() string
+	}
+
+	// Config (aspect.Config) is the internal struct used for the Aspect proto
+	// at present it only has an additional field that stores the converted
+	// and typed proto
+	Config struct {
+		istioconfig.Aspect
```
geeknoid: This remains very confusing.

This type is called Config and its first member is something called an Aspect. This is not related to the type Aspect right above though. One Aspect is an interface with methods, another Aspect is a block of config.

pkg/aspect/types.go
```diff
@@ -0,0 +1,107 @@
...
+	// ex: listChecker.Aspect adds --> CheckList(Symbol string) (bool, error)
+	// The extended interface is implemented by adapter impls.
+	// ex: //adapter/ipListChecker aspectState implements the listChecker.Aspect.
+	Aspect interface {
+		io.Closer
+		// ImplName returns the official name of adapter implementation ex. "istio/statsd"
+		ImplName() string
+	}
+
+	// Config (aspect.Config) is the internal struct used for the Aspect proto
```
geeknoid: Why repeat (aspect.Config)?
What makes this "an internal struct".
This talks about the Aspect proto, which is unrelated to the Aspect interface. Confusing.

pkg/aspect/types.go
```diff
@@ -0,0 +1,107 @@
...
+	// This interface is extended by  aspects; (metrics, listChecker, ...)
+	// The extended interface is implemented by adapter implementations
+	Adapter interface {
+		io.Closer
+		// Name returns the official name of this adapter. ex. "istio/statsd".
+		Name() string
+		// Description returns a user-friendly description of this adapter.
+		Description() string
+		// DefaultConfig returns a default configuration struct for this
+		// adapter Implmentation. This will be used by the configuration system to establish
```
geeknoid: implementation

pkg/aspect/types.go
```diff
@@ -0,0 +1,107 @@
...
+		// the shape of the block of configuration state passed to the NewAspect method.
+		DefaultConfig() (implConfig proto.Message)
+		// ValidateConfig determines whether the given configuration meets all correctness requirements.
+		ValidateConfig(implConfig proto.Message) error
+	}
+
+	// AdapterConfig (aspect.AdapterConfig) is the internal struct used for the Adapter proto
```
geeknoid: Same questions as above. Why repeat aspect.AdapterConfig, what makes this "internal", etc.

pkg/aspect/types.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	// Manager manages a specific aspect and presets a uniform interface
+	// to the rest of system
+	Manager interface {
+		// Execute dispatches to the given aspect using  config
+		// A cached instance of Aspect is provided that was previously obtained by
+		// calling NewAspect
+		// The evaluation is done under the context of an attribute bag and using
+		// an expression evaluator
+		Execute(cfg *CombinedConfig, asp Aspect, attrs attribute.Bag, mapper expr.Evaluator) (*Output, error)
```
geeknoid: Why does Execute need the config? Shouldn't that have been captured by NewAspect?

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
 func (a *adapterState) Close() error {
 	return nil
 }
 
-func (a *adapterState) DefaultAspectConfig() adapter.AspectConfig {
-	return &AspectConfig{
-		ProviderURL:     "http://localhost",
-		RefreshInterval: time.Minute,
-		TimeToLive:      time.Minute * 10,
+func (a *adapterState) NewAspect(c *listChecker.Config) (listChecker.Aspect, error) {
```
geeknoid: There's an asymmetry here. DefaultConfig and ValidateConfig deal in terms of proto.Message, but NewAspect doesn't. 

I think it'd prefer NewAspect to take a proto.Message like the other methods, as well as a listChecker.Config. And then listChecker.Config wouldn't carry an ImplConfig anymore, which would mean it could be a straight proto I think.

geeknoid: More specifically, what if the signature was:

NewAspect(cfg *proto.Message, lc listcheckerpb.Config)

Then we don't need the type listChecker.Config.

pkg/aspect/types.go
mandarjog: https://github.com/istio/api/blob/master/istio/config/v1/cfg.proto#L90
maintaining names from  Aspect.params and Adapter.args

adapter/ipListChecker/adapter.go
mandarjog: This is where we started, NewAspect takes an aspect specific config and an Impl config. Previously you had strong preference in using only 1 struct so, those were combined.

pkg/aspect/types.go
mandarjog: look at listChecker.manager's Execute and NewAspect methods.

Where to use the configuration is a choice for the aspect manager. 

The aspect manager can act on as much aspect specific config it wants to and then only use the subset of the configuration to call NewAspect.

In the above example, Both "checkAttribute" and "blackList" from 
`listcheckerpb "istio.io/api/istio/config/v1/aspect/listChecker"` are fully processed by aspect manager. So you just don't want to send down the protobuf config in its entirety. 

Think about what happens when both aspect manager *and* adapterImpl process the `blacklist` flag. They will end up negating outputs 2 times.

geeknoid:
*[adapter/ipListChecker/adapter.go, line 15 at r4](https://reviewable.io:443/reviews/istio/mixer/112#-K_6Ch8fn6ZaAGlu_WEv-r4-15:-K_KLoRoDBPpot2m4Pqu:bqbmvkw) ([raw file](https://github.com/istio/mixer/blob/257c0eaabfea533881a0a1f030976831f047995b/adapter/ipListChecker/adapter.go#L15)):*

Giving this another look over, I now realize that the new design completely foregoes adapter-level config, all we have is an aspect level configuration.

I really believe there is a need for global adapter-level config, distinct from aspect-level config. As I mentioned previously, we use this kind of stuff all the time in Google. You can supply command-line options to a server when it starts up which is global to the server, independent of any configuration you may give to the server for individual tenants.

Since adapters don't have command-line options, adapter-level config would be the logical equivalent.
</blockquote></details>

May the solution would in fact to allow adapters to have command-line options? If they use the normal Go flags feature, it should be picked up by our Cobra front-end and should just work as-is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/112)*

geeknoid:
*[adapter/ipListChecker/adapter.go, line 104 at r1](https://reviewable.io:443/reviews/istio/mixer/112#-K_6Ch8fn6ZaAGlu_WEv-r1-104:-K_KNL7Er6s22MMa_vMZ:bc7guri) ([raw file](https://github.com/istio/mixer/blob/0d1f15a09ddf7477aa11417491e2a0d9222c5df1/adapter/ipListChecker/adapter.go#L104)):*

Removed it. 
Perhaps we should add a   Ready() bool ? so synchronous nature is not forced?

</blockquote></details>

The synchronous call hasn't actually been removed, right?

In any case, I think the sync call is actually necessary for proper operation. It just means the aspect manager above needs to orchestrate the startup and be prepared to do things in parallel.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/112)*

geeknoid:
Review status: 0 of 24 files reviewed at latest revision, 25 unresolved discussions.

---

*[adapter/ipListChecker/adapter.go, line 76 at r4](https://reviewable.io:443/reviews/istio/mixer/112#-K_6Ch8fn6ZaAGlu_WEv-r4-76:-K_KMQ1Ikv9FsIFjZ5-m:b4awtbj) ([raw file](https://github.com/istio/mixer/blob/257c0eaabfea533881a0a1f030976831f047995b/adapter/ipListChecker/adapter.go#L76)):*

This is where we started, NewAspect takes an aspect specific config and an Impl config. Previously you had strong preference in using only 1 struct so, those were combined.
</blockquote></details>

Yes, you're right. Looking at the broader change though, the current PR doesn't feel right though, and in fact I think my suggestion here isn't right either.

The generic aspect config should not be given to the adapter code. Only the adapter-specific config should be given. The aspect config should instead only be consumed by the aspect manager. So I suggest instead that the signature here only be:

  NewAspect(cfg proto.Message) (Aspect, error)

This is consistent with the signature for ValidateConfig and DefaultConfig which deal in terms of proto.Message. And it won't leak aspect-level config into the adapter.

---

*[pkg/aspect/types.go, line 49 at r4](https://reviewable.io:443/reviews/istio/mixer/112#-K_6Ch8g_kjQzuzeFvmz-r4-49:-K_KP7DixyNo1OQutuFz:b-ht3f9r) ([raw file](https://github.com/istio/mixer/blob/257c0eaabfea533881a0a1f030976831f047995b/pkg/aspect/types.go#L49)):*

https://github.com/istio/api/blob/master/istio/config/v1/cfg.proto#L90
maintaining names from  Aspect.params and Adapter.args
</blockquote></details>

We should fix those names in the config format to be consistent too.

---

*[pkg/aspect/types.go, line 101 at r4](https://reviewable.io:443/reviews/istio/mixer/112#-K_6Ch8g_kjQzuzeFvmz-r4-101:-K_KPWgC0HbIJ5fzDeBb:b-899a9s) ([raw file](https://github.com/istio/mixer/blob/257c0eaabfea533881a0a1f030976831f047995b/pkg/aspect/types.go#L101)):*

look at listChecker.manager's Execute and NewAspect methods.

Where to use the configuration is a choice for the aspect manager. 

The aspect manager can act on as much aspect specific config it wants to and then only use the subset of the configuration to call NewAspect.

In the above example, Both "checkAttribute" and "blackList" from 
`listcheckerpb "istio.io/api/istio/config/v1/aspect/listChecker"` are fully processed by aspect manager. So you just don't want to send down the protobuf config in its entirety. 

Think about what happens when both aspect manager *and* adapterImpl process the `blacklist` flag. They will end up negating outputs 2 times.  
</blockquote></details>

I really don't like how Execute and NewAspect compose together.

In the proposal I sent separately, the coupling between these two is improved. NewAspect returns an AspectWrapper instead of an Aspect. This wrapper captures the config necessary for use by Execute. It also performs the proto struct -> Go struct conversion internally so we don't need the wrapping Config and AdapterConfig types, which substantially simplifies the end to end story. The general API surface can deal in terms of the original config protos directly, avoiding new names and new types.

---

*[pkg/aspect/listChecker/types.go, line 30 at r3](https://reviewable.io:443/reviews/istio/mixer/112#-K_6Ch8g_kjQzuzeFvmy-r3-30:-K_KQZMuYew5VUzPoQLU:b-5g35rs) ([raw file](https://github.com/istio/mixer/blob/714a942039d1f2441a3b347b2f0e162784715917/pkg/aspect/listChecker/types.go#L30)):*

This is an optimization.

These are go friendly version of https://github.com/istio/api/blob/master/istio/config/v1/cfg.proto#L90.
The only difference is that they are typed. 
They convert the Params / Args from proto.Struct to the actual underlying protobuf.

The following code is from listChecker/manager.go Execute ()
```go
// check if mapping is available
	if acfg, found = cfg.Aspect.TypedParams.(*listcheckerpb.Config); !found {
		return nil, fmt.Errorf("Unexpected type %#v", cfg.Aspect.TypedParams)
	}

	// CheckAttribute should be processed and sent to input
	if symbolExpr, found = cfg.Aspect.Inputs[acfg.CheckAttribute]; !found {
		return nil, fmt.Errorf("Mapping for %s not found", acfg.CheckAttribute)
	}
```
without `TypedParams` we would need to convert from protobuf.Struct to  listcheckerpb.Config on every call. 

</blockquote></details>

This is addressed by introducing an AspectWrapper type as mentioned in my other comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/112)*

adapter/ipListChecker/adapter.go
wora: Why this name does not match package name? It seems confusing to use slightly different name for the same thing.

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
 	return "Checks whether an IP address is present in an IP address list."
 }
 
-func (a *adapterState) DefaultAdapterConfig() adapter.AdapterConfig {
-	return &AdapterConfig{}
+func (a *adapterState) DefaultConfig() proto.Message {
+	return &Config{
+		ProviderUrl:     "http://localhost",
+		RefreshInterval: 60,
+		Ttl:             120,
```
wora: Shouldn't TTL decided by the caller? If the config is stale, are we going to return false or true or error?

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
 }
 
-func (a *adapterState) ValidateAdapterConfig(config adapter.AdapterConfig) error {
-	_ = config.(*AdapterConfig)
-	return nil
-}
+func (a *adapterState) ValidateConfig(cfg proto.Message) (err error) {
+	c, ok := cfg.(*Config)
+	if !ok {
+		return fmt.Errorf("Invalid message type %#v", cfg)
```
wora: The error message should be self contained within a log statement. We should not assume any context when people scanning the logs. Things like "invalid message type xxx" is not meaningful to solve the problem. Something like "Invalid config message type xxx. IPListChecker requires config message type xxx."

adapter/ipListChecker/adapter.go
```diff
@@ -15,74 +15,93 @@
...
+	}
+	var u *url.URL
 
-func (a *adapterState) Configure(config adapter.AdapterConfig) error {
-	return a.ValidateAdapterConfig(config)
+	if u, err = url.Parse(c.ProviderUrl); err == nil {
+		if u.Scheme == "" || u.Host == "" {
+			err = fmt.Errorf("Scheme (%s) and Host (%s) cannot be empty", u.Scheme, u.Host)
```
wora: We need to generate formal errors. This seems to be free text errors for everything. It is hard to handle at higher level.

adapter/ipListChecker/adapter.go
geeknoid: This a default config option. The adapter considers the list valid for the TTL even when it can't contact the server. After that, it flushes its cached list and fails any list checks.

PR #120, opened 2017-01-03
Move genericListChecker to the new plan.

geeknoid:
This also introduces a newAdapter method which is useful
for tests, when the nominal adapterState{} is not just an
empty struct.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/120)


adapter/ipListChecker/adapter.go
```diff
@@ -26,16 +26,20 @@ import (
 
 const (
 	// ImplName is the canonical name of this implementation
-	ImplName = "istio/IPListChecker"
+	ImplName = "istio/ipListChecker"
```
mandarjog: re @wora comment earlier, should we make it istio.io/IPListChecker?

geeknoid:
The comment was saying "why doesn't the name match the namespace". So this
change is making it in fact match the namespace....

mandarjog:
I thought the comment said it should be
istio.io/IpListChecker  ?

PR #123, opened 2017-01-04
Simplify adapter implementations a bit.

geeknoid:
A proposal to simplify adapter implementations a bit:

- Remove aspect.ImplName, this can be handled by AspectWrapper.

- We can guarantee that the configs passed to ValidateConfig are of the right type, so each adapter doesn't need to check.

- We can guarantee that NewAspect is only ever called with validated configs, so each adapter doesn't need to check.

- Reformat for higher-density.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/123)

### DIFF
--- a/adapter/denyChecker/adapter.go
+++ b/adapter/denyChecker/adapter.go
@@ -15,8 +15,6 @@
 package denyChecker
 
 import (
-	"fmt"
-
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/genproto/googleapis/rpc/code"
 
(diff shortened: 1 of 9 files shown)


adapter/denyChecker/adapter.go
```diff
@@ -15,59 +15,29 @@
...
-
-func (a *adapterState) Name() string {
-	return ImplName
-}
-
-func (a *adapterState) Description() string {
-	return "Deny every check request"
-}
+func newAdapter() denyChecker.Adapter                          { return &adapterState{} }
+func (a *adapterState) Name() string                           { return "istio/denyChecker" }
```
mandarjog: We should define these as constants (?)

adapter/genericListChecker/adapter.go
```diff
@@ -15,58 +15,26 @@
...
+func (a *adapterState) Close() error                           { return nil }
+func (a *adapterState) ValidateConfig(cfg proto.Message) error { return nil }
+func (a *adapterState) DefaultConfig() proto.Message           { return &Config{} }
 
 func (a *adapterState) NewAspect(cfg proto.Message) (listChecker.Aspect, error) {
-	if err := a.ValidateConfig(cfg); err != nil {
-		return nil, err
-	}
-
 	return newAspect(cfg.(*Config))
```
mandarjog: This makes sense. 
We should however ensure that Manager.NewAspect is always wrapped in a defer / recover block. 
A naked assertion will cause a panic

adapter/denyChecker/adapter.go
geeknoid: Why do we need the extra LOCs to declare them as constants?

adapter/genericListChecker/adapter.go
geeknoid: That's true in general right? We should make the mixer tolerant of crashes in adapters if we can.

adapter/denyChecker/adapter.go
douglas-reid: i don't know that there is an overwhelming need. This may be a personal prefs thing (i like having the name/desc consts at the top of the adapters as sort of an easily discoverable, readable, self-documenting thing, but I can see the argument against).

Does the name need to be exported or referenced directly?  Before, it was exported.

adapter/genericListChecker/adapter.go
douglas-reid: While true, do we not want to check the type assertion here when we can? 

```
c, ok := cfg.(*Config)
if !ok {
  return nil, errors.New("Supplied config does not match required configuration type")
}
return newAspect(c)
```

geeknoid: I'm specifically trying to remove boilerplate in this change, I removed this check and associated error report on purpose to keep implementations of adapters as succinct as we can.

adapter/denyChecker/adapter.go
geeknoid: At this point, I don't see a need to have the names and descriptions as constants since we are registering the adapters systematically. So the names can be found by looking at the registered adapters.

adapter/genericListChecker/adapter.go
mandarjog: We will have this code in the aspect managers. 
We *will* have to assert types there.

geeknoid: Correct. And in general, we'll catch any panics from an adapter and implement a recovery policy.

mandarjog:
If we can find a tool to format trivial functions on a single line, that would be awesome.

geeknoid:
PTAL, can I get a green light?

PR #128, opened 2017-01-04
Introduce the ConfigError type.

geeknoid:
ConfigError is intended to be used alongside multierrors. It provides a way for adapters to returned slightly structured errors from the ValidateConfig method. At a higher level, we can extract ConfigError objects and recover the field name against which the error was declared. That should be sufficient for us to point the user to a particular line of input code where the problem originated from.

The use of multierror also makes it so we encourage implementations of ValidateConfig to return all config errors in one shot, as opposed to failing fast after the first error is encountered. This will deliver a better experience for the operator as they'll get all the errors at once instead of playing whack-a-mole.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/128)

### DIFF
--- a/adapter/ipListChecker/BUILD
+++ b/adapter/ipListChecker/BUILD
@@ -12,10 +12,11 @@ go_library(
     deps = [
         "@com_github_golang_glog//:go_default_library",
         "@in_gopkg_yaml_v2//:go_default_library",
+        "//pkg/adapter:go_default_library",
         "//pkg/aspect:go_default_library",
         "//pkg/aspectsupport:go_default_library",
         "//pkg/aspect/listChecker:go_default_library",
-        "@com_github_golang_protobuf//proto:go_default_library"
+        "@com_github_golang_protobuf//proto:go_default_library",
     ],
 )
 
--- a/adapter/ipListChecker/adapter.go
+++ b/adapter/ipListChecker/adapter.go
@@ -15,16 +15,15 @@
 package ipListChecker
 
 import (
-	"fmt"
 	"net/url"
 
 	"github.com/golang/protobuf/proto"
 
+	"istio.io/mixer/pkg/adapter"
 	"istio.io/mixer/pkg/aspect/listChecker"
 	"istio.io/mixer/pkg/aspectsupport"
 )
 
-// Register registration entry point
 func Register(r aspectsupport.Registry) error {
 	return r.RegisterCheckList(newAdapter())
 }
@@ -33,22 +32,26 @@ type adapterState struct{}
 
 func newAdapter() listChecker.Adapter { return &adapterState{} }
 func (a *adapterState) Name() string  { return "istio/ipListChecker" }
+
 func (a *adapterState) Description() string {
 	return "Checks whether an IP address is present in an IP address list."
 }
 
 func (a *adapterState) Close() error { return nil }
 
-func (a *adapterState) ValidateConfig(cfg proto.Message) error {
+func (a *adapterState) ValidateConfig(cfg proto.Message) (ce *adapter.ConfigErrors) {
 	c := cfg.(*Config)
 
-	if u, err := url.Parse(c.ProviderUrl); err == nil {
+	u, err := url.Parse(c.ProviderUrl)
+	if err != nil {
+		ce = ce.Append("ProviderUrl", err)
+	} else {
 		if u.Scheme == "" || u.Host == "" {
-			err = fmt.Errorf("Scheme (%s) and Host (%s) cannot be empty", u.Scheme, u.Host)
+			ce = ce.Appendf("ProviderUrl", "URL scheme and host cannot be empty")
 		}
-		return err
 	}
-	return nil
+
+	return
 }
 
 func (a *adapterState) DefaultConfig() proto.Message {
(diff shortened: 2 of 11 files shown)


adapter/ipListChecker/adapter.go
```diff
@@ -61,14 +63,18 @@ func (a *adapterState) ValidateConfig(cfg proto.Message) (err error) {
 	if !ok {
```
mandarjog: Can we change ValidateConfig to return ConfigError instead ?

mandarjog: It can derive from multi error if needed.

geeknoid:
OK, I can do that.

adapter/ipListChecker/BUILD
```diff
@@ -12,10 +12,12 @@ go_library(
     deps = [
         "@com_github_golang_glog//:go_default_library",
         "@in_gopkg_yaml_v2//:go_default_library",
+        "//pkg/adapter:go_default_library",
```
douglas-reid: Nit: in general, it might be nice if we decided on a convention for build file formats. I suggest using [buildifier](https://github.com/bazelbuild/buildifier). this would end up having these deps be sorted.

pkg/adapter/configError.go
```diff
@@ -0,0 +1,47 @@
...
+type ConfigError struct {
+	Field      string
+	Underlying error
+}
+
+// Append adds a ConfigError to a multierror. This function is intended
+// to be used by adapter's ValidateConfig method to report errors
+// in configuration. The field parameter indicates the name of the
+// specific configuration field name that is problematic.
+func Append(base *me.Error, field string, format string, args... interface{}) *me.Error {
```
douglas-reid: can this be written as:

`func Append(base *me.Error, field, format string, args... interface{})`

(remove the unnecessary 'string' that follows field)?

pkg/adapter/configError.go
```diff
@@ -0,0 +1,47 @@
...
+	me "github.com/hashicorp/go-multierror"
+)
+
+// ConfigError represents an error encountered while validating a block of configuration state.
+type ConfigError struct {
+	Field      string
+	Underlying error
+}
+
+// Append adds a ConfigError to a multierror. This function is intended
```
douglas-reid: why not expose a NewConfigError() method (or even ConfigErrorf()) that took field, format, and args and then use the existing Append() method?

`func NewConfigError(field, format string, args... interface{}) ConfigError {...}`

or

`func ConfigErrorf(field, format string, args... interface{})`

The calling code could then be something like:

`me.Append(adapter.ConfigErrorf("field1", "Bad ...: %s", fieldVal))`

Which strikes me as more readable (and closer to how fmt.Errorf(), etc., work).

Also, configError.go wouldn't have to import go-multierror.

geeknoid:
PTAL.

Followed Doug's suggestions. Definitely simpler...

adapter/ipListChecker/adapter.go
```diff
@@ -35,25 +33,27 @@ type adapterState struct{}
...
 func (a *adapterState) Name() string  { return "istio/ipListChecker" }
+
 func (a *adapterState) Description() string {
 	return "Checks whether an IP address is present in an IP address list."
 }
 
 func (a *adapterState) Close() error { return nil }
 
-func (a *adapterState) ValidateConfig(cfg proto.Message) me.Errors {
+func (a *adapterState) ValidateConfig(cfg proto.Message) error {
```
mandarjog: I think it is more natural for ValidateConfig to return ConfigError.

As an adapter author, by simply looking at the API, I will investigate how to construct a config Error and do the correct thing.

There is too much here that is not apparent in the API signature. 
The interaction between multi error / ConfigError has to be studied.

geeknoid:
As it is now, ConfigError only holds a single error, so it can't be the
return value for ValidateConfig. I'd need to introduce a ConfigErrors type
fo that purpose. Do you think it is worth it?
​

mandarjog:
We just need one type that can hold multiple configuration errors. I think ConfigError can hold a list?
Or we can call it ConfigErrors.

kyessenov:
I wonder if using annotations on .proto and auto-generated validator is the way to go, e.g. https://github.com/mwitkow/go-proto-validators. It would be nice to have at least some reflection facilities for proto sources (field names, source line numbers to attach to errors) without manually writing all the code.

geeknoid:
@kyessenov 
Regarding using a generic proto validator, that is a noble goal if such a thing exists. What you pointed to is somewhat of a toy, it only supports very minimal validation on individual fields. It won't support anything complex and anything involving multiple fields.

In general, I expect the config for individual adapters to be fairly simple, so implementing the validation logic will require a small amount of code.

pkg/adapter/configError.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// The usage pattern for this type is pretty simple:
+//
+// 	func (a *adapterState) ValidateConfig(cfg proto.Message) (ce *adapter.ConfigErrors) {
+//      c := cfg.(*Config)
+//      if c.Url == nil {
+// 			ce = ce.Appendf("Url", "Must have a valid URL")
+//		}
+//		if c.RetryCount < 0 {
+//          ce = ce.Appendf("RetryCount", "Expecting >= 0, got %d", cfg.RetryCount)
```
douglas-reid: formatting looks funky in these comments...

Issue #127, opened 2017-01-04
Should dynamically build adapter proto files instead of having .pb.go files checked in.


geeknoid:
Instead of having files like adapter/denyChecker/cfg.pb.go checked in, they should be generated on the fly during a build based on the checked in .proto file.

douglas-reid:
We have an approach for this now.  Will address other existing *.pb.go files in adapter package in short order.

Issue #125, opened 2017-01-04
Update denyChecker config to use a google.rpc.Status


geeknoid:
I'd like the adapter configuration at adapter/denyChecker/cfg.proto to use a google.rpc.Status instead of an int32. But I can't figure out how to get the dependencies setup properly to get that to compile.

Would you mind taking a look to see if you can figure it out?

douglas-reid:
Marking as resolved, based on #130.

PR #130, opened 2017-01-04
Setup proto gen for denyChecker config.

douglas-reid:
This should address #125 and #127.

@geeknoid please let me know if this is a reasonable direction.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/130)


geeknoid:
This definitely works. Mandar was trying to get things such that the Config type was actually in the same package as the rest of the adapter code. I'm OK with this organization though if he is. And I definitely like not having the pb.go file checked in.

mandarjog:
did you try using the same package as the code ?
The only issue I had with it was co-existence with go tools.

adapter/denyChecker/adapter_test.go
```diff
@@ -28,23 +31,23 @@ func TestAll(t *testing.T) {
...
+	a, err = b.NewAspect(&pb.Config{&status.Status{Code: int32(code.Code_INVALID_ARGUMENT)}})
 	if err != nil {
 		t.Errorf("Unable to create aspect: %v", err)
 	}
 
-	status = a.Deny()
-	if status.Code != int32(code.Code_INVALID_ARGUMENT) {
-		t.Errorf("a.Deny returned %d, expected %d", status.Code, int32(code.Code_INVALID_ARGUMENT))
+	s = a.Deny()
+	if s.Code != int32(code.Code_INVALID_ARGUMENT) {
```
douglas-reid: code.Code is an enum. status.Code is an int32. without the cast, you will get something like: `invalid operation: ... (mismatched types int32 and Code)`

douglas-reid:
I did not try to merge packages. I wasn't aware this was a goal. I'm not convinced that is the right thing to do, but I can play with it to see how such a thing _might_ work. I'm skeptical that it won't cause issues and/or confusion.

mandarjog:
I think being able to keep all adapter code together is a good goal, but visually it is together since the config file is right there. As a new adapter author I would find it confusing as to why is this not in the current directory.  

That said, it is ok for now. We can revisit it later.

PR #135, opened 2017-01-06
Add support for aspect environments.

geeknoid:
Introduce adapter.Env to provide generic services to adapters which integrates into the broader mixer world. For starters, this means providing a logger that adapters use instead of any old arbitrary logger they happen to link to. In the future, maybe we use the Env to handle adapter metric pumping, timeout handling, or other global concerns.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/135)

### DIFF
--- a/pkg/aspectsupport/uber/logger.go
+++ b/pkg/aspectsupport/uber/logger.go
@@ -1,0 +1,43 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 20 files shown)


pkg/aspectsupport/uber/logger.go
```diff
@@ -0,0 +1,42 @@
...
+type logger struct {
+	aspect string
+}
+
+func newLogger(aspect string) *logger {
+	return &logger{aspect: aspect}
+}
+
+func (l *logger) Infof(format string, args ...interface{}) {
+	glog.Infof(l.aspect + ":" + format, args...)
```
mandarjog: The issue with this way of logging is that glog will always claim that `uber.go:31` is the logline.

We need to fork it.

geeknoid:
PTAL.

Fixed the stack frame issue you mentioned. No need to fork glog.

Note that I expect we'll eventually want the logger implementation to be a bit more sophisticated than what it is here. For now though, the goal is to set things up so adapters can be written against the right model.

PR #136, opened 2017-01-06
Integrate a bunch of linters into travis and our environment.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/136)

### DIFF
--- a/adapter/genericListChecker/BUILD
+++ b/adapter/genericListChecker/BUILD
@@ -5,15 +5,15 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "go_default_library",
     srcs = [
-        "aspect.go",
         "adapter.go",
+        "aspect.go",
         "cfg.pb.go",
     ],
     deps = [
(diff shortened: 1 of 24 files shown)


geeknoid:
This integrates stuff into Travis now. PTAL.

adapter/denyChecker/config_proto
```diff
@@ -0,0 +1 @@
+bazel-genfiles/adapter/denyChecker
```
douglas-reid: nit: do we want a comment explaining this file?

adapter/genericListChecker/BUILD
```diff
@@ -29,10 +29,10 @@ load("@org_pubref_rules_protobuf//go:rules.bzl", "go_proto_library")
 
 go_proto_library(
     name = "genericListChecker",
+    output_to_workspace = False,
```
douglas-reid: nit: i think False is the default, so we can remove this line if we want.

geeknoid:
Whops, that config_proto file should not be checked in, I'll remove it from
the PR.

I'll strip out all output_to_workspace in the build files.

PR #138, opened 2017-01-06
Enable more linters.

geeknoid:
Fix a couple lint errors, including an actual bug!

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/138)


pkg/aspectsupport/listChecker/manager.go
```diff
@@ -86,7 +86,7 @@ func (*manager) DefaultConfig() proto.Message {
 func (*manager) ValidateConfig(implConfig proto.Message) (ce *aspect.ConfigErrors) {
 	lc := implConfig.(*listcheckerpb.Config)
 	if lc.CheckAttribute == "" {
-		ce.Appendf("check_attribute", "Missing")
+		ce = ce.Appendf("check_attribute", "Missing")
```
douglas-reid: should we also add a test that actually exercises this bit and checks it?  Seems like the right thing to do.

pkg/aspectsupport/uber/env_test.go
```diff
@@ -24,5 +24,5 @@ func TestEnv(t *testing.T) {
 	log := env.Logger()
 	log.Infof("Test%s", "ing")
 	log.Warningf("Test%s", "ing")
-	log.Errorf("Test%s", "ing")
+	_ = log.Errorf("Test%s", "ing")
```
douglas-reid: is this required by a linter?  seems ... unnecessary.

geeknoid: Yes, there's a linter that demands you acknowledge that functions return errors and don't drop 'em on the floor. In general, that's a good thing IMO.

pkg/aspectsupport/listChecker/manager.go
geeknoid: Once we get code coverage up and running, we'll need to expand our tests considerably, including here. I'd rather not address that case in this PR if you don't mind...

geeknoid:
PTAL. Added golint which required many more warnings to be addressed.

pkg/aspectsupport/listChecker/manager.go
douglas-reid: OK. It sounds reasonable to postpone, given the scope of this PR.

pkg/aspectsupport/uber/env_test.go
douglas-reid: I guess... but the difference between "_ =" and dropping the error on the floor is not clear to me.  Isn't the improvement something like:

```
if err := log.Errorf(..); err != nil {
   t.Errorf("Errorf(): unexpected error %v", err) 
}
```

geeknoid:
Is a cue to the reader of the code which says "I know this function returns
an error and I have chosen to ignore it".

In another environment I worked on, we started out with a magic function
call to capture this:

  Result.Ignore(myFunction())

We ended up promoting this to a language feature:

  ignore myFunction()

and it turns out golang calls this feature "_ =" instead of "ignore" :-)

douglas-reid:
Restated: if we care about not allowing ignored errors, shouldn't we go full monty here in the test and actually check the error that is possibly returned?

Ignoring the error, even explicitly, seems to violate the spirit of the lint check to me.

In general though, this LGTM. I'd like to see it pass the build first however.

pkg/aspectsupport/uber/env_test.go
geeknoid: OK, I get your point. I added a test to make sure we get a valid error return.

mandarjog:
Log.Errorf is different in this regard.It returned an error because it constructs an error from the given arguments. It did not encounter this error. So it is ok to ignore.

douglas-reid:
Wait, why are we logging an error and also generating it for further processing?  That seems a bit like an anti-pattern to me.  Is the assertion that the error will be handled *without logging* by the mixer once it is outside of the adapter code?

geeknoid:
All linters run and pass.

PTAL.

.travis.yml
```diff
@@ -33,13 +34,8 @@ before_install:
   - cat .bazelrc.travis .bazelrc.orig > .bazelrc
 
 script:
-  #- bazel --output_base=${HOME}/bazel/outbase test //...
   - bazel test //...
-  - bin/bazel_to_go.py
-  - ls -l ${TRAVIS_BUILD_DIR}
-  - go test -v ./pkg/... ./adapter/... ./cmd/...
```
douglas-reid: do we want to keep this line, or not?  i believe it was added to ensure `go test` and `bazel test` interop.

geeknoid:
​Running the linters achieves the same effect, and it's actually more
thorough (since we don't have tests for everything yet)​

PR #139, opened 2017-01-06
Enable code coverage.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/139)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 [![GoDoc](https://godoc.org/github.com/istio/mixer?status.svg)](http://godoc.org/github.com/istio/mixer)
 [![Build Status](https://travis-ci.org/istio/mixer.svg?branch=master)](https://travis-ci.org/istio/mixer)
 [![Go Report Card](https://goreportcard.com/badge/github.com/istio/mixer)](https://goreportcard.com/report/github.com/istio/mixer)
+[![codecov.io](https://codecov.io/github/istio/mixer/coverage.svg?branch=master)](https://codecov.io/github/istio/mixer?branch=master)
 
 The Istio mixer provides the foundation of the Istio service mesh
 design. It is responsible for insulating the Istio proxy and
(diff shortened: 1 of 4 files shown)


bin/preplinters.sh
```diff
@@ -4,10 +4,9 @@
...
 ## mixer source base.
 
-set -ev
-
 go get -u github.com/alecthomas/gometalinter
 go get -u github.com/bazelbuild/buildifier/buildifier
 go get -u github.com/3rf/codecoroner
 gometalinter --install >/dev/null
 bin/bazel_to_go.py
+ln -s -f bazel-genfiles/adapter/denyChecker adapter/denyChecker/config_proto
```
douglas-reid: aside: will we have to add this link on all adapter configs?  it seems like we need a better way...

geeknoid: Yeah, this is not a great story. Maybe Mandar can figure out a more generic approach in his Python script.

README.md
```diff
@@ -4,6 +4,7 @@
 [![GoDoc](https://godoc.org/github.com/istio/mixer?status.svg)](http://godoc.org/github.com/istio/mixer)
 [![Build Status](https://travis-ci.org/istio/mixer.svg?branch=master)](https://travis-ci.org/istio/mixer)
 [![Go Report Card](https://goreportcard.com/badge/github.com/istio/mixer)](https://goreportcard.com/report/github.com/istio/mixer)
+[![codecov](https://codecov.io/gh/istio/mixer/branch/master/graph/badge.svg?token=m2jmf9iXJp)](https://codecov.io/gh/istio/mixer)
```
douglas-reid: question: is that token in this URL necessary? is it OK to have it exposed?

geeknoid:
I just took that line from the Mixologist README.md, so you tell me :-)

douglas-reid:
If it is the old token, then we should remove it for sure. I _think_ the token is only needed for private repos. @mandarjog, any insights?

geeknoid:
Seems to be working, PTAL.

PR #140, opened 2017-01-07
Add logger aspect to the mixer.

douglas-reid:
This adds a basic aspect with a Log([]Entry) interface to be
implemented by various adapters. A simple manager is provided for
downscoping to an Entry from the attribute.Bag based on the
configured Input attributes.

Tested with: pkg/aspectsupport/logger$ bazel test ...
Test coverage: 90%.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/140)

### DIFF
--- a/pkg/aspect/logger/logger.go
+++ b/pkg/aspect/logger/logger.go
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc.
+// Copyright 2017 Google Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/aspect/logger/logger.go
+++ b/pkg/aspect/logger/logger.go
@@ -1,0 +1,48 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logger
+
+import (
+	"github.com/golang/protobuf/proto"
+	"istio.io/mixer/pkg/aspect"
+)
+
+type (
+	// Aspect is the interface for adapters that will handle logs data
+	// within the mixer.
+	Aspect interface {
+		aspect.Aspect
+
+		// Log directs a backend adapter to process a batch of
+		// log entries derived from potentially several Report() calls.
+		Log([]Entry) error
+	}
+
+	// Entry is a set of key-value pairs of attributes that together
+	// constitute the data for log entry. The key
+	// is the attribute name, and the value is a typed value for the
+	// named attribute.
+	Entry map[string]interface{}
+
+	// Adapter is the interface for building Aspect instances for
+	// mixer logging backends.
+	Adapter interface {
+		aspect.Adapter
+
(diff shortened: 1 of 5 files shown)


pkg/aspect/logger/logger.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2016 Google Inc.
```
geeknoid: 2017

pkg/aspect/logger/logger.go
```diff
@@ -0,0 +1,48 @@
...
+		// Log directs a backend adapter to process a batch of
+		// log entries derived from potentially several Report() calls.
+		Log([]Entry) error
+	}
+
+	// Entry is a set of key-value pairs of attributes that together
+	// constitute the data for log entry. The key
+	// is the attribute name, and the value is a typed value for the
+	// named attribute.
+	Entry map[string]interface{}
```
geeknoid: This means the adapters need access to the known attribute type descriptors. Perhaps these can be provided as part of the environment?

pkg/aspectsupport/logger/manager.go
```diff
@@ -0,0 +1,102 @@
...
+// NewManager returns an aspect manager for the logger aspect.
+func NewManager() aspectsupport.Manager {
+	return &manager{}
+}
+
+func (m *manager) NewAspect(c *aspectsupport.CombinedConfig, a aspect.Adapter, env aspect.Env) (aspectsupport.AspectWrapper, error) {
+	// cast to logger.Adapter from aspect.Adapter
+	logAdapter, ok := a.(logger.Adapter)
+	if !ok {
+		return nil, fmt.Errorf("Adapter of incorrect type. Expected logger.Adapter got %#v %T", a, a)
```
geeknoid: Error strings should start with a lowercase letter (lint warning...)

pkg/aspectsupport/logger/manager.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	// TODO: replace when c.Adapter.TypedParams is ready
+	cpb := logAdapter.DefaultConfig()
+	if c.Adapter.Params != nil {
+		if err := structToProto(c.Adapter.Params, cpb); err != nil {
+			return nil, fmt.Errorf("Could not parse adapter config: %v", err)
```
geeknoid: Lowercase

pkg/aspect/logger/logger.go
douglas-reid: and I thought I had caught all of these...

pkg/aspect/logger/logger.go
douglas-reid: the values themselves should already have types, as they come out of the mapper (as I understand it). so, for logs, this might not be that big of a deal, as we will probably have a payload attribute that is ultimately treated as raw text or json and we can count on a json serializer, etc., to build the rest appropriately based on the existing type.

but yes, if the adapters want to do something smart with the inputs, they will need to understand their types.

do inputs (derived attributes) have type descriptors?  where will that information live?

either these smarts need to go into the Manager, so it can extract an entry of typed attributes or the attribute descriptors need to be passed, or we can pass a slimmed down attribute.Bag as the Entry or something else.

pkg/aspectsupport/logger/manager.go
douglas-reid: good catch. i thought lowercase was preferred, but thought I saw that we weren't doing that elsewhere and then thought I my memory was bad. I'll update here and below.

douglas-reid:
PTAL. Addressed date/caps issues.

I don't have a good answer for passing down inputs to adapters in a typed fashion at this point (without adapters being written with full knowledge of the expected type for the inputs ahead of time).

pkg/aspect/logger/logger.go
geeknoid: Ultimately, the mixer will have in its possession all of the attribute descriptors. Thus, for a given attribute name, it will be able to know the attribute's data type.

PR #149, opened 2017-01-10
Mention 'make lint' and 'make fmt'

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/149)


mandarjog:
@geeknoid  we could add this doc, however the local pre-commit hook that I submitted will just do these things before allowing a commit.
We could just state that you should enable that pre-commit hook?

geeknoid:
I think having these as separate items distinct from the hook is still desirable. So we can augment this doc to mention the pre-commit hook in your PR.

PR #150, opened 2017-01-10
Add local pre-commit hook

mandarjog:
Run bin/pre-commit to install the pre-commit hook
It will run
fmt
tests

before allowing the commit

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/150)


mandarjog:
Good idea re: linters.

The name `pre-commit` is defined by git. This is the actual hook.
It is clever that is also installs itself if it is called directly.

PR #154, opened 2017-01-11
Add the attribute.Value to extract untyped attribute values from a bag.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/154)

### DIFF
--- a/pkg/attribute/bag_test.go
+++ b/pkg/attribute/bag_test.go
@@ -260,3 +260,84 @@ func TestBadTimestamp(t *testing.T) {
 	}
 	defer at.EndRequest()
 }
+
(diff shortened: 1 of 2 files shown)


pkg/attribute/bag_test.go
```diff
@@ -260,3 +260,84 @@ func TestBadTimestamp(t *testing.T) {
 	}
 	defer at.EndRequest()
 }
+
+func TestValu(t *testing.T) {
```
douglas-reid: TestValue ?

geeknoid:
Fixed test name. PTAL.

PR #156, opened 2017-01-12
Move aspect manager config from istio.io/api into the mixer repo.

geeknoid:
This follows the same pattern as adapter configs in terms of how
we handle the protos.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/156)

### DIFF
--- a/pkg/aspectsupport/denyChecker/config/config.proto
+++ b/pkg/aspectsupport/denyChecker/config/config.proto
@@ -14,13 +14,10 @@
 
 syntax = "proto3";
 
-package adapter.denyChecker.config;
+package pkg.adaptersupport.denyChecker.config;
 
 option go_package="config";
 
-import "google/rpc/status.proto";
-
+// Configures a deny checker aspect
 message Params {
-	// The error to return when denying a request.
-	google.rpc.Status error = 1;
 }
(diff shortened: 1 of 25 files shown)


pkg/aspectsupport/denyChecker/config/config.proto
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package pkg.adaptersupport.denyChecker.config;
+
```
douglas-reid: non-blocker: Should we set an `option go_package="config";` in these proto files?  And if so, can we rename the core message to be Params (or similar) to match the layout of the yaml and to avoid config.Config stutter?

this would allow us to simply:
`import "istio.io/mixer/pkg/aspectsupport/denyChecker/config"` instead of:
`import pb "istio.io/mixer/pkg/aspectsupport/denyChecker/config"`

geeknoid: Done.

PR #157, opened 2017-01-12
Update logger aspect manager handling of aspect config.

douglas-reid:
This PR updates the manager for the logger aspect to properly report the config for the logger.Aspect as the DefaultConfig(). It uses the logger aspect config to build a more detailed logger.Entry to pass into Log() calls. The existing stdioLogger is updated to reflect these changes to the logger aspect config handling.

A few notes:
* no LogEntryDescriptors are currently provided, nor are they looked up. A subsequent PR will address providing an initial "standard" LogEntryDescriptor for short-term use by the mixer.
* in BUILD.api, quota_descriptor.proto is disabled, pending https://github.com/istio/api/pull/20
* parts of pkg/aspectsupport/logger/manager.go are not covered by the existing tests, as it is expected that they will soon be replaced by config-related PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/157)

### DIFF
--- a/adapter/stdioLogger/config/config.proto
+++ b/adapter/stdioLogger/config/config.proto
@@ -48,21 +48,11 @@ message Params {
     // Format of the value of the payload attribute.
     PayloadFormat payload_format = 2;
 
-    // The name of the attribute that will be used to derive log severity.
-    // If this attribute is not found in the set of attributes passed to this
-    // implementation, a default severity of INFO will be used.
-    string severity_attribute = 3;
-
-    // The name of the attribute that will be used to derive the timestamp
-    // for the corresponding log enty. If this attribute is not specified, the
-    // value from time.Now() will be used as the timestamp for the log.
-    string timestamp_attribute = 4;
-
-    // The format of the value of the timestamp_attribute. It is expected that
+    // The format to use when serializing timestamps. It is expected that
     // this format string will follow the conventions of golang's time.Time
     // package and use the reference date as defined in:
     // https://golang.org/pkg/time/#pkg-constants.
     //
-    // Example: "2006-01-02T15:04:05Z07:00"
+    // Default Value: "2006-01-02T15:04:05Z07:00" (RFC3339)
     string timestamp_format = 5;
 }
--- a/adapter/stdioLogger/config/config.proto
+++ b/adapter/stdioLogger/config/config.proto
@@ -31,12 +31,4 @@ message Params {
     // Selects which standard stream to write to for log entries.
     // STDERR is the default Stream.
     Stream log_stream = 1;
-
-    // The format to use when serializing timestamps. It is expected that
-    // this format string will follow the conventions of golang's time.Time
-    // package and use the reference date as defined in:
-    // https://golang.org/pkg/time/#pkg-constants.
-    //
-    // Default Value: "2006-01-02T15:04:05Z07:00" (RFC3339)
-    string timestamp_format = 2;
 }
--- a/adapter/stdioLogger/config/config.proto
+++ b/adapter/stdioLogger/config/config.proto
@@ -38,5 +38,5 @@ message Params {
     // https://golang.org/pkg/time/#pkg-constants.
     //
     // Default Value: "2006-01-02T15:04:05Z07:00" (RFC3339)
-    string timestamp_format = 5;
+    string timestamp_format = 2;
 }
--- a/adapter/stdioLogger/config/config.proto
+++ b/adapter/stdioLogger/config/config.proto
@@ -21,32 +21,16 @@ option go_package = "config";
 message Params {
 
     // Stream is used to select between different logs output sinks.
-    // STDERR is the default Stream.
     enum Stream {
-
         // STDERR refers to os.Stderr.
         STDERR = 0;
-
         // STDOUT refers to os.Stdout.
         STDOUT = 1;
     }
 
     // Selects which standard stream to write to for log entries.
+    // STDERR is the default Stream.
     Stream log_stream = 1;
-
-    // PayloadFormat details the currently supported logging payload formats.
-    // TEXT is the default payload format.
-    enum PayloadFormat {
-
-        // Indicates a payload format of raw text.
-        TEXT = 0;
-
-        // Indicates a structured payload (JSON object).
-        STRUCTURED = 1;
-    }
-
-    // Format of the value of the payload attribute.
-    PayloadFormat payload_format = 2;
 
     // The format to use when serializing timestamps. It is expected that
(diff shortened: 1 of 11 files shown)


douglas-reid:
Updated to reflect commit of istio/api#20.

adapter/stdioLogger/config/config.proto
```diff
@@ -50,30 +45,24 @@ message Params {
...
     // If this attribute is not found in the set of attributes passed to this
     // implementation, a default severity of INFO will be used.
-    string severity_attribute = 5;
+    string severity_attribute = 3;
 
     // The name of the attribute that will be used to derive the timestamp
     // for the corresponding log enty. If this attribute is not specified, the
     // value from time.Now() will be used as the timestamp for the log.
-    string timestamp_attribute = 6;
+    string timestamp_attribute = 4;
```
douglas-reid: Perhaps. I've moved it to aspect config for now. OK with leaving it in for this CL with a subsequent move? We can refactor after this CL (as it involves multiple CLs across repos).

adapter/stdioLogger/stdioLogger.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Package stdioLogger provides an implementation of the mixer logger aspect
-// that writes logs (serialized as JSON) to a standard expectedStream (stdout | stderr).
+// that writes logs (serialized as JSON) to a standard stream (stdout | stderr).
```
douglas-reid: ? this definitely writes structured logs via the writeJSON call.

adapter/stdioLogger/stdioLogger.go
```diff
@@ -81,20 +79,13 @@ func (a *adapter) NewAspect(env aspect.Env, cfg proto.Message) (logger.Aspect, e
...
-	if c.LogName != "" {
-		name = c.LogName
-	}
-
 	pFmt := textFmt
 	if c.PayloadFormat == config.Params_STRUCTURED {
 		pFmt = structFmt
 	}
 
 	return &aspectImpl{
```
douglas-reid: I have a different preference. But I also have a way to work around this issue that I think may satisfy both of us.

OK with leaving it as is for now, and dealing with the fallout during the great re-org next week? I'll draft an example of my idea in time for that discussion.

douglas-reid:
PTAL. I believe this matches this afternoon's attributes discussion.  We can move stuff from aspect config to log_entry_descriptor (like timestamp) in subsequent CLs.

adapter/stdioLogger/stdioLogger.go
geeknoid: My comment relates to the PayloadFormat stuff. What's the different between TEXT and STRUCTURED then?

pkg/aspect/logger/logger.go
```diff
@@ -30,10 +32,23 @@ type (
...
-	// Entry is a set of key-value pairs of attributes that together
-	// constitute the data for log entry. The key is the attribute name,
-	// and the value is a typed value for the named attribute.
-	Entry map[string]interface{}
+	// Entry is the set of data that together constitutes a log entry.
+	Entry struct {
+		// LogName is the name of the log in which to record this entry.
+		LogName string
+		// Labels are a set of metadata associated with this entry.
+		// For instance, Labels can be used to describe the monitored
```
geeknoid: I think we'll want monitored resources to be an explicit field in the Entry struct. I was waiting for more stuff to be in place to start thinking about these things though...

pkg/aspect/logger/logger.go
```diff
@@ -30,10 +32,23 @@ type (
...
+		// resource that corresponds to the log entry.
+		Labels map[string]interface{}
+		// Payload is the actual data for which the entry is being
+		// generated.
+		Payload string
+		// Timestamp is the time value for the log entry
+		Timestamp time.Time
+		// Severity indicates the log level for the log entry.
+		// Example levels: "INFO", "WARNING", "500"
+		Severity string
```
geeknoid: The valid values here need to be well-known and defined up front. So maybe an enum is a better choice?

pkg/aspectsupport/logger/manager.go
```diff
@@ -68,30 +90,100 @@ func (m *manager) NewAspect(c *aspectsupport.CombinedConfig, a aspect.Adapter, e
 		inputs = c.Aspect.Inputs
```
geeknoid: I don't understand what Aspect.Inputs is for. Who fills that in and how does it relate to things like severityAttribute?

douglas-reid: This is the stuff that is generated from an expression.  It is filled in by the operator -- to generate a value from some combination over the original source attributes.

For another example, see: https://github.com/istio/mixer/blob/master/pkg/aspectsupport/listChecker/manager.go#L109

adapter/stdioLogger/stdioLogger.go
douglas-reid: This is about the format of the payload string itself.  It is the difference between:

`"textPayload":"{\"val\":\"42\", \"obj\":{\"val\":\"false\"}}"}`
and
`"structPayload":{"obj":{"val":"false"},"val":"42"}`

in the output log.

However, I think the confusion alone means that this should be aspect configuration and NOT adapter config and that Entry itself should have different values. I'm making those changes now.

pkg/aspect/logger/logger.go
douglas-reid: Let's start with just Labels for now (which supports monitored resources indirectly as is). We can add monitored resources in a subsequent CL, when we have a clearer understanding of whether they should exist across all adapters or not.

pkg/aspect/logger/logger.go
douglas-reid: Do we have to prescribe severity values for all adapters up front?  This was what I was mentioning in our discussion yesterday.  Different systems will have different severity definitions, etc. Letting the individual adapters convert from a string to their individual schema (ints, different enums, etc.) seems more flexible.

geeknoid: Given our discussion, what we want at the moment is to lock down the API to the aspect. So all incoming parameters are well-known and fully specified, no escape hatch. The job of the adapter is to map from this generic API to what its backend expects. The mapping might be lossy and that's OK.

douglas-reid:
PTAL.

Changes:

- `logger.Severity` changed to an enum.
- payload_format has been moved to log_entry_descriptor
- legacy aspect config stuff removed from `BUILD.api`

douglas-reid:
Thanks for the review!

PR #158, opened 2017-01-12
Initial implementation of tracing in the mixer with opentracing

ZackButcher:
Initial implementation of tracing in the mixer with opentracing. In opentracing a tracer starts spans, spans together form a trace. We generate a span which represents the entire stream of calls between a client and mixer, with a child span for each request served by the mixer. We also introduce a client interceptor which propagates the mixer's current span to the callee.

There's still a bit more to do to get opentracing into a better state, but I think this is a reasonable start.

Implementation notes:
- I chose to handle the root span in the GRPCServer's streamLoop method rather than in gRPC's ServerStreamInterceptor (doesn't allow mutation of the context, so we can't propagate span metadata through the mixer) or its transport layer handler (allows mutating the context but
 doesn't bookend the stream so we can't finish the root span) due to the shortcomes mentioned inline.
- Quite a bit of this code should eventually migrate into github.com/grpc-ecosystem/grpc-opentracing.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/158)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -132,6 +132,12 @@ new_go_repository(
     importpath = "github.com/hashicorp/errwrap",
 )
 
+new_go_repository(
+    name = "com_github_opentracing_opentracing_go",
+    commit = "ac5446f53f2c0fc68dc16dc5f426eae1cd288b34",
+    importpath = "github.com/opentracing/opentracing-go",
+)
+
 load("//:repositories.bzl", "new_git_or_local_repository")
 
 new_git_or_local_repository(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -138,6 +138,19 @@ new_go_repository(
     importpath = "github.com/opentracing/opentracing-go",
 )
 
+new_go_repository(
+    name = "com_github_opentracing_basictracer",
+    commit = "1b32af207119a14b1b231d451df3ed04a72efebf",
+    importpath = "github.com/opentracing/basictracer-go"
+)
+
+# Transitive dep of com_github_opentracing_basictracer
+new_go_repository(
+    name = "com_github_gogo_protobuf",
+    commit = "909568be09de550ed094403c2bf8a261b5bb730a",
+    importpath = "github.com/gogo/protobuf"
+)
+
 load("//:repositories.bzl", "new_git_or_local_repository")
 
 new_git_or_local_repository(
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -19,8 +19,6 @@ import (
 	"os"
 
 	"github.com/golang/glog"
-	tracer "github.com/opentracing/basictracer-go"
-	ot "github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
 )
 
@@ -48,6 +46,9 @@ type rootArgs struct {
 
 	// mixerAddress is the full address (including port) of a mixer instance to call.
 	mixerAddress string
+
+	// enableTracing controls whether client-side traces are generated for calls to the mixer.
+	enableTracing bool
 }
 
 // A function used for error output.
@@ -84,20 +85,13 @@ func withArgs(args []string, errorf errorFn) {
 		"List of name/value timestamp attributes specified as name1=value1,name2=value2,...")
 	rootCmd.PersistentFlags().StringVarP(&rootArgs.bytesAttributes, "bytes_attributes", "", "",
 		"List of name/value bytes attributes specified as name1=b0:b1:b3,name2=b4:b5:b6,...")
+	// TODO: implement an option to specify how traces are reported (hardcoded to report to stdout right now).
+	rootCmd.PersistentFlags().BoolVarP(&rootArgs.enableTracing, "trace", "", false,
+		"Whether to trace rpc executions")
 
 	rootCmd.AddCommand(checkCmd(rootArgs, errorf))
 	rootCmd.AddCommand(reportCmd(rootArgs, errorf))
 	rootCmd.AddCommand(quotaCmd(rootArgs, errorf))
-
-	// The tracer will keep track of spans in memory and print them to stdout when the client exits.
-	recorder := tracer.NewInMemoryRecorder()
-	defer func() {
-		for _, span := range recorder.GetSpans() {
-			//fmt.Printf("%v\n", span)
-			glog.Infof("%v\n", span)
-		}
-	}()
-	ot.SetGlobalTracer(tracer.New(recorder))
 
 	if err := rootCmd.Execute(); err != nil {
 		errorf(err.Error())
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -19,6 +19,8 @@ import (
 	"os"
 
 	"github.com/golang/glog"
+	tracer "github.com/opentracing/basictracer-go"
+	ot "github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
 )
 
@@ -86,6 +88,16 @@ func withArgs(args []string, errorf errorFn) {
 	rootCmd.AddCommand(checkCmd(rootArgs, errorf))
 	rootCmd.AddCommand(reportCmd(rootArgs, errorf))
 	rootCmd.AddCommand(quotaCmd(rootArgs, errorf))
+
+	// The tracer will keep track of spans in memory and print them to stdout when the client exits.
+	recorder := tracer.NewInMemoryRecorder()
+	defer func() {
+		for _, span := range recorder.GetSpans() {
+			//fmt.Printf("%v\n", span)
+			glog.Infof("%v\n", span)
+		}
+	}()
+	ot.SetGlobalTracer(tracer.New(recorder))
 
 	if err := rootCmd.Execute(); err != nil {
 		errorf(err.Error())
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -15,12 +15,16 @@
 package main
 
 import (
+	"bufio"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 
+	tracer "github.com/opentracing/basictracer-go"
+	ot "github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
 
 	"istio.io/mixer/pkg/api"
@@ -43,12 +47,31 @@ func serverCmd(errorf errorFn) *cobra.Command {
 		Use:   "server",
 		Short: "Starts the mixer as a server",
 		Run: func(cmd *cobra.Command, args []string) {
-			fmt.Printf("Starting gRPC server on port %v", sa.port)
+			fmt.Printf("Starting gRPC server on port %v, type 'exit' to stop gracefully.\n", sa.port)
 
-			err := runServer(sa)
+			// The tracer will keep track of spans in memory and print them to stdout when the server finishes.
+			recorder := tracer.NewInMemoryRecorder()
+			t := tracer.New(recorder)
+			defer func() {
+				for _, span := range recorder.GetSpans() {
+					fmt.Printf("%v\n", span)
+				}
+			}()
+
+			srv, err := newServer(sa, t)
 			if err != nil {
 				errorf("%v", err)
+				os.Exit(1)
 			}
+
+			go func() {
+				if err := srv.Start(); err != nil {
+					errorf("%v", err)
+					os.Exit(1)
+				}
+			}()
+
+			loopTillStopped(srv)
 		},
 	}
 	serverCmd.PersistentFlags().UintVarP(&sa.port, "port", "p", 9091, "TCP port to use for the mixer's gRPC API")
@@ -62,14 +85,14 @@ func serverCmd(errorf errorFn) *cobra.Command {
 	return &serverCmd
 }
 
-func runServer(sa *serverArgs) error {
+func newServer(sa *serverArgs, tracer ot.Tracer) (*api.GRPCServer, error) {
 	var serverCert *tls.Certificate
 	var clientCerts *x509.CertPool
 
 	if sa.serverCertFile != "" && sa.serverKeyFile != "" {
 		sc, err := tls.LoadX509KeyPair(sa.serverCertFile, sa.serverKeyFile)
 		if err != nil {
-			return fmt.Errorf("failed to load server certificate and server key: %v", err)
+			return nil, fmt.Errorf("failed to load server certificate and server key: %v", err)
 		}
 		serverCert = &sc
 	}
@@ -79,7 +102,7 @@ func runServer(sa *serverArgs) error {
 		for _, clientCertFile := range strings.Split(sa.clientCertFiles, ",") {
 			pem, err := ioutil.ReadFile(clientCertFile)
 			if err != nil {
-				return fmt.Errorf("failed to load client certificate: %v", err)
+				return nil, fmt.Errorf("failed to load client certificate: %v", err)
 			}
 			clientCerts.AppendCertsFromPEM(pem)
 		}
@@ -96,13 +119,25 @@ func runServer(sa *serverArgs) error {
 		ClientCertificates:   clientCerts,
 		Handlers:             api.NewMethodHandlers(),
 		AttributeManager:     attrMgr,
+		EnableTracing:        true,
+		Tracer:               tracer,
 	}
 
 	var grpcServer *api.GRPCServer
 	var err error
 	if grpcServer, err = api.NewGRPCServer(&grpcServerOptions); err != nil {
-		return fmt.Errorf("unable to initialize gRPC server " + err.Error())
+		return nil, fmt.Errorf("unable to initialize gRPC server " + err.Error())
 	}
 
-	return grpcServer.Start()
+	return grpcServer, nil
+}
+
+func loopTillStopped(srv *api.GRPCServer) {
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		if scanner.Text() == "exit" {
+			srv.Stop()
+			return
+		}
+	}
 }
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -15,20 +15,19 @@
 package main
 
 import (
-	"bufio"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"strings"
 
-	tracer "github.com/opentracing/basictracer-go"
+	bt "github.com/opentracing/basictracer-go"
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/spf13/cobra"
 
 	"istio.io/mixer/pkg/api"
 	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/tracing"
 )
 
 type serverArgs struct {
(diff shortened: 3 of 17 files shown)


cmd/client/main.go
```diff
@@ -87,6 +89,16 @@ func withArgs(args []string, errorf errorFn) {
 	rootCmd.AddCommand(reportCmd(rootArgs, errorf))
 	rootCmd.AddCommand(quotaCmd(rootArgs, errorf))
 
+	// The tracer will keep track of spans in memory and print them to stdout when the client exits.
+	recorder := tracer.NewInMemoryRecorder()
+	defer func() {
+		for _, span := range recorder.GetSpans() {
+			//fmt.Printf("%v\n", span)
+			glog.Infof("%v\n", span)
```
geeknoid: We're using fmt.Printf for the main output of a command, and glog only for the error output, so we should be consistent here and use fmt.Printf.

Separately, we can argue whether the way the mixc and mixs handle output is desirable or not, but let's be consistent in the meantime at least.

cmd/client/main.go
```diff
@@ -87,6 +89,16 @@ func withArgs(args []string, errorf errorFn) {
 	rootCmd.AddCommand(reportCmd(rootArgs, errorf))
 	rootCmd.AddCommand(quotaCmd(rootArgs, errorf))
 
+	// The tracer will keep track of spans in memory and print them to stdout when the client exits.
+	recorder := tracer.NewInMemoryRecorder()
```
geeknoid: I think we want tracing to be controlled by a command-line flag, so by default this output is not produced.

cmd/server/server.go
```diff
@@ -43,12 +47,31 @@ func serverCmd(errorf errorFn) *cobra.Command {
 		Use:   "server",
 		Short: "Starts the mixer as a server",
 		Run: func(cmd *cobra.Command, args []string) {
-			fmt.Printf("Starting gRPC server on port %v", sa.port)
+			fmt.Printf("Starting gRPC server on port %v, type 'exit' to stop gracefully.\n", sa.port)
 
-			err := runServer(sa)
+			// The tracer will keep track of spans in memory and print them to stdout when the server finishes.
```
geeknoid: This doesn't seem like an acceptable way to do things in a server that can run indefinitely...

Like in the client, we should make collection of tracing info optional. We probably also should stream the output to a file rather than pile it up in memory.

cmd/server/server.go
```diff
@@ -43,12 +47,31 @@ func serverCmd(errorf errorFn) *cobra.Command {
...
+			defer func() {
+				for _, span := range recorder.GetSpans() {
+					fmt.Printf("%v\n", span)
+				}
+			}()
+
+			srv, err := newServer(sa, t)
 			if err != nil {
 				errorf("%v", err)
+				os.Exit(1)
```
geeknoid: errorf calls os.Exit, so here and elsewhere don't call it elsewhere. The idea is that in a test environment, the test can supply a different errorf function instead of having the code under test call os.Exit.

WORKSPACE
```diff
@@ -132,6 +132,25 @@ new_go_repository(
...
+)
+
+new_go_repository(
+    name = "com_github_opentracing_basictracer",
+    commit = "1b32af207119a14b1b231d451df3ed04a72efebf",
+    importpath = "github.com/opentracing/basictracer-go"
+)
+
+# Transitive dep of com_github_opentracing_basictracer
+new_go_repository(
```
ZackButcher: Yup. We should probably open an issue to resolve this, apparently k8s also uses github.com/gogo/protobuf. If they do we might want to as well.

pkg/api/grpcServer.go
```diff
@@ -76,6 +83,10 @@ type GRPCServerOptions struct {
 	// AttributeManager holds a pointer to an initialized AttributeManager to use when
 	// processing incoming attribute requests.
 	AttributeManager attribute.Manager
+
+	// Tracer is the tracer instance to use with OpenTracing. If EnableTracing is false
```
geeknoid: Instead of EnableTracing and this field, can we just have this field and compare it to nil to see if tracing is desired or not?

pkg/tracing/metadata.go
```diff
@@ -0,0 +1,40 @@
+package tracing
```
geeknoid: Add copyright

pkg/tracing/metadata.go
```diff
@@ -0,0 +1,40 @@
...
+	metadata.MD
+}
+
+func (w metadataReaderWriter) Set(key, val string) {
+	// The GRPC HPACK implementation rejects any uppercase keys here.
+	//
+	// As such, since the HTTP_HEADERS format is case-insensitive anyway, we
+	// blindly lowercase the key (which is guaranteed to work in the
+	// Inject/Extract sense per the OpenTracing spec).
+	key = strings.ToLower(key)
```
geeknoid: This can lead to silent overwriting of keys (FOO vs foo). Maybe this should instead just fail if the incoming key has any uppercase letters?

cmd/server/server.go
```diff
@@ -43,12 +47,31 @@ func serverCmd(errorf errorFn) *cobra.Command {
 		Use:   "server",
 		Short: "Starts the mixer as a server",
 		Run: func(cmd *cobra.Command, args []string) {
-			fmt.Printf("Starting gRPC server on port %v", sa.port)
+			fmt.Printf("Starting gRPC server on port %v, type 'exit' to stop gracefully.\n", sa.port)
 
-			err := runServer(sa)
+			// The tracer will keep track of spans in memory and print them to stdout when the server finishes.
+			recorder := tracer.NewInMemoryRecorder()
```
ZackButcher: We definitely don't want to long term, but as a first impl for mixc/mixs I think it's fine for now (it's also the only way the basictracer impl records spans). Ultimately selecting the tracer will need to be config driven.

Another option would be implementing a simple recorder which dumps to stdout or glog and doesn't hold anything in memory. Personally I like this option and it simplifies some of the server code (no need to stop the server gracefully, so cmd/server/server.go goes back to almost exactly how it was before this PR).

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,113 @@
+package tracing
```
geeknoid: Add copyright.

pkg/tracing/BUILD
```diff
@@ -0,0 +1,32 @@
...
+        "@com_github_opentracing_opentracing_go//log:go_default_library",
+        "@org_golang_google_grpc//:go_default_library",
+        "@org_golang_google_grpc//metadata:go_default_library",
+        "@org_golang_google_grpc//tap:go_default_library",
+        "@org_golang_x_net//context:go_default_library",
+    ],
+)
+
+go_test(
+    name = "tracing_test",
```
ZackButcher: All of the test targets except adapter/stdioLogger are named `<pkg name>_test`. I'm fine with either but as-is this naming is more consistent with the rest of the codebase.

pkg/tracing/metadata.go
ZackButcher: Sure. This was taken directly from [otgrpc](https://github.com/grpc-ecosystem/grpc-opentracing/blob/master/go/otgrpc/shared.go) and ideally they'll take a PR to rework their code so we don't need to replicate this.

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,113 @@
...
+
```
ZackButcher: Done

pkg/tracing/tracing.go
ZackButcher: Done

pkg/tracing/metadata.go
ZackButcher: Done

pkg/tracing/tracing.go
ZackButcher: Done

pkg/tracing/metadata.go
ZackButcher: Done

ZackButcher:
Here's mixs example spans for mixc making a check call to mixs:

    2017-01-12 15:10:01.0511497 -0800 PST /istio.mixer.v1.Mixer/Check-1 101.428µs trace: 3113216478738902926; span: 8637876685754802239; parent: 8576519700601980156; tags: map[]; logs: [{2017-01-12 15:10:01.051153826 -0800 PST [gRPC request:attribute_update:<dictionary:<key:0 value:"serviceName" > dictionary:<key:1 value:"peerId" > string_attributes:<key:0 value:"foo" > string_attributes:<key:1 value:"bar" > > ]} {2017-01-12 15:10:01.051249355 -0800 PST [gRPC response:result:<code:12 > ]}]
    2017-01-12 15:10:01.050843885 -0800 PST /istio.mixer.v1.Mixer/Check 563.719µs trace: 3113216478738902926; span: 8576519700601980156; parent: 0; tags: map[span.kind:server component:gRPC]; logs: []

Where `/istio.mixer.v1.Mixer/Check` is the root span for the stream and `/istio.mixer.v1.Mixer/Check-#` are the children spans (with # incrementing with each message received).

mixc's single span for the same call is:

    2017-01-12 15:10:01.049575332 -0800 PST /istio.mixer.v1.Mixer/Check 1.094139ms trace: 5061861906585099018; span: 357884457575469219; parent: 0; tags: map[span.kind:client]; logs: []

I need to dig into why the parent wasn't propagated in the call, but that doesn't need to block the rest of the PR I think.

----------------------
As far as [WrappedServerStream](https://github.com/mwitkow/go-grpc-middleware/blob/master/doc.md#type-wrappedserverstream) is concerned, the only thing that would let us do is remove [these two lines in streamLoop](https://github.com/ZackButcher/mixer/blob/opentracing/pkg/api/grpcServer.go#L160-L161). I'm pro-using it if we need the mutable context elsewhere, otherwise I don't have a strong feeling either way.

ZackButcher:
Also updated how reporting is handled in mixs/mixc, PTAL.

cmd/client/main.go
```diff
@@ -82,6 +85,9 @@ func withArgs(args []string, errorf errorFn) {
 		"List of name/value timestamp attributes specified as name1=value1,name2=value2,...")
 	rootCmd.PersistentFlags().StringVarP(&rootArgs.bytesAttributes, "bytes_attributes", "", "",
 		"List of name/value bytes attributes specified as name1=b0:b1:b3,name2=b4:b5:b6,...")
+	// TODO: implement an option to specify how traces are reported (hardcoded to report to stdout right now).
+	rootCmd.PersistentFlags().BoolVarP(&rootArgs.enableTracing, "trace", "", true,
```
geeknoid: I think this should default to false.

cmd/server/server.go
```diff
@@ -85,6 +91,13 @@ func runServer(sa *serverArgs) error {
 		}
 	}
 
+	var tracer ot.Tracer
+	if sa.enableTracing {
+		tracer = bt.New(tracing.StdoutRecorder())
+	} else {
+		tracer = ot.NoopTracer{}
```
geeknoid: Should this just be nil, given the code in grpcServer?

pkg/tracing/basictracing.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2016 Google Inc.
```
geeknoid: 2017

pkg/tracing/metadata.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2016 Google Inc.
```
geeknoid: 2017

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2016 Google Inc.
```
geeknoid: 2017

douglas-reid:
Thanks for the example! That is a tremendous help.

WORKSPACE
geeknoid: I opened an issue to switch our codegen to gogoprotobuf. It definitely seems to provide quite an improvement both in usability and in perf of marshaling/unmarshaling.

pkg/api/grpcServer.go
```diff
@@ -155,14 +169,22 @@ func (s *GRPCServer) streamLoop(stream grpc.ServerStream, request proto.Message,
 			return err
 		}
 
+		var span ot.Span
```
geeknoid: Shouldn't the span-related code in this function be conditional on whether tracing is enabled or not? This seems like a potentially large amount of work to be doing in the primary API dispatch loop.

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return noopSpan
+}
+
+// StartRootSpan creates a span that is the root of all Istio spans in the current request context. This span will be a
+// child of any spans propagated to the server in the request's metadata. The returned span is retrievable from the
+// context via tracing.RootSpan.
+func StartRootSpan(ctx context.Context, operationName string) (ot.Span, context.Context) {
+	md, ok := metadata.FromContext(ctx)
+	if !ok {
```
geeknoid: Combine onto previous line

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,132 @@
...
+		md = metadata.New(nil)
+	}
+	spanContext, err := ot.GlobalTracer().Extract(ot.TextMap, metadataReaderWriter{md})
+	if err != nil {
```
geeknoid: Combine onto previous line

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// propagateSpan inserts metadata about the span into the context's metadata so that the span is propagated to the receiver.
+// This should be used to prepare the context for outgoing calls.
+//
+// TODO: consider creating a public version of this method for use at individual call sites if the interceptor isn't
+// sufficient for some reason.
+func propagateSpan(ctx context.Context, tracer ot.Tracer, span ot.Span) context.Context {
+	md, ok := metadata.FromContext(ctx)
+	if !ok {
```
geeknoid: Combine onto previous line

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,132 @@
...
+		md = metadata.New(nil)
+	}
+	err := tracer.Inject(span.Context(), ot.TextMap, metadataReaderWriter{md})
+	if err != nil {
```
geeknoid: Combine onto previous line

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2017 Google Inc.
```
geeknoid: Can you get code coverage to 100% in this file and in metadata.go? You're pretty close, just a few more edge cases to handle.

pkg/api/grpcServer.go
ZackButcher: Opentracing is written to default to a no-op tracer, so the call to StartSpanFromContext amounts to [extracting a key from the context](https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L43) and [returning a static value (the no-op span)](https://github.com/opentracing/opentracing-go/blob/master/noop.go#L52). This shouldn't even result in an allocation, other than the span we declare here (and we could even hoist this declaration out of the loop so the allocation only occurs at stream creation, though I'd expect the compiler to perform that optimization already). [Calls on the no-op span themselves do no work](https://github.com/opentracing/opentracing-go/blob/master/noop.go#L37-L49).

I thought that was sufficiently low overhead that keeping the code simpler by not introducing branches was worth it. A different option we have is to keep the is-tracing-enabled state in some object that's a wrapper around the tracer, and use that object to create spans. It could short-circuit to the no-op span if tracing isn't enabled, while still keeping this code conceptually condition free.

geeknoid: Although the span calls may themselves be nops, I was worried about the ancillary calls that are used to prepare the parameters for the span calls, like log.Object and fmt.Sprintf. 

I just looked at log.Object and it's a cheap call, so that's fine. But fmt.Sprintf is very expensive and should definitely be hoisted outside of the loops.

pkg/api/grpcServer.go
```diff
@@ -155,14 +169,22 @@ func (s *GRPCServer) streamLoop(stream grpc.ServerStream, request proto.Message,
...
+		span, ctx = ot.StartSpanFromContext(ctx, fmt.Sprintf("%s-%d", methodName, i))
+		span.LogFields(log.Object("gRPC request", request))
```
douglas-reid: How important do we think logging the fields will be?  I don't know that this matters immediately, but I imagine these requests can get quite large.  Should we make logging request and response fields optional (like a debug mode)? 

I think the real value will be in the spans themselves.

pkg/api/grpcServer.go
ZackButcher: I see two options:
1. Remove the dynamic span name, and only create spans using `methodName`. No `fmt.Sprintf` or string concatenation, but it's a bit harder for humans to parse the spans (each span still gets a unique span ID, and the start time/duration are recorded, so still possible to piece together the sequencing).
2. Keep the dynamic name, replace the `fmt.Sprintf` with a string concatenation.

Which do you prefer?

geeknoid: I'd go with 1) for now. If someone asks, we can then see if adding an integer counter would be useful.

pkg/tracing/tracing.go
ZackButcher: Refactored this into a method since it's repeated. Side effect is these `if`s go away.

pkg/api/grpcServer.go
ZackButcher: My long term plan is that logging the request/response is controlled by a user option (otgrpc exposes this, but I haven't come up with a clean way to integrate it into our streaming code). I agree we probably don't want to be logging req/resp in general.

pkg/tracing/tracing.go
ZackButcher: Got as close as is reasonable: the only branches not covered are error cases in the tracer impl and in gRPC's metadata class not maintaining its own invariants. To do more would require implementing fake tracer and metadata classes, which I don't think is worth the effort.

pkg/api/grpcServer.go
ZackButcher: I can remove logging req/resp in this PR and we can add it back when we get around to implementing tracing options in general.

pkg/api/grpcServer.go
```diff
@@ -140,13 +151,16 @@ func (s *GRPCServer) Stop() {
...
-func (s *GRPCServer) streamLoop(stream grpc.ServerStream, request proto.Message, response proto.Message, handler handlerFunc) error {
+func (s *GRPCServer) streamLoop(stream grpc.ServerStream, request proto.Message, response proto.Message, handler handlerFunc, methodName string) error {
 	tracker := s.attrMgr.NewTracker()
 	defer tracker.Done()
 
-	for {
+	root, ctx := tracing.StartRootSpan(stream.Context(), methodName)
+	defer root.Finish()
+
+	for i := 1; ; i++ {
```
geeknoid: I think we don't need to count "i" anymore, right?

pkg/tracing/basictracing.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2017 Google Inc.
```
geeknoid: This file has 0% test coverage. Can a quick test handle this?

ZackButcher: Derp, done. This resulted in me updating the stdout recorder to take an io.Writer instead.

PR #159, opened 2017-01-12
Add metrics aspect to mixer.

douglas-reid:
This PR adds the metrics aspect to mixer. This will allow the development of metrics backends that generate stats from Report() calls.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/159)

### DIFF
--- a/pkg/aspect/metrics/metrics.go
+++ b/pkg/aspect/metrics/metrics.go
@@ -82,6 +82,7 @@ type (
 	}
 )
 
+// String returns the string-valued metric value for a metrics.Value.
 func (v Value) String() (string, error) {
 	if v, ok := v.metricValue.(string); ok {
 		return v, nil
@@ -89,6 +90,7 @@ func (v Value) String() (string, error) {
 	return "", errors.New("metric value is not a string")
 }
 
+// Bool returns the boolean metric value for a metrics.Value.
 func (v Value) Bool() (bool, error) {
 	if v, ok := v.metricValue.(bool); ok {
 		return v, nil
@@ -96,6 +98,7 @@ func (v Value) Bool() (bool, error) {
 	return false, errors.New("metric value is not a boolean")
 }
 
+// Int64 returns the int64-valued metric value for a metrics.Value.
 func (v Value) Int64() (int64, error) {
 	if v, ok := v.metricValue.(int64); ok {
 		return v, nil
@@ -103,6 +106,7 @@ func (v Value) Int64() (int64, error) {
 	return 0, errors.New("metric value is not an int64")
 }
 
+// Float64 returns the float64-valued metric value for a metrics.Value.
 func (v Value) Float64() (float64, error) {
 	if v, ok := v.metricValue.(float64); ok {
 		return v, nil
--- a/pkg/aspect/metrics/metrics.go
+++ b/pkg/aspect/metrics/metrics.go
@@ -18,6 +18,7 @@
 package metrics
 
 import (
+	"errors"
 	"time"
 
 	"github.com/golang/protobuf/proto"
@@ -42,7 +43,7 @@ type (
 		Record([]Value) error
 	}
 
-	// Value holds an single metric value that will be generated through
+	// Value holds a single metric value that will be generated through
 	// a Report() call to the mixer. It is synthesized by the mixer, based
 	// on mixer config and the attributes passed to Report().
 	Value struct {
@@ -54,14 +55,6 @@ type (
 		// Labels provide metadata about the metric value. They are
 		// generated from the set of attributes provided by Report().
 		Labels map[string]interface{}
-		// StringValue is used to pass a string-valued metric value.
-		StringValue string
-		// Int64Value is used to pass a integer-valued metric value.
-		Int64Value int64
-		// Float64Value is used to pass a double-valued metric value.
-		Float64Value float64
-		// BoolValue is used to pass a boolean-valued metric value.
-		BoolValue bool
 		// StartTime marks the beginning of the period for which the
 		// metric value is being reported. For instantaneous metrics,
 		// StartTime records the relevant instant.
@@ -70,6 +63,8 @@ type (
 		// value is being reported. For instantaneous metrics, EndTime
 		// will be set to the same value as StartTime.
 		EndTime time.Time
+
+		metricValue interface{}
 	}
 
 	// Kind defines the set of known metrics types that can be generated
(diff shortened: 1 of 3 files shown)


pkg/aspect/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	// reporting within the mixer.
+	Aspect interface {
+		aspect.Aspect
+
+		// Record directs a backend adapter to record the list of values
+		// that have been generated from Report() calls.
+		Record([]Value) error
+	}
+
+	// Value holds an single metric value that will be generated through
```
geeknoid: an -> a

douglas-reid:
*[pkg/aspect/metrics/metrics.go, line 30 at r1](https://reviewable.io:443/reviews/istio/mixer/159#-KaIzB0rjM9iEIcwtHj-:-KaiGqdwtzjM3MXU5ih8:bxu03nd) ([raw file](https://github.com/istio/mixer/blob/a85b1023a9e4eba21d7f6a157c0e7790531fecc8/pkg/aspect/metrics/metrics.go#L30)):*
<details><summary><i>Previously, ZackButcher (Zack) wrote…</i></summary><blockquote>

> COUNTER is used to record increasing cumulative values.

Some metric backends allow decrementing counters (e.g. statsd), how do we plan to handle that case?
</blockquote></details>

At the moment, we don't have any plans for that.  I'd say we should address it when we compare schemas across metrics systems.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159)*

douglas-reid:
*[pkg/aspect/metrics/metrics.go, line 45 at r1](https://reviewable.io:443/reviews/istio/mixer/159#-KaIxJ3hUYEXEFQD8cF--r1-45:-KaiHI4pb9_X8-6TVbxP:b-896fix) ([raw file](https://github.com/istio/mixer/blob/a85b1023a9e4eba21d7f6a157c0e7790531fecc8/pkg/aspect/metrics/metrics.go#L45)):*

an -> a
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159)*

douglas-reid:
*[pkg/aspect/metrics/metrics.go, line 58 at r1](https://reviewable.io:443/reviews/istio/mixer/159#-KaIzhQPNQjadvOUvvaV:-KaiHJi_Bus066ss967o:baoh1iq) ([raw file](https://github.com/istio/mixer/blob/a85b1023a9e4eba21d7f6a157c0e7790531fecc8/pkg/aspect/metrics/metrics.go#L58)):*
<details><summary><i>Previously, ZackButcher (Zack) wrote…</i></summary><blockquote>

> // StringValue is used to pass a string-valued metric value.
> 		StringValue string
> 		// Int64Value is used to pass a integer-valued metric value.
> 		Int64Value int64
> 		// Float64Value is used to pass a double-valued metric value.
> 		Float64Value float64
> 		// BoolValue is used to pass a boolean-valued metric value.
> 		BoolValue bool

I don't like exposing the inner value fields publicly, I prefer providing a set of typesafe functions for handling the values in a metric. [opentracing's log package is very similar](https://github.com/opentracing/opentracing-go/blob/master/log/field.go).

I imagine something like:

    func StringValue(name string, kind Kind, labels map[string]interface{}, value string, start time.Time, end time.Time) Value {...}
    func Int64Value(name string, kind Kind, labels map[string]interface{}, value int64, start time.Time, end time.Time) Value {...}

    func (v Value) Extract() interface{} {
        return v.innerValue
    }
    
    func String(v Value) (string, err) {
       if val, ok := v.Extract().(string); ok {
          return val, nil
       }
       return nil, fmt.Errorf("Value is not a string")
    }

    func Int64(v Value) (int64, err) {
       if val, ok := v.Extract().(int64); ok {
          return val, nil
       }
       return nil, fmt.Errorf("Value is not an int64")
    }
</blockquote></details>

OK.  Add the methods as suggested to Value itself. Please take a look.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159)*

pkg/aspect/metrics/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+		aspect.Adapter
+
+		// NewAspect returns a new quota implementation, based on the
+		// supplied Aspect configuration for the backend.
+		NewAspect(env aspect.Env, config proto.Message) (Aspect, error)
+	}
+)
+
+// String returns the string-valued metric value for a metrics.Value.
+func (v Value) String() (string, error) {
```
ZackButcher: To capture some of the talk in slack: upon thinking about this more, I'm not sure I'm convinced it's an improvement over making the user typecast themselves: we essentially changed the interface from

    s, ok := v.val.(string)
    if !ok { ...}

to

    s, err := v.String()
    if err != nil { ... }

However, this is easy enough to fix later, so I think we should proceed and we can reevaluate down the line.

ZackButcher:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159)*

ZackButcher:
:lgtm:

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159#-:-KaiQK2dGPYsyZ9B1x51:bnfp4nl)*

douglas-reid:
*[pkg/aspect/metrics/metrics.go, line 86 at r2](https://reviewable.io:443/reviews/istio/mixer/159#-KaIxJ3hUYEXEFQD8cF--r2-86:-KaiU-aes8qyFR_si6Y4:bb6rskq) ([raw file](https://github.com/istio/mixer/blob/84227392b4cf5dbfb92dfbdf68d3eaa561bc791c/pkg/aspect/metrics/metrics.go#L86)):*
<details><summary><i>Previously, ZackButcher (Zack) wrote…</i></summary><blockquote>

To capture some of the talk in slack: upon thinking about this more, I'm not sure I'm convinced it's an improvement over making the user typecast themselves: we essentially changed the interface from

    s, ok := v.val.(string)
    if !ok { ...}

to

    s, err := v.String()
    if err != nil { ... }

However, this is easy enough to fix later, so I think we should proceed and we can reevaluate down the line.
</blockquote></details>

ACK.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159)*

geeknoid:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/159#-:-KaiV7IyEXm-sGGekzr3:bnfp4nl)*

PR #165, opened 2017-01-18
Wire up the mixer to handle requests with adapters

ZackButcher:
Wire up a version of the Istio Mixer that uses the DenyChecker adapter impl to reject all Mixer.Check calls. We introduce a new api.StaticHandlers struct which is an implementation of api.MethodHandlers. This implementation is configured at compile time. This exercises pkg/registry/registrar codepaths in the adapter's impl and the pkg/aspectsupport/uber/manager. We also introduce a rudimentary Registry which is a simple wrapper around a map[string]aspect.Adapter.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/165)


ZackButcher:
Two questions:
- Should staticHandlers.go just be collapsed into methodHandlers.go?
- What do we do with erroneous linter complaints? In the registry impl the linter wants Register* methods to accept an aspect.Adapter instead of a specific type of Adapter (e.g. listChecker.Adapter). We must accept the specific version to satisfy the Registrar interface, but the linter cannot identify that.

pkg/expr/identity.go
```diff
@@ -0,0 +1,69 @@
...
+type identity struct{}
+
+// NewIdentityEvaluator returns an evaluator that performs no evaluations; instead it uses the provided mapExpression
+// as the key into the attribute bag.
+func NewIdentityEvaluator() Evaluator {
+	return identity{}
+}
+
+// Eval attempts to extract the key `mapExpression` from the attribute bag. It performs no evaluation.
+func (identity) Eval(mapExpression string, bag attribute.Bag) (interface{}, error) {
```
douglas-reid: this looks a lot like `attribute.Value(attribute.Bag, name string)` from bag.go.  Can we reuse Value() and translate the results to have the same impact?

pkg/api/staticHandlers.go
```diff
@@ -0,0 +1,170 @@
...
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/ptypes/struct"
+
+	"google.golang.org/genproto/googleapis/rpc/code"
+	"google.golang.org/genproto/googleapis/rpc/status"
+
+	denyadapter "istio.io/mixer/adapter/denyChecker"
```
douglas-reid: thinking out loud: do we need StaticHandlers to have denyChecker hardcoded in? What if we changed `NewStaticHandlers()` to be something like:
```
type StaticBinding struct {
    RegisterFn func(r registry.Registrar) error
    Manager aspectsupport.Manager
    // related CombinedConfig stuff here too, I guess.
}

func NewStaticHandlers([]StaticBinding) *StaticHandlers {
...
    if err := binding.RegisterFn(registry); err != nil {
...
    managers.append(managers, binding.Manager)
...
```

Then we could have //cmd/server depend on //adapters for the time being and avoid having //pkg/api depend on //adapters (maybe not all that important, but feels useful).

It might even allow us to build a noconfig_server.go which we run until the config-full version is ready.

Thoughts? Things I've missed?

ZackButcher: I think this is spot on. One of the reasons I created a new class rather than modify methodHandlers was because it felt weird having //pkg/api depend //adapters. I think I'll update this PR to:
- Use the ubermanager in methodHandlers.go, combining the existing methodHandlers with the changes in staticHandlers.go relating to the uber manager and request dispatch.
- Have the methodHandlers constructor accept a config struct like you described

This should mean a lot less throwaway code and the correct dependencies (//pkg/api doesn't know about //adapters, it's //cmd/server that wires them together).

pkg/expr/identity.go
ZackButcher: Ah; I typically do method discovery via auto-complete so I only looked at attribute.Bag. Didn't even think to look at the free funcs in the package. Updated.

ZackButcher:
Updated per your comments, PTAL. I folded staticHandlers into methodHandlers, and changed mixs provide the static config when it sets up the grpc server.

pkg/aspectsupport/uber/registry.go
```diff
@@ -0,0 +1,78 @@
...
+func NewRegistry() *Registry {
+	return &Registry{adaptersByName: make(map[string]aspect.Adapter)}
+}
+
+// ByImpl returns the implementation with aspect.Adapter.Name() == adapterName.
+func (r *Registry) ByImpl(adapterName string) (aspect.Adapter, bool) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	adapter, ok := r.adaptersByName[adapterName]
+	return adapter, ok // yet `return r.adaptersByName[adapterName]` doesn't typecheck.
```
geeknoid: I've had this problem too before. Feels like a compiler bug to me...

pkg/expr/identity.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+
+	"istio.io/mixer/pkg/attribute"
+)
+
+// identity is an evaluator that expects mapExpression to be a key in the attribute bag. It does no evaluation, only lookup.
+type identity struct{}
+
+// NewIdentityEvaluator returns an evaluator that performs no evaluations; instead it uses the provided mapExpression
```
geeknoid: Can this just be merged in the normal expression evaluator?

@mandarjog As we discussed, an expression should be either a lone attribute name, a litteral value, or a complex expression that produces an attribute value.

ZackButcher: Do we have a normal evaluator already? I only created this because I needed *some* evaluator and this was the simplest case to implement.

ZackButcher:
Updated per comments, PTAL. As an aside, I don't like how the github UI handles a rebase + push --force into my remote.

ZackButcher:
Oh sorry, I should've asked: what's our stance on merging in: should I have waited for other LGTM's, and should hit the actual merge button or the final approver, or what?

douglas-reid:
I raised this issue awhile back. Perhaps we should adopt a primary reviewer (assignee) and give them final authority. However, I do think that when there are pending comments/questions from other reviewers, we should hold merge until they have had a chance to respond to the updated commits.

Issue #164, opened 2017-01-18
Add AccessLogger Aspect to Mixer


douglas-reid:
Tracking issue

Mixer will soon have a (application) logger aspect. We should consider adding an access log aspect that provides a `LogAccess([]accesslogger.Entry)` method for adapters.

douglas-reid:
This aspect has been added -- and a manager for the aspect has been added as well. Marking as done.

PR #172, opened 2017-01-19
Add accessLogger aspect to mixer.

douglas-reid:
This PR establishes an access logging aspect (as distinct from the generic logger aspect) for use by Istio. This aspect will be used to generate access logs (think [Common Logging Format](https://en.wikipedia.org/wiki/Common_Log_Format)).

Addresses issue #164.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/172)


pkg/aspect/accessLogger/accessLogger.go
```diff
@@ -0,0 +1,63 @@
...
+	// a single log entry. It is the job of the aspect manager to produce
+	// the access log string, based on a configuration and pass the entries
+	// to the various backend adapters.
+	//
+	// Example: &Entry{
+	// 	LogName: "logs/access_log",
+	// 	Log: "127.0.0.1 - testuser [10/Oct/2000:13:55:36 -0700] "GET /test.gif HTTP/1.0" 200 2326",
+	// }
+	Entry struct {
+		// LogName is the name of the access log stream to which the
```
mandarjog: 1. If this is the textAccessLog aspect, then `Entry struct` is ok.
2. Need to define the AspectConfig proto.   It will have `accessLogFormat   - %v %H ...`

If however we also want to support `json`, yml etc we could do it this way.
```
AspectConfig {
     repeated string labelnames
}
Entry {
   map[string] LabelValues
}
```

Then
```
TextAccessLogImpl will define 
LogFormat 

JsonAccessLogImpl may define something else, similarly yaml accessLog.

**Comment by douglas-reid** [comment]
PTAL. 

I've updated `aspectlogger.Entry` to include a `Labels` field. 

I believe this matches our discussion from this afternoon. This should enable aspect config to control LogFormat AND Label selection (attribute downscoping) without prescribing how the Access Log entries will be encoded by the backend adapters.

Also, as discussed, for now, Labels is `map[string]interface{}`. We will likely want to move to something like a scoped attribute bag in the near future.

**Comment by douglas-reid** [comment]
This (and the associated issue) came out of discussion at the stand-up Tuesday. It was felt that having a distinct accessLogger aspect was a key point in the design reviews and is separate (in users' minds) from generic logger.

There is certainly going to be a lot of overlap -- but it is likely we can provide better defaults for accessLogger than we can for generic logger, etc.

Thoughts?

**Comment by geeknoid** [comment]
A logger is a logger is a logger.

If we want to provide an explicit stackDriverAccessLogger adapter, that's
all fine and dandy. But I don't see why the interface to this adapter needs
to be different than the interface to other logging adapters.

Let me put this another way: If this case requires a new aspect manager,
than we should stop pretending that we actually have an abstraction layer.
We should just get rid of aspect managers as a whole and put all the logic
directly into the adapters. I don't want a whole bunch of aspect managers
and just 1 corresponding adapter for each. A single aspect must be able to
cover multiple distinct adapters.

**Comment by douglas-reid** [comment]
I'm confused a bit by the mapping you are suggesting.  In my head, we have `accessLogger` and `logger` aspects.  Then, `stdioLogger` adapter implements both aspects. It would provide both `Log([]logger.Entry)` and `LogAccess([]accessLogger.Entry)`.

We could definitely have the logger aspect manager take config that covers both types of logging, but I'm not sure that makes things simpler / easier to understand.

For instance, I don't think we will have access logs described by `LogEntryDescriptor`s (they don't have `payloads` in the same sense). We _could_ add the format bits to `logger` config and have it generate `TextPayload`s and cram the access log into the `logger.Entry` struct.  Having `accessLogger` separates the two and _might_ make it simpler to understand the two distinct classes of logs that flow through the system.

Let me finish the `accessLogger` config proto and add it to this PR and then we can discuss how this would co-exist with `logger`. Sound OK?

**Comment by mandarjog** [comment]
Since aspects are user visible we should operate at a level of abstraction that our users are comfortable with.  `AccessLogs` and `Logs` are 2 very different things from an end user perspective.  Logs are free form, and initiated by intentionally thru a log.Info or similar statement. 

`AccessLogs` on the other hand have well known formats. (Apache / common / combined etc)

So I think accesslog as a separate aspect makes sense.

**Comment by douglas-reid** [comment]
OK. I've added an initial stab at what the aspect configuration would look like for an aspectLogger aspect.

PTAL and let me know how this looks.

File: pkg/aspectsupport/accessLogger/config/config.proto [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+    repeated string attributes = 2;
+
+    // This will format string will be applied to the set of
+    // attributes specified to generate a textual access log entry.
+    //
+    // If this is not specified, the mixer will attempt to use a
+    // default format that closely aligns with the Common Log Format.
+    //
+    // Default:
+    // `{{or (.source_ip) "-"}} - {{or (.source_user) "-"}} [{{.timestamp.Format "02/Jan/2006:15:04:05 -0700"}}] "{{or (.method) "-"}} {{or (.api_resource) "-"}} {{or (.protocol) "-"}}" {{or (.status_code) "-"}} {{or (.response_size) "-"}}`
```
mandarjog: go template expression language is not the easiest to read, but for now it gets the point across.
Once we have an expression evaluator, we can use the same language.

douglas-reid: Agreed. I updated the comments to call out the template documentation. Anything that can simplify would be great.

douglas-reid:
@geeknoid should we have a quick chat re: accessLogger aspect. I want to make sure we are all in agreement before moving forward.

pkg/aspectsupport/accessLogger/config/config.proto
```diff
@@ -0,0 +1,54 @@
...
+
+    // A list of names of attributes that will be used to generate
+    // access log entries.
+    repeated string attributes = 2;
+
+    // This template will be applied to the map of attributes names to values
+    // generated by the mixer from the specified set of attributes
+    // to create a textual access log entry.
+    //
+    // The template MUST be specified in a format compatible with:
```
geeknoid: I have strong reservations whether this is appropriate for human-beings that aren't go programmers. An operator trying to use Istio will be mystified by this stuff.

geeknoid:
Nah, I give up. I don't believe it is desirable or necessary, but I think
we need to move forward a bit for this to become evident to everyone else.
So let's move forward, and then we'll delete it :-)

pkg/aspectsupport/accessLogger/config/config.proto
douglas-reid: I totally understand. At the moment, however, I'm not sure how to best "hide" this bit.

We'd need something that maps to/from attribute values to [Custom Log Formats](https://httpd.apache.org/docs/1.3/mod/mod_log_config.html#formats) (or similar) to allow the operator to specify in the language they are more familiar with (something like: `"%h %l %u %t \"%r\" %>s %b"`).

If you have a clever approach for that, I'd be very interested.

Perhaps we could maintain a mapping in the manager (once we have a schema for well-known attributes)?

mandarjog:
Once we have the expression language in place, there could be convergence here.
Until then go template, as painful as it is to use, is easy to implement.

PR #173, opened 2017-01-19
Add a Tracer struct to the tracing package

ZackButcher:
This new struct wraps an opentracing.Tracer with a flag that's true if tracing is enabled in the mixer. This flag is used to short-circuit potentially expensive tracing operations if tracing is disabled. In the future this struct will also hold user-specified tracing options; it also lays the groundwork for being able to force tracing on a per-request basis.

We also rework the tracing package given newly discovered info, namely that the gRPC's StreamClientInterceptor does *not* behave in the same way as the StreamServerInterceptor. It looks like we won't be able to use gRPC's interceptor machinery to handle tracing. As a result, we implement span creation and propagation directly in mixc's check and quota commands.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/173)

### DIFF
--- a/cmd/client/BUILD
+++ b/cmd/client/BUILD
@@ -20,6 +20,7 @@ go_library(
         "@com_github_istio_api//:mixer/v1",
         "@com_github_opentracing_basictracer//:go_default_library",
         "@com_github_opentracing_opentracing_go//:go_default_library",
+        "@com_github_opentracing_opentracing_go//ext:go_default_library",
         "@com_github_spf13_cobra//:go_default_library",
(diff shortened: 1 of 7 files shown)


douglas-reid:
Any idea why this PR is including other commits?  Did we need to rebase against upstream/master or something first? It makes it hard to look at the changes contained in this PR itself.

ZackButcher:
Those subsequent commits are me attempting to fix, not sure how I screwed up.

ZackButcher:
Ok, back to sanity. PTAL.

PR #176, opened 2017-01-20
Add workaround for glog to work properly with the cobra command-line …

geeknoid:
…parser.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/176)

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -65,7 +65,13 @@ func withArgs(args []string, errorf errorFn) {
 		Short: "Invoke the API of a running instance of the Istio mixer",
 	}
 	rootCmd.SetArgs(args)
-	rootCmd.Flags().AddGoFlagSet(flag.CommandLine)
+	rootCmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
+
+	// hack to make flag.Parsed return true such that glog is happy
+	// about the flags having been parsed
+	fs := flag.NewFlagSet("", flag.ContinueOnError)
+	_ = fs.Parse([]string{})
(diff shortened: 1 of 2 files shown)


ZackButcher:
Oh good, the glog flags-not-parsed spam is super annoying.

cmd/client/main.go
```diff
@@ -65,7 +65,13 @@ func withArgs(args []string, errorf errorFn) {
 		Short: "Invoke the API of a running instance of the Istio mixer",
 	}
 	rootCmd.SetArgs(args)
-	rootCmd.Flags().AddGoFlagSet(flag.CommandLine)
+	rootCmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
+
+	// hack to make flag.Parsed return true such that glog is happy
+	// about the flags having been parsed
+	fs := flag.NewFlagSet("", flag.ContinueOnError)
```
kyessenov: Can you explain why this is needed?
I have almost exact same code, and it works well with glog.

geeknoid:
​As mentioned in my previous comment, glog has some code to check whether
flag.Parsed() returns true. When using cobra, the flag.Parse() method is
normally never called and hence flag.Parsed() always returns false. When
you try to output with glog, it will then always print a message saying
"Parse not called" and then assume some default behavior. The reality is of
course that the flags have been parsed, that the values the glog is
interested in reading have all been populated. But they've been populated
through cobra/pflags and not the normal flag package.

So this hack is just enough to have flag.Parsed() return true to glog so
that glog works as we want it to.​

PR #177, opened 2017-01-20
Make the mixer robust to adapter panics and end processing early on timeout

ZackButcher:
Wrap the dispatch to our adapters in a function which will recover them panicking. This wrapper also polls for the cancellation of the request's context and aborts early if it is canceled (e.g. because we exceeded the requests's deadline).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/177)

### DIFF
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -95,11 +95,7 @@ func (m *Manager) Execute(cfg *aspect.CombinedConfig, attrs attribute.Bag, mappe
 		return nil, fmt.Errorf("could not find registered adapter %#v", cfg.Builder.Impl)
 	}
 
-	var asp aspect.Wrapper
-	if asp, err = m.cacheGet(cfg, mgr, adapter); err != nil {
-		return nil, err
-	}
-
+	// Both cacheGet and asp.Execute call adapter-supplied code, so we need to guard against both panicking.
 	defer func() {
 		if r := recover(); r != nil {
 			out = nil // invalidate whatever partial result we got; should we set this to a Code.Code_INTERNAL or similar?
@@ -107,6 +103,12 @@ func (m *Manager) Execute(cfg *aspect.CombinedConfig, attrs attribute.Bag, mappe
 			return
 		}
 	}()
+
+	var asp aspect.Wrapper
+	if asp, err = m.cacheGet(cfg, mgr, adapter); err != nil {
+		return nil, err
+	}
+
 	// TODO act on adapter.Output
 	return asp.Execute(attrs, mapper)
 }
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -82,7 +82,7 @@ func (m *Manager) Registry() *Registry {
 
 // Execute performs the aspect function based on CombinedConfig and attributes and an expression evaluator
 // returns aspect output or error if the operation could not be performed
-func (m *Manager) Execute(cfg *aspect.CombinedConfig, attrs attribute.Bag, mapper expr.Evaluator) (*aspect.Output, error) {
+func (m *Manager) Execute(cfg *aspect.CombinedConfig, attrs attribute.Bag, mapper expr.Evaluator) (out *aspect.Output, err error) {
 	var mgr aspect.Manager
 	var found bool
 
@@ -96,11 +96,17 @@ func (m *Manager) Execute(cfg *aspect.CombinedConfig, attrs attribute.Bag, mappe
 	}
 
 	var asp aspect.Wrapper
-	var err error
 	if asp, err = m.cacheGet(cfg, mgr, adapter); err != nil {
 		return nil, err
 	}
 
+	defer func() {
+		if r := recover(); r != nil {
+			out = nil // invalidate whatever partial result we got; should we set this to a Code.Code_INTERNAL or similar?
+			err = fmt.Errorf("adapter '%s' panicked with '%v'", cfg.Builder.Name, r)
+			return
+		}
+	}()
 	// TODO act on adapter.Output
 	return asp.Execute(attrs, mapper)
 }
--- a/pkg/adapterManager/manager_test.go
+++ b/pkg/adapterManager/manager_test.go
@@ -118,12 +118,12 @@ func TestRecovery_NewAspect(t *testing.T) {
 		&istioconfig.Adapter{Name: name},
 	}
 
-	if _, err := m.Execute(cfg, nil, nil); err != nil {
-		if !strings.Contains(err.Error(), "NewAspect") {
-			t.Errorf("Expected err from panic with message containing 'NewAspect', actual: %v", err)
-		}
-	} else {
+	_, err := m.Execute(cfg, nil, nil)
+	if err == nil {
 		t.Error("Aspect threw, but got no err from manager.Execute")
+	}
+	if !strings.Contains(err.Error(), "NewAspect") {
+		t.Errorf("Expected err from panic with message containing 'NewAspect', actual: %v", err)
 	}
 }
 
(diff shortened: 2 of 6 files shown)


ZackButcher:
I verified this by both having the manager panic and adding a deadline flag to mixc. In both cases we got the expected behavior. If we're interested in a flag for setting the request deadline in mixc I can include it in this PR too.

pkg/api/methodHandlers.go
```diff
@@ -131,6 +138,38 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
...
+				} else {
+					c <- result{nil, fmt.Errorf("adapter '%s' panicked", conf.Adapter.Name)}
+				}
+			}
+		}()
+		o, err := h.mngr.Execute(conf, ab, h.eval)
+		c <- result{o, err}
+	}()
+
+	select {
```
ZackButcher: Actually, giving this more thought: if we keep the timeout check above in `execute` we can remove the check here and not spin up a new goroutine for the adapter, keeping it all in the request thread. I *think* that'd be better, thoughts?

douglas-reid:
Missing tests?

pkg/api/methodHandlers.go
```diff
@@ -131,6 +138,38 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
...
+	type result struct {
+		out *aspectsupport.Output
+		err error
+	}
+	c := make(chan result)
+	// TODO: should we use some threadpool impl instead of just spawning a thread?
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				glog.Warningf("Recovered from panic in adapter '%s'", conf.Adapter.Name)
```
geeknoid: Should log this as an error.

Should disable/remove the aspect once it has failed so we don't end up continuously failing. Maybe we should keep a counter and allow N failures per minute or something....

ZackButcher:
There's no tests for the entire file, I can start some.

ZackButcher:
Updated to keep things in the request thread; the semantics are essentially unchanged but we don't have to run the adapter in a different thread. Do we like this version more or less than the other?

geeknoid:
We ultimately need to imbue the code with parallelism. At some point, we
need a goroutine pool that will take requests from the front-end and issue
the calls into the adapters. The adapters will block on I/O, but we can
keep pumping incoming requests (until some predetermined limit at which
point we need to start failing additional requests)

I think the old mixologist code already has this kind of model.

Given this, seems like it might be time to start embracing go routines and
not shy away...

douglas-reid:
Yeah, the prototype used go routines. We learned some lessons in this regard. @mandarjog can give you a full dump of the details (debugging, performance troubleshooting, etc.).

ZackButcher:
The version previously written had no parallelism of adapter execution, the only parallelism was to enable monitoring both the context's cancellation and the execution of the adapter. The current version is a little worse in that we wait for the adapter to finish executing before returning a `DEADLINE_EXCEEDED` to the user, where previously we returned it immediately, but there's no extra expenditure of resources since we can't kill the goroutine executing the adapter anyway.

It's as easy to implement parallel execution of adapters with the code as currently written compared to how it was previously written (changing `h.execWrapper(ctx, conf, ab)` to `go h.execWrapper(ctx, conf, ab)`) but I think the current version is easier to understand.

pkg/api/methodHandlers.go
```diff
@@ -131,6 +136,24 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
...
+// It also monitors the RPC's deadline for timeouts.
+func (h *methodHandlers) execWrapper(ctx context.Context, conf *aspectsupport.CombinedConfig, ab attribute.Bag) (out *aspectsupport.Output, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			glog.Errorf("Recovered from panic in adapter '%s'", conf.Adapter.Name)
+			out = nil
+			if panicErr, ok := r.(error); ok {
+				err = panicErr
+			} else {
+				err = fmt.Errorf("adapter '%s' panicked", conf.Adapter.Name)
```
mandarjog: we still want recovers return value captured in the error.

```go
fmt.Errorf("adapter '%s' panicked with '%v'", conf.Adapter.Name, r)
```

pkg/api/methodHandlers.go
```diff
@@ -116,9 +116,14 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
 	// get a new context with the attribute bag attached
 	ctx = attribute.NewContext(ctx, ab)
 	for _, conf := range h.configs[method] {
-		// TODO: plumb ctx through uber.manager.Execute
-		_ = ctx
-		out, err := h.mngr.Execute(conf, ab, h.eval)
+		select {
+		case <-ctx.Done():
+			return newStatusWithMessage(code.Code_DEADLINE_EXCEEDED, ctx.Err().Error())
```
mandarjog: Would be nice to send back results available so far...

ZackButcher: I'm not sure we want to do that. For example, what if my auth adapter takes too long but would've denied the request, and all of the other adapters I executed gave an OK result? I don't think we want to return an OK in that case, right?

There's also another problem with that's written: that it's possible for Done to be closed for a reason other than deadline exceeded (e.g. callers cancels the request).

ZackButcher:
Thought about this over the weekend: I think this is better if we push catching adapter panics down into the manager, so that manager.Execute always returns either output or an err. I think this will make things cleaner when we move to concurrent adapter execution.

I also updated the panic-handling code to always return the recover, even if its not an error, by wrapping it in another error providing context on the failure.

Still open is how to handle the context finishing before we're done executing adapters. As written, the loop in methodHandlers exit as soon as an adapter returns a not-OK code. Mandar suggested returning the output we have to far, but I'm not sure we want to return an OK if the request is canceled or we deadline exceed. Added a TODO to look into this behavior later.

pkg/adapterManager/manager.go
```diff
@@ -96,11 +96,17 @@ func (m *Manager) Execute(cfg *aspect.CombinedConfig, attrs attribute.Bag, mappe
 	}
 
 	var asp aspect.Wrapper
-	var err error
 	if asp, err = m.cacheGet(cfg, mgr, adapter); err != nil {
 		return nil, err
 	}
 
+	defer func() {
```
mandarjog: We should wrap *all* AdapterImpl code in a try-catch.

asp.Execute and  m.CacheGet both call AdapterImpl supplied code. 
pl. move defer before the cacheGet call.

pkg/adapterManager/manager_test.go
```diff
@@ -38,12 +43,51 @@ type (
...
+	testAspect struct {
+		throw bool
+	}
 )
 
 func (m *fakemgr) Kind() string {
 	return m.kind
 }
 
+func newTestManager(name string, throwOnNewAspect bool, aspectThrowOnExecute bool) testManager {
```
ZackButcher: Do we have a manager/aspect for testing already? If not, we should create one (I'll need to another to use in a different package as part of testing the pkg/api surface in a PR coming soon).

pkg/adapterManager/manager_test.go
```diff
@@ -60,3 +104,47 @@ func TestManager(t *testing.T) {
...
+	if err := m.Registry().RegisterDenyChecker(cacheThrow); err != nil {
+		t.Errorf("Failed to register deny checker in test setup with err: %v", err)
+	}
+
+	cfg := &aspect.CombinedConfig{
+		&istioconfig.Aspect{Kind: name},
+		&istioconfig.Adapter{Name: name},
+	}
+
+	if _, err := m.Execute(cfg, nil, nil); err != nil {
```
douglas-reid: might consider re-ordering this:

```
_, err := m.Execute...
if err == nil {
   t.Error("Execute(): expected error")
}
if !strings... {
    t.Errorf("Execute() => got '%v', want message containing 'NewAspect'", err)
}
```

That would eliminate the nested ifs.

Same suggestion for `TestRecovery_AspectExecute`.

pkg/api/methodHandlers_test.go
```diff
@@ -0,0 +1,57 @@
...
+	// we're skipping NewMethodHandlers so we don't have to deal with config since configuration should've matter when we have a canceled ctx
+	handler := &methodHandlers{
+		configs: map[Method][]*aspect.CombinedConfig{Check: {&aspect.CombinedConfig{}}},
+	}
+
+	cancel()
+
+	s := handler.execute(ctx, attribute.NewManager().NewTracker(), &mixerpb.Attributes{}, Check)
+	if s.Code != int32(code.Code_DEADLINE_EXCEEDED) {
+		t.Errorf("Didn't return deadline exceeded when handler.execute was called with a canceled context, got status: %v", s)
```
douglas-reid: super nit: reword error message to something like:

`"execute(canceledContext) => %v, wanted status with code %v", s,  code.Code_DEADLINE_EXCEEDED"`

PR #178, opened 2017-01-20
Refactor NewAspect() methods to match aspect names.

douglas-reid:
This will allow a single adapter to provide multiple aspects, as it avoids name collision that would result in `method redeclared` errors. This should make the builder -> aspect impl relationship clearer, as well.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/178)


douglas-reid:
@mandarjog I was just about to send a PR for `stdioLogger` impl'ing `logger` and `accessLogger`. Would that be sufficient?

douglas-reid:
File #179 to track testing/verification.

PR #180, opened 2017-01-21
Refactor

geeknoid:
pkg/aspect -> pkg/adapter
pkg/adaptertesting -> pkg/adapterTesting

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/180)


adapter/ipListChecker/adapter.go
```diff
@@ -40,7 +40,7 @@ func (a *adapterState) Description() string {
 
 func (a *adapterState) Close() error { return nil }
 
-func (a *adapterState) ValidateConfig(cfg aspect.Config) (ce *aspect.ConfigErrors) {
+func (a *adapterState) ValidateConfig(cfg adapter.Config) (ce *adapter.ConfigErrors) {
```
mandarjog: Even though adapter.ConfigErrors work well here, 
they look out of place with aspect managers which also impement `ValidateConfig`
perhaps. `mixer.ConfigErrors` would work best.

adapter/stdioLogger/stdioLogger.go
```diff
@@ -22,33 +22,32 @@ import (
...
 )
 
 // Register adds the stdioLogger adapter to the list of logger.Aspects known to
 // mixer.
-func Register(r registry.Registrar) error { return r.RegisterLogger(&adapter{}) }
+func Register(r registry.Registrar) error { return r.RegisterLogger(&adapterState{}) }
 
-func (a *adapter) Name() string { return "istio/stdioLogger" }
-func (a *adapter) Description() string {
+func (a *adapterState) Name() string { return "istio/stdioLogger" }
```
mandarjog: isn't `state` implied even if you use just `adapter`. It is the receiver of these methods so it will hold configuration and other state.

geeknoid: Can't use "adapter" since it conflicts with the package named "adapter". Thus the rename.

adapter/genericListChecker/adapter.go
```diff
@@ -29,13 +29,13 @@ func Register(r registry.Registrar) error {
...
-func (a *adapterState) DefaultConfig() aspect.Config                             { return &config.Params{} }
+func newAdapter() listChecker.Adapter                                              { return &adapterState{} }
+func (a *adapterState) Name() string                                               { return "istio/genericListChecker" }
+func (a *adapterState) Description() string                                        { return "Checks whether a string is present in a list." }
+func (a *adapterState) Close() error                                               { return nil }
+func (a *adapterState) ValidateConfig(c adapter.Config) (ce *adapter.ConfigErrors) { return }
+func (a *adapterState) DefaultConfig() adapter.Config                              { return &config.Params{} }
 
-func (a *adapterState) NewListChecker(env aspect.Env, c aspect.Config) (listChecker.Aspect, error) {
+func (a *adapterState) NewListChecker(env adapter.Env, c adapter.Config) (listChecker.Aspect, error) {
```
geeknoid: Done

adapter/ipListChecker/adapter.go
geeknoid: Let's defer this one to a subsequent PR.

geeknoid:
PTAL

geeknoid:
New issue created.

douglas-reid:
Are we still planning to rename *Adapter to *Builder or *Factory?  Otherwise, we have the stutter of `adapter.DenyCheckerAdapter`.  I'd like to move towards something like: `adapter.DenyCheckerFactory` and `adapter.DenyChecker` (then `NewDenyChecker()` could return `adapter.DenyChecker`)

Also, any reason not to make `pkg/adapterTesting` to `pkg/adapter/testing`, as implementors will want to be exposed to it for their own testing purposes?

PR #184, opened 2017-01-22
Config Struct - dispatch prep

mandarjog:
Wireup new config structs aspectmanagers down
1. Use config.Combined struct
2. Remove proto.struct from everywhere, it is not longer needed
3. Up test coverage for files touched.

Removed  repeated validation code from managers.
They are expected to get a validated config.
It is ok to panic since we will ensure that the dispatcher will catch all the panics.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/184)

### DIFF
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -61,7 +61,7 @@ func cacheKey(cfg *config.Combined) (*CacheKey, error) {
 		Impl: cfg.Adapter.GetImpl(),
 	}
 
-	//TODO shas and store with params
+	//TODO compute shas and store with params
 	var b bytes.Buffer
 	// use gob encoding so that we don't rely on proto marshal
 	enc := gob.NewEncoder(&b)
@@ -72,7 +72,6 @@ func cacheKey(cfg *config.Combined) (*CacheKey, error) {
 		}
 		ret.AdapterParamsSha = sha1.Sum(b.Bytes())
 	}
-
 	b.Reset()
 	if cfg.Adapter.GetParams() != nil {
 		if err := enc.Encode(cfg.Aspect.GetParams()); err != nil {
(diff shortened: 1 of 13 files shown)


pkg/adapterManager/manager.go
```diff
@@ -44,27 +49,43 @@ type Manager struct {
 // CacheKey is used to cache fully constructed aspects
 // These parameters are used in constructing an aspect
 type CacheKey struct {
-	Kind   string
-	Impl   string
-	Params string
-	Args   string
+	Kind             string
+	Impl             string
+	AdapterParamsSha [sha1.Size]byte
```
douglas-reid: nit: i feel like these should be named *SHA1.

pkg/adapterManager/manager_test.go
```diff
@@ -18,21 +18,29 @@ import (
 	"strings"
 	"testing"
 
-	istioconfig "istio.io/api/mixer/v1/config"
+	"istio.io/mixer/pkg/config"
+	configpb "istio.io/mixer/pkg/config/proto"
 
+	"fmt"
```
douglas-reid: looks like the imports could use some cleanup.

pkg/aspect/loggerManager_test.go
```diff
@@ -22,14 +22,51 @@ import (
...
-	"github.com/golang/protobuf/ptypes/struct"
 	"istio.io/mixer/pkg/adapter"
-	"istio.io/mixer/pkg/aspect/config"
+	. "istio.io/mixer/pkg/aspect/config"
 	"istio.io/mixer/pkg/attribute"
 	"istio.io/mixer/pkg/expr"
 
-	configpb "istio.io/api/mixer/v1/config"
 	dpb "istio.io/api/mixer/v1/config/descriptor"
+	"istio.io/mixer/pkg/config"
```
douglas-reid: let's move this up to the other block and move the '.' down to this block (with a comment on why the '.' import).

mandarjog:
abandoning this in favor of PR #189

Issue #162, opened 2017-01-13
Build docker images in CI


kyessenov:
Please build server docker images in your CI/CD workflow to be utilized by Istio manager.
Common pattern is to tag your images as:
    gcr.io/istio-testing/mixer-alpha:GIT_SHA

douglas-reid:
docker images are now being generated on successful merges by travis and pushed to GCR (istio-test GCP project). there is likely more work to do to keep the image building current with config changes, etc., but I'm marking this issue as closed for now.

PR #186, opened 2017-01-23
Introduce docker build bits for mixer.

douglas-reid:
At the moment, building a docker image for mixer appears to only work on
linux machines. As such, these builds should only be conducted on linux
hosts.

This setup should closely align with the manager build images and push
images to the same istio-test repo.

This attempts to begin to address #162.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/186)


docker/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -ex
+
+gcloud config set project istio-test
```
mandarjog: we should rename the file  gcloud_build.sh.
It is important to have a script that just works in our environment.

douglas-reid: OK. This bit just duplicates the docker/build.sh from istio/manager. I'll rename to gcloud_build.sh. Ultimately, we just need a place to put our images.

douglas-reid:
PTAL.

Moved: `docker/build.sh` to `docker/gcloud_build.sh`

BUILD.debian
```diff
@@ -0,0 +1,17 @@
...
+# Extract .xz files
+genrule(
+    name = "jessie_tar",
+    srcs = ["docker-brew-debian-b39d31635ca26c8b1f3d982090ba8d54167c4d85/jessie/rootfs.tar.xz"],
+    outs = ["jessie_tar.tar"],
+    cmd = "cat $< | xzcat >$@",
+)
+
+docker_build(
+    name = "jessie",
```
douglas-reid: does envoy require ubuntu (and trusty in particular)? seems like we might want xenial (if we stick with ubuntu).

just looking at the image sizes via pantheon, the envoy and manager images seem to be significantly larger than the mixer image (~60-70 vs 50 MBs). I was already considering moving mixer to jessie-slim to decrease the image size. not sure how important that will become.

kyessenov: I tried making Envoy with public debian xenial image but there was mismatch with GLIBC dependencies, and I did not proceed. For better or worse, it's ubuntu trusty for now. The thing with docker is that if we all share base layer, docker will only download it once.

douglas-reid:
PTAL. Updated to xenial as the base image.

PR #187, opened 2017-01-23
Rationalize the implementation of the aspect managers.

geeknoid:
They look the same, behave the same, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/187)

### DIFF
--- a/pkg/aspect/listCheckerManager.go
+++ b/pkg/aspect/listCheckerManager.go
@@ -29,10 +29,10 @@ type (
 	listCheckerManager struct{}
 
 	listCheckerWrapper struct {
-		cfg          *CombinedConfig
-		adapter      adapter.ListCheckerBuilder
-		aspect       adapter.ListCheckerAspect
-		aspectConfig *config.ListCheckerParams
+		aspect         adapter.ListCheckerAspect
+		blacklist      bool
+		inputs         map[string]string
+		checkAttribute string
 	}
 )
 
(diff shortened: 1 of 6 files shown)


pkg/aspect/manager.go
```diff
@@ -61,3 +66,12 @@ type (
 		Execute(attrs attribute.Bag, mapper expr.Evaluator) (*Output, error)
 	}
 )
+
+func structToProto(in *structpb.Struct, out proto.Message) error {
```
ZackButcher: This func feels a bit out of place here (especially since jsonpb, structpb, etc weren't included previously), but shoving it off to a util.go file in this package doesn't feel great either.

geeknoid:
I think it's temporary, we'll have a better solution for this RSN.

pkg/aspect/listCheckerManager.go
```diff
@@ -42,28 +42,37 @@ func NewListCheckerManager() Manager {
 }
 
 // NewAspect creates a listChecker aspect.
-func (listCheckerManager) NewAspect(cfg *CombinedConfig, ga adapter.Builder, env adapter.Env) (Wrapper, error) {
-	aa, ok := ga.(adapter.ListCheckerBuilder)
-	if !ok {
-		return nil, fmt.Errorf("adapter of incorrect type; expected adapter.ListCheckerBuilder got %#v %T", ga, ga)
+func (m listCheckerManager) NewAspect(c *CombinedConfig, b adapter.Builder, env adapter.Env) (Wrapper, error) {
+	aspectCfg := m.DefaultConfig().(*config.ListCheckerParams)
```
douglas-reid: aside: it sure feels like these structToProto() blocks should be in some way abstracted. other than the little bit of typing, these are just repeated from manager to manager.

mandarjog:
Please do not merge this before merging the config PR.
Config PR changes the aspect managers, registries and a number of other things.

All `structpb` has been removed from the code, it is no longer needed.

PR #188, opened 2017-01-24
Implement parallel execution of adapters

ZackButcher:
This PR introduces a `pool` struct which represents a pool of worker goroutines. This pool is used by methodHandlers to execute adapters in parallel. A pool is chosen rather than naively creating a goroutine for each adapter so that the maximum resource usage of the mixer server can be bound (as a function of number of workers in the pool).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/188)


pkg/api/methodHandlers.go
```diff
@@ -119,25 +123,39 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
 
 	// get a new context with the attribute bag attached
 	ctx = attribute.NewContext(ctx, ab)
-	var out *aspect.Output
+
+	numAdapters := len(h.configs[method])
+
+	// Loop over our configs, executing the adapter for each on another thread. Because enqueue blocks until a worker
```
geeknoid: thread -> go routine.

They're very different you know... :-)

pkg/api/methodHandlers.go
```diff
@@ -159,6 +177,11 @@ func (h *methodHandlers) Quota(ctx context.Context, tracker attribute.Tracker, r
 	response.Result = newQuotaError(code.Code(status.Code))
 }
 
+// Shutdown is used to gracefully terminate the methodHandlers.
+func (h *methodHandlers) Shutdown() {
```
geeknoid: Need test for shutdown...

pkg/api/pool.go
```diff
@@ -0,0 +1,125 @@
...
+// pool refers to a group of goroutinues that execute tasks and return results
+type pool struct {
+	work chan<- task     // Used to hand work items to the workers in the pool
+	quit chan<- struct{} // Used to kill the workers in the queue
+	wg   *sync.WaitGroup // Used to block shutdown until all workers complete
+}
+
+// newPool returns a new pool of workers; it should be called once at server startup, and requestGroup should be called
+// per request to get a handle to enqueue work on the pool.
+func newPool(size uint) *pool {
```
geeknoid: Make this an int and remove the check & panic below, force the range checking to move up into the cmd/server code instead. This will increase coverage in this code.

pkg/api/pool.go
```diff
@@ -0,0 +1,125 @@
...
+	ab     attribute.Bag
+	mapper expr.Evaluator
+	done   chan<- result
+}
+
+// pool refers to a group of goroutinues that execute tasks and return results
+type pool struct {
+	work chan<- task     // Used to hand work items to the workers in the pool
+	quit chan<- struct{} // Used to kill the workers in the queue
+	wg   *sync.WaitGroup // Used to block shutdown until all workers complete
```
geeknoid: Seems like this doesn't need to be a pointer.

pkg/api/pool.go
```diff
@@ -0,0 +1,125 @@
...
+		wg:   wg,
+	}
+}
+
+// requestGroup is called on the pool to get a handle to enqueue all of the adapter executions for a single request into the pool.
+// The returned channel is the channel the adapters write their results to, and the enqueueFunc is used to enqueue the execution
+// of a single adapter. The enqueueFunc will bock until there is a worker available to perform the work being enqueued.
+//
+// numAdapters is used to size the buffer of the result chan; it's important that the pool's workers not block returning results
+// to avoid a deadlock amongst request threads attempting to enqueue work onto the pool, so we need a buffer large enough for all
```
geeknoid: threads -> go routines

pkg/api/methodHandlers.go
ZackButcher: The test `TestQuit` tested it, renamed to `TestShutdown`

pkg/api/pool.go
ZackButcher: [`Waitgroup`s can't be copied after use](https://golang.org/pkg/sync/#WaitGroup), but we're handing it off to other goroutines, so it must be a pointer, right? Or, at least it must be a pointer when it's handed to the goroutine, the pool itself can hold the instance directly since the pool is only passed around as a pointer.

pkg/api/pool_test.go
```diff
@@ -0,0 +1,192 @@
...
+		enqueue(nil, cfg)
+	}
+
+	defer func() {
+		if r := recover(); r == nil {
+			t.Error("enqueue(nil, nil) got nil, want panic and non-nil recover.")
+		}
+	}()
+	enqueue(nil, cfg)
+	t.Error("enqueue(nil, nil) should panic")
```
douglas-reid: this isn't `enqueue(nil, nil)` though, is it? cfg looks to be assigned a non-nil value. have i missed something?

pkg/api/pool_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}()
+
+	go func() {
+		p.shutdown()
+		close(succeed)
+	}()
+
+	select {
+	case <-fail:
+		t.Error("pool.shutown() didn't complete in the expected time")
```
douglas-reid: s/shutown/shutdown/

pkg/api/pool_test.go
```diff
@@ -0,0 +1,192 @@
...
+		time.Sleep(10 * time.Millisecond)
+	}
+	if len(res) != 2 {
+		t.Errorf("After unblocking queue expected two finished tasks, got %d", len(res))
+	}
+
+	for i := 0; i < 2; i++ {
+		r := <-res
+		if r.out.Code != code.Code_OK {
+			t.Error("Got back bad code.")
```
douglas-reid: please add what the bad code was to the error message (ex: `got %v, want OK`)

pkg/api/pool_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	close(blockChan) // unblock the queue
+	<-second         // block the test thread till the second enqueue completes
+
+	// It takes a little time for the two goroutines to write their results; we loop to make the test more reliable.
+	for count := 0; len(res) != 2 && count < 5; count++ {
+		time.Sleep(10 * time.Millisecond)
+	}
+	if len(res) != 2 {
+		t.Errorf("After unblocking queue expected two finished tasks, got %d", len(res))
```
douglas-reid: nit: reorder message: `got %d finished tasks, want 2`.

pkg/api/methodHandlers.go
ZackButcher: Whoops, wrong shutdown function; added one for MethodHandlers.Shutdown.

pkg/api/pool_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if len(res) != 0 {
+		t.Errorf("Expected result chan to be empty before work completes, got length: %d", len(res))
+	}
+
+	close(blockChan) // unblock the queue
+	<-second         // block the test thread till the second enqueue completes
+
+	// It takes a little time for the two goroutines to write their results; we loop to make the test more reliable.
+	for count := 0; len(res) != 2 && count < 5; count++ {
```
douglas-reid: can we do this without the sleeps?  some sort of loop on the res channel that pulls data into a slice with a deadline or something?

pkg/api/pool.go
```diff
@@ -0,0 +1,125 @@
...
+// adapters in a request to return their values. To enforce this invariant, enqueueFunc will panic if called more than numAdpaters times.
```
douglas-reid: is panic() the best way to handle this? could we have an error channel and close the result channel, or something else similar?

ZackButcher: I chose a panic because this constraint being violated indicates a logical error in the methodHandlers code and the pool struct is private to the API package so it won't be used elsewhere. We can't close the result channel since that means the go routines that were already started would panic when returning results. We could change enqueueFunc from

    type enqueueFunc func(ctx context.Context, config *aspect.CombinedConfig)

to

    type enqueueFunc func(ctx context.Context, config *aspect.CombinedConfig) error

and return an error when enqueue is called too many times. However, I think a panic is appropriate in this case for the same reason an unlocked mutex panicking when unlocked again is appropriate: it's an indication of some other error occurring.

ZackButcher:
Addressed comments, PTAL.

pkg/api/methodHandlers.go
```diff
@@ -119,25 +121,39 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
...
 	// get a new context with the attribute bag attached
 	ctx = attribute.NewContext(ctx, ab)
-	var out *aspect.Output
+
+	numAdapters := len(h.configs[method])
+
+	// Loop over our configs, executing the adapter for each on another go routine. Because enqueue blocks until a worker
+	// is available it could block for a long time, so we check to see if our context was canceled each time we attempt
+	// to enqueue work.
+	results, enqueue := h.pool.requestGroup(h.mngr, ab, h.eval, numAdapters)
```
mandarjog: The new handler takes an `Executor` interface and delegates execution to it.  (it is implemented by adapter manager). What if we change it to a BatchExecutor interface. which take []config.Combined

ExecuteBatch(ctx, []config.Combined, ...)

behind that batch can be a pool for parallel execution, or a non parallel model.
It would mean that pool code will live outside api.   pkg/pool, It will be constructed by  server.go and simply used by api.Handler

ZackButcher: SGTM, I'll make that change.

ZackButcher:
Just reworked this significantly in light of the config handling changes from @mandarjog, PTAL.

pkg/adapterManager/parallelManager.go
```diff
@@ -0,0 +1,153 @@
...
+
+// BulkExecute takes a set of configurations and Executes all of them in parallel.
+func (p *ParallelManager) BulkExecute(ctx context.Context, cfgs []*config.Combined, attrs attribute.Bag) ([]*aspect.Output, error) {
+	numCfgs := len(cfgs)
+	// TODO: since we don't currently implement using different adapter sets per request,
+	// len(cfgs) is constant of the life of the configuration for all requests. We should probably use a
+	// pool for both the array of aspect outputs here and the result channels from requestGroup.
+	results := make([]*aspect.Output, numCfgs)
+
+	r, enqueue := p.requestGroup(attrs, numCfgs)
```
mandarjog: could the requestGroup just take `[]cfg` as an argument?
since euque is an async operation, the main cancel point here is on line `81`.

mandarjog: @ZackButcher this should not hold the merging.

pkg/adapterManager/parallelManager.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2016 Google Inc.
```
geeknoid: 2017

pkg/adapterManager/parallelManager.go
```diff
@@ -0,0 +1,153 @@
...
+// request go routines, i.e. it's global.
+type ParallelManager struct {
+	*Manager
+
+	work chan<- task     // Used to hand tasks to the workers in the pool
+	quit chan<- struct{} // Used to shutdown the workers in the pool
+	wg   *sync.WaitGroup // Used to block shutdown until all workers complete
+}
+
+// NewParallelManager a new Manager who's BulkExecute method calls manager.Execute on each config in parallel.
```
geeknoid: rephrase.

pkg/adapterManager/parallelManager.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, cfg := range cfgs {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("failed to enqueue all config executions with err: %v", ctx.Err())
+		default:
+			enqueue(ctx, cfg)
+		}
+	}
+
+	for i := 0; i < numCfgs; i++ {
```
geeknoid: for i := range numCfgs {

pkg/api/handler.go
```diff
@@ -56,7 +56,10 @@ type Handler interface {
 // Executor executes any aspect as described by config.Combined.
 type Executor interface {
 	// Execute performs actions described in combined config using an attribute bag.
-	Execute(cfg *config.Combined, attrs attribute.Bag) (*aspect.Output, error)
+	Execute(ctx context.Context, cfg *config.Combined, attrs attribute.Bag) (*aspect.Output, error)
```
geeknoid: Why do we need both Execute and BulkExecute? Given the usage, I'd rather have a single method called Execute that takes a [].

pkg/adapterManager/parallelManager.go
ZackButcher: Unfortunately this does not work in go, range only works over arrays, slices, strings, maps, and channels.

cmd/server/server.go
```diff
@@ -133,7 +138,17 @@ func runServer(sa *serverArgs) error {
...
 	configManager := config.NewManager(eval, adapterMgr.AspectValidatorFinder(), adapterMgr.BuilderValidatorFinder(),
 		sa.globalConfigFile, sa.serviceConfigFile, time.Second*time.Duration(sa.configFetchIntervalSec))
-	handler := api.NewHandler(adapterMgr, adapterMgr.AspectMap())
+
+	var handler api.Handler
+	if sa.workerPoolSize == 0 {
+		handler = api.NewHandler(adapterMgr, adapterMgr.AspectMap())
+	} else {
+		poolsize := int(sa.workerPoolSize)
+		if poolsize < 0 {
```
douglas-reid: is this possible?  isnt' workerPoolSize a `uint`?

pkg/adapterManager/manager_test.go
```diff
@@ -215,12 +214,66 @@ func TestManager(t *testing.T) {
...
+		r := getReg(true)
+		mgr := newFakeMgrReg(&fakewrapper{})
+		m := newManager(r, mgr, mapper, nil)
+
+		errStr := ""
+		if _, err := m.Execute(context.Background(), c.cfgs, attrs); err != nil {
+			errStr = err.Error()
+		}
+		if !strings.Contains(errStr, c.errString) {
+			t.Errorf("[%d] expected: '%s' \ngot: '%s'", idx, c.errString, errStr)
```
douglas-reid: nit: reorder for got, then want.

pkg/adapterManager/parallelManager.go
```diff
@@ -0,0 +1,123 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adapterManager
+
+import (
+	"fmt"
+	"sync"
+
```
douglas-reid: remove this line, and re-sort.

cmd/server/server.go
ZackButcher: We cast the uint to an int, so it's possible that the interpretation as bits is a negative number. i.e. uint64 max = 18446744073709551615, int64 max = 9223372036854775807

douglas-reid: i think the cast will fail with a message like: "...overflows uint" instead of overflowing.

see: https://play.golang.org/p/XoNhhkP5J2.

douglas-reid: nevermind. i was thinking of consts... not applicable here.

douglas-reid: just to show i'm not crazy: https://play.golang.org/p/NoyjdX8pSt

ZackButcher: It's cool that consts do that at least. I had no idea those kinds of checks were performed at compile time.

ZackButcher: Rather, I should say I didn't know go failed compilation when invalid constant casts are performed.

PR #189, opened 2017-01-24
Add config resolve and config command line params

mandarjog:
This PR wires up config struct and uses config resolve to select and dispatch `config.Combined` to aspectManagers.

1. Remove proto.struct from everywhere, it is not longer needed
2. Up test coverage for files touched.
3. Removed repeated validation code from managers.
    The aspect managers are expected to get a validated config.
    It is ok to panic since we will ensure that the dispatcher will catch all the panics.
4. Add adapter.Inventory that maintains inventory of all adapters (builders) available in the system
5. Add aspect.Inventory that maintains inventory of all aspects (aspect Managers) in the system
6. Use (serviceConfig and globalConfig) as  command line parameters for config input.

This part will be replaced by a single URL that points to a repository location.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/189)


pkg/config/api.go
```diff
@@ -0,0 +1,30 @@
...
+	// CheckMethod represents Check operation of the mixer api.
+	CheckMethod APIMethod = iota
+	// ReportMethod represents Report operation of the mixer api.
+	ReportMethod
+	// QuotaMethod represents Quota operation of the mixer api.
+	QuotaMethod
+)
+
+// APIMethods is the authoritative place that lists supported api methods.
+var APIMethods = []APIMethod{CheckMethod, ReportMethod, QuotaMethod}
```
geeknoid: Somewhat ikky to expose a mutable global variable outside of the package. Probably should be made internal and have a function expose the array.

pkg/config/manager_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		if mt.errStr == "" && le == nil {
+			continue
+		}
+
+		if !strings.Contains(le.Error(), mt.errStr) {
+			t.Errorf("[%d] Unexpected error. Expected %s\nGot: %s\n", idx, mt.errStr, le)
+		}
+
+		time.Sleep(time.Second * 1)
```
geeknoid: Why is this here? Can we have real synchronization? If not, then at least a comment explaining why this hack is here.

pkg/config/manager.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+// ChangeListener listens for config change notifications.
+type ChangeListener interface {
+	ConfigChange(cfg *Runtime)
+}
+
+// Manager represents the config Manager.
+// It is responsible for fetching and receiving configuration changes.
+// It applied validated changes to the registered config change listeners.
```
geeknoid: applied -> applies

pkg/config/manager.go
```diff
@@ -0,0 +1,174 @@
...
+	scSha   [sha1.Size]byte
+	gcSha   [sha1.Size]byte
+
+	lock      *sync.RWMutex
+	lastError error
+}
+
+// ManagerArgs are constructor args for a manager
+type ManagerArgs struct {
+	Eval          expr.Evaluator
```
geeknoid: All these fields deserve comments.

pkg/config/manager.go
```diff
@@ -0,0 +1,174 @@
...
+	AspectF       ValidatorFinder
```
geeknoid: Please expand "F" to "Finder"

pkg/config/manager.go
```diff
@@ -0,0 +1,174 @@
...
+// api.Handler listens for config changes.
+type Manager struct {
+	ManagerArgs
+
+	cl      []ChangeListener
+	closing chan bool
+	scSha   [sha1.Size]byte
+	gcSha   [sha1.Size]byte
+
+	lock      *sync.RWMutex
```
geeknoid: Why is this a pointer to a lock, instead of just an instance of a lock inline?

pkg/adapter/registrar.go
```diff
@@ -17,18 +17,18 @@ package adapter
...
-	RegisterQuota(QuotaBuilder) error
+	RegisterQuota(QuotaBuilder)
 }
 
-// RegisterFn is a function the mixer invokes to trigger adapters to registers
-// their aspect builders.
-type RegisterFn func(Registrar) error
+// MustRegisterFn is a function the mixer invokes to trigger adapters to register
+// their aspect builders. It must succeed or panic().
+type MustRegisterFn func(Registrar)
```
mandarjog: I have re-named it back to `Register`. There are prior examples of `MustRegister` which means `RegisterOrPanic`

pkg/config/manager.go
```diff
@@ -0,0 +1,174 @@
...
+	lastError error
+}
+
+// ManagerArgs are constructor args for a manager
+type ManagerArgs struct {
```
ZackButcher: Same as before, I think the struct-as-args is an anti-pattern. Where I've seen struct-as-args used the struct represents options that affect the behavior of the thing being constructed, and they're typically optional. In this case all the arguments are required (except loop delay, I suppose), and you even embed the Args struct directly into the manager. You also don't get any encapsulation since all of these fields are public to enable construction outside of the package; since the Args struct is embedded in the Manager I can change the values arbitrarily after construction. Even though 6 args to the NewManager func isn't great, I think it's better than the struct version.

Alternatively, do what [gRPC does, have a private options struct](https://github.com/grpc/grpc-go/blob/master/server.go#L107) (or, in this case, args struct) and provide functions to operate on them. But, since all of these args are required that just forces more boilerplate everywhere.

pkg/config/manager.go
```diff
@@ -0,0 +1,181 @@
...
+	cl      []ChangeListener
+	closing chan bool
+	scSHA   [sha1.Size]byte
+	gcSHA   [sha1.Size]byte
+
+	sync.RWMutex
+	lastError error
+}
+
+// NewManager returns a config.Manager given ManagerArgs.
```
geeknoid: Comment is out of date.

pkg/api/handler.go
```diff
@@ -119,10 +88,24 @@ func (h *methodHandlers) execute(ctx context.Context, tracker attribute.Tracker,
 
 	// get a new context with the attribute bag attached
 	ctx = attribute.NewContext(ctx, ab)
-	for _, conf := range h.configs[method] {
+
+	cfg := h.cfg.Load().(*config.Runtime)
+
+	if cfg == nil {
+		// config has NOT been loaded yet
+		glog.Warningf("configuration is not available")
```
mandarjog: done.

mandarjog:
Addressed review comments.  PTAL.

adapter/inventory.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"istio.io/mixer/adapter/denyChecker"
+	"istio.io/mixer/adapter/genericListChecker"
+	"istio.io/mixer/adapter/ipListChecker"
+	"istio.io/mixer/adapter/stdioLogger"
+	"istio.io/mixer/pkg/adapter"
+)
+
+// Inventory  returns inventory of all available adapters.
```
douglas-reid: nit: extra space between Inventory and returns.

cmd/server/server.go
```diff
@@ -26,15 +26,17 @@ import (
...
+	"istio.io/mixer/pkg/adapterManager"
 	"istio.io/mixer/pkg/api"
+	"istio.io/mixer/pkg/aspect"
 	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/config"
 	"istio.io/mixer/pkg/tracing"
 
-	denyadapter "istio.io/mixer/adapter/denyChecker"
-	"istio.io/mixer/pkg/aspect"
+	"time"
```
douglas-reid: this looks like it needs cleanup... i would expect std packages to be grouped, etc.

pkg/config/manager_test.go
```diff
@@ -0,0 +1,141 @@
+package config
```
geeknoid: Need copyright.

pkg/aspect/accessLoggerManager_test.go
```diff
@@ -19,14 +19,16 @@ import (
 	"testing"
 	"text/template"
 
-	"github.com/golang/protobuf/ptypes/struct"
 	"istio.io/mixer/pkg/adapter"
-	"istio.io/mixer/pkg/aspect/config"
+	aconfig "istio.io/mixer/pkg/aspect/config"
```
douglas-reid: looks like we could clean this up a bit.  awkward newline between groups of external imports... if we want to keep that, i'd suggest moving the named imports to their own block and otherwise combining.

pkg/aspect/loggerManager_test.go
```diff
@@ -19,17 +19,38 @@ import (
...
 	"istio.io/mixer/pkg/adapter"
-	"istio.io/mixer/pkg/aspect/config"
 	"istio.io/mixer/pkg/aspect/test"
+
+	aconfig "istio.io/mixer/pkg/aspect/config"
 	"istio.io/mixer/pkg/attribute"
 	"istio.io/mixer/pkg/expr"
 
-	configpb "istio.io/api/mixer/v1/config"
+	"errors"
```
douglas-reid: as before, this should be cleaned up.

pkg/aspect/loggerManager_test.go
```diff
@@ -19,17 +19,38 @@ import (
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/ptypes/empty"
 	dpb "istio.io/api/mixer/v1/config/descriptor"
+	"istio.io/mixer/pkg/config"
+	configpb "istio.io/mixer/pkg/config/proto"
+)
+
+type (
```
douglas-reid: this stuff was killed in a recent CL. Let's not bring it back. In particular, testLogger has been moved to test.Logger.

pkg/aspect/loggerManager_test.go
```diff
@@ -254,3 +268,20 @@ func TestLoggerManager_ValidateConfig(t *testing.T) {
 		t.Errorf("ValidateConfig(): unexpected error: %v", err)
 	}
 }
+
+func (t *testLogger) NewLogger(e adapter.Env, m adapter.AspectConfig) (adapter.LoggerAspect, error) {
```
douglas-reid: this stuff is all available in the test package.  not needed here.

pkg/aspect/manager.go
```diff
@@ -20,19 +20,16 @@ package aspect
 import (
 	"google.golang.org/genproto/googleapis/rpc/code"
 
-	istioconfig "istio.io/api/mixer/v1/config"
+	"istio.io/mixer/pkg/config"
+
+	"io"
```
douglas-reid: clean.

pkg/config/manager.go
```diff
@@ -0,0 +1,181 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"crypto/sha1"
+	"io/ioutil"
+	"time"
+
```
douglas-reid: clean.

douglas-reid:
Can we edit the description for the PR to provide more context and a fuller description of the changes?  There is a lot going on here, and it would be nice to document some of the reasoning behind and capabilities provided by this PR.

mandarjog:
I will update the description and update the test package changes.

For package re ordering can that be automated ? 
Do you use a tool?

Regards, 
Sent from my Motorola Next

Mandar U Jog

pkg/config/manager_test.go
mandarjog: done.

mandarjog:
Review changes made.

re: goimports, I have only made changes that can be automated.
I don't think we should hand craft an ordering. 
I have used `goimports -local istio.io  -w` 
The order being "stardard imports, third party imports, istio imports"

I have also added check_license to linters.

pkg/config/manager_test.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+
+	"istio.io/mixer/pkg/adapter"
+
+	"sync"
+	"time"
```
douglas-reid: this still doesn't look correct to me. I know that you've run goimports, but there is an obvious flaw here. "sync" and "time" should be moved up with the other standard import packages and the two "istio.io/mixer/..." blocks should be combined into one.

pkg/adapterManager/manager.go
```diff
@@ -18,122 +18,193 @@ import (
 	"fmt"
 	"sync"
```
douglas-reid: any idea why there is an extra newline here? it feels like the "bytes"... section should be combined (and sorted) with the "fmt" and "sync" import block.

pkg/config/manager_test.go
mandarjog: done.
I need to remove blank lines either manually, or thru gogland "Code cleanup".
gofmt / goimports does not eliminate blank lines.

bin/check_license.sh
```diff
@@ -0,0 +1,24 @@
...
+	if [[ $fn == *.pb.go ]];then
+		continue
+	fi
+	head -20 $fn | grep "Apache License, Version 2" > /dev/null
+	if [[ $? -ne 0 ]]; then
+		echo "${fn} missing license"
+		ret=$(($ret+1))
+	fi
+
+	head -20 $fn | grep Copyright | grep Google > /dev/null
```
geeknoid: Can't require a Google copyright in an OSS project, right?

bin/linters.sh
```diff
@@ -64,4 +64,6 @@ prep_linters
 
 run_linters
 
+$SCRIPTPATH/check_license.sh
```
geeknoid: This should be moved up to be in the run_linters function.

mandarjog:
Done.

PR #191, opened 2017-01-24
Add a check for the disabled tracer in PropagateSpan and extractSpanContext

ZackButcher:
Add a check for the disabled tracer in PropagateSpan and extractSpanContext. PropagateSpan was previously private and only called by other methods in the package which already checked the tracer state, but was made public without adding the check. A check was added to extractSpanContext as well to prevent the same from happening in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/191)

### DIFF
--- a/pkg/tracing/tracing_test.go
+++ b/pkg/tracing/tracing_test.go
@@ -35,7 +35,7 @@ func TestCurrentSpan(t *testing.T) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "first")
 
 	if currentSpan := CurrentSpan(ctx); currentSpan != span {
-		t.Errorf("Failed to extract the current span from the context, expected '%v' actual '%v'; context: '%v'", span, currentSpan, ctx)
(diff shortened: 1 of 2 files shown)


ZackButcher:
This should fix the sigsegv in https://github.com/istio/mixer/issues/190

pkg/tracing/tracing_test.go
```diff
@@ -128,4 +128,13 @@ func TestDisabledTracer(t *testing.T) {
 	if span, _ := tracer.StartSpanFromContext(context.Background(), ""); span != noopSpan {
 		t.Errorf("Expected disabled tracer to return noop span, actual: %+v", span)
 	}
+
+	ctx := context.Background()
+	if _, nctx := tracer.PropagateSpan(ctx, noopSpan); nctx != ctx {
+		t.Errorf("PropagateSpan on disabled tracer modified the context; expected: %v, actual %v", ctx, nctx)
```
douglas-reid: nit: prefer got, want ordering to want, got ordering in error messages:

https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

PR #192, opened 2017-01-25
Implement an opentracing carrier using Istio's attribute system

ZackButcher:
Opentracing defines carriers which are responsible for propagating trace state between servers. These carriers expose the following interface:

    Set(key, value string)
    ForeachKey(handler func(key, val string) error) error

We implement that interface on top of the mixer's `attribute.MutableBag`. This change introduces the carrier struct itself, and modifies the `Bag` interface to expose a way to iterate over the elements of the bag, `ForEach(func(key string, val interface{}) bool)`.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/192)

### DIFF
--- a/pkg/attribute/bag.go
+++ b/pkg/attribute/bag.go
@@ -45,7 +45,7 @@ type Bag interface {
 	// returns false iteration is stopped.
 	//
 	// The view into the values is provided in this form so that the bag can keep iteration thread safe.
-	Foreach(handler func(key string, val interface{}) bool)
+	ForEach(handler func(key string, val interface{}) bool)
 
 	// Done indicates the bag can be reclaimed.
 	Done()
--- a/pkg/attribute/bag.go
+++ b/pkg/attribute/bag.go
@@ -41,6 +41,12 @@ type Bag interface {
 	// Bytes returns the named attribute if it exists.
 	Bytes(name string) ([]uint8, bool)
 
+	// Foreach allows a caller to process all of the (key, value) pairs this bag knows about. If the provided func
+	// returns false iteration is stopped.
+	//
+	// The view into the values is provided in this form so that the bag can keep iteration thread safe.
+	Foreach(handler func(key string, val interface{}) bool)
+
 	// Done indicates the bag can be reclaimed.
 	Done()
 }
--- a/pkg/attribute/bag.go
+++ b/pkg/attribute/bag.go
@@ -26,38 +26,38 @@ type Bag interface {
 	// String returns the named attribute if it exists.
 	String(name string) (string, bool)
 
-	// StringKeys returns a snapshot of all keys corresponding string attributes this bag knows about.
+	// StringKeys returns a snapshot of all keys corresponding to string attributes this bag knows about.
 	StringKeys() []string
 
 	// Int64 returns the named attribute if it exists.
 	Int64(name string) (int64, bool)
 
-	// Int64Keys returns a snapshot of all keys corresponding int64 attributes this bag knows about.
+	// Int64Keys returns a snapshot of all keys corresponding to int64 attributes this bag knows about.
 	Int64Keys() []string
 
 	// Float64 returns the named attribute if it exists.
 	Float64(name string) (float64, bool)
 
-	// Float64Keys returns a snapshot of all keys corresponding float64 attributes this bag knows about.
+	// Float64Keys returns a snapshot of all keys corresponding to float64 attributes this bag knows about.
 	Float64Keys() []string
 
 	// Bool returns the named attribute if it exists.
 	Bool(name string) (bool, bool)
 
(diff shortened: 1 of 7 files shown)


pkg/attribute/bag.go
```diff
@@ -41,6 +41,12 @@ type Bag interface {
 	// Bytes returns the named attribute if it exists.
 	Bytes(name string) ([]uint8, bool)
 
+	// Foreach allows a caller to process all of the (key, value) pairs this bag knows about. If the provided func
+	// returns false iteration is stopped.
+	//
+	// The view into the values is provided in this form so that the bag can keep iteration thread safe.
+	ForEach(handler func(key string, val interface{}) bool)
```
geeknoid: The current approach will lead to deadlock if the handler happens to cause code to call some bag functions. Additionally, given that we're catching panics around adapter code, it's possible that a locked bag will be locked, the handler will cause a panic, we'll catch the panic, and the lock will not be unlocked.

Would be more reliable to implement instead a method that returns a flat [] of the bag's content. It's slower, requires allocation, but it can be made 100% safe.

pkg/attribute/bag_test.go
```diff
@@ -15,6 +15,7 @@
 package attribute
```
geeknoid: The attribute package currently has 100% test coverage. This change would reduce this by not testing a considerable number of early exit paths. Please add tests to all these paths so we stay at 100%.

pkg/attribute/mutableBag.go
```diff
@@ -157,6 +157,53 @@ func (mb *mutableBag) SetBytes(name string, value []uint8) {
...
+		}
+	}
+	for key, val := range mb.bytes {
+		if !handler(key, val) {
+			mb.RUnlock()
+			return
+		}
+	}
+
+	// We're done with this bag, there's no reason to hold the read lock while the iterating over the parent's keys.
```
geeknoid: the iterating -> iterating

pkg/tracing/attributes.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"context"
+	"strings"
+
+	"istio.io/mixer/pkg/attribute"
+)
+
+const (
+	prefix       = "istio-ot-"
```
geeknoid: We'll need to elevate this to some global registry of known attribute names at some point...

pkg/tracing/attributes.go
```diff
@@ -0,0 +1,82 @@
...
+		return true
+	})
+	// If we made it through with no problems, err == nil, otherwise it's some error from handler.
+	return err
+}
+
+// Set adds (key, val) to the set of string attributes in the request scope. It prefixes the key with an istio-private
+// string to avoid collisions with other attributes when arbitrary tracer implementations are used.
+func (c *AttributeCarrier) Set(key, val string) {
+	c.mb.SetString(prefix+key, val)
```
geeknoid: For such an approach to be reliable, there needs to be a character which is not allowed in normal attributes names which is part of your prefix. We could declare, for example, that attribute names can never start with : and then you could have a prefix of :istio-ot- with no risk of ever conflicting with another attribute name.

ZackButcher: Yup, I thought the same; gRPC does this with http headers - any header that begins with a colon is gRPC internal. Alternatively, we could force tracers to be configured like everything else in the system and not dynamically generate tracer propagation attributes (this seems like a PITA for our users though, since the specific data a given tracer impl propagates may not be easy to find).

pkg/attribute/bag.go
ZackButcher: I forgot about the reentry problem. The tracing code will only ever need the keys that correspond to string attributes. I guess the question becomes: do we care about providing a handle to let a user iterate over all keys, do we want to provide a handle to iterate over the keys of each attribute type (StringKeys, Int64Keys, etc) so we get more granular locks and less allocations in the common case, or do we want to provide a way to get only the keys we need right now (the string keys). Thoughts?

ZackButcher:
Addressed comments, PTAL.

pkg/attribute/bag.go
```diff
@@ -26,21 +26,39 @@ type Bag interface {
 	// String returns the named attribute if it exists.
 	String(name string) (string, bool)
 
+	// StringKeys returns a snapshot of all keys corresponding string attributes this bag knows about.
```
geeknoid: Here and in other comments: "corresponding TO"

pkg/attribute/mutableBag.go
```diff
@@ -157,6 +222,19 @@ func (mb *mutableBag) SetBytes(name string, value []uint8) {
 	mb.Unlock()
 }
 
+func (mb *mutableBag) ByteKeys() []string {
```
geeknoid: ByteKeys -> BytesKeys

ZackButcher:
Addressed comments, PTAL.

PR #194, opened 2017-01-25
Add docker image build to travis CI script.

douglas-reid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/194)


douglas-reid:
NOT for review at this time.  This PR exists as a placeholder for testing perms issues, etc., that will impact the ability to push images to a location.

douglas-reid:
PTAL. I believe that I've got the auth bits setup correctly now. With this setup, the docker image should only be updated on merges. For now, it uses a service account on the istio-test project to push to GCR.

This should address the final bit of #162.

docker/travis_gcloud.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -ev
+if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
+	openssl aes-256-cbc -K $encrypted_2f660428f0db_key -iv $encrypted_2f660428f0db_iv -in istio-test.json.enc -out istio-test.json -d
```
mandarjog: are the $encrypted_2f660428f0db_key set in env vars for the build?

douglas-reid: yes. as best I can tell, they were added when I used the travis CLI to encrypt the key file.

PR #196, opened 2017-01-25
Add DefaultBuilder to //pkg/adapter package.

douglas-reid:
This new type will enable adapter impls to embed generic builder behavior into their implementations, saving the need for boilerplate code across the adapter implementations.

As part of the introduction of this new type, the existing adapter implementations have been refactored to use the DefaultBuilder. This has mainly resulted in the removal of the myriad builder.go files in the adapter impl packages. With their removal, the remaining aspect.go files have been renamed to match the package name. To match this rename, the adapter_test.go files have been renamed as well. This should be sufficient for the simple impls we have to date. In the future, adapter impl packages may be split apart into ${aspect}.go or similar scheme.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/196)


douglas-reid:
PTAL. Merged in the config changes (mostly removed error from Register() methods).

PR #202, opened 2017-01-27
Initial quota support.

geeknoid:
- Introduce a simple quota aspect manager. This is largely a placeholder 
until we rejigger things next week to streamline the config story.

- Introduce a simple fully memory-resident quota adapter.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/202)

### DIFF
--- a/adapter/memQuota/config/config.proto
+++ b/adapter/memQuota/config/config.proto
@@ -12,24 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package adapter
+syntax = "proto3";
 
-import (
-	"istio.io/mixer/adapter/denyChecker"
-	"istio.io/mixer/adapter/genericListChecker"
-	"istio.io/mixer/adapter/ipListChecker"
-	"istio.io/mixer/adapter/memQuota"
-	"istio.io/mixer/adapter/stdioLogger"
-	"istio.io/mixer/pkg/adapter"
-)
+package adapter.memQuota.config;
 
-// Inventory returns inventory of all available adapters.
-func Inventory() []adapter.RegisterFn {
-	return []adapter.RegisterFn{
-		denyChecker.Register,
-		genericListChecker.Register,
-		ipListChecker.Register,
-		stdioLogger.Register,
-		memQuota.Register,
-	}
+option go_package="config";
+
+message Params {
+	// Minimum number of seconds that deduplication is possible for a given operation.
+	int32 min_deduplication_window_seconds = 1;
 }
--- a/adapter/memQuota/memQuota.go
+++ b/adapter/memQuota/memQuota.go
@@ -1,0 +1,307 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package memQuota provides a simple in-memory quota implementation. It's
+// trivial to set up, but it has various limitations:
+//
+// - Obviously, the data set must be able to fit in memory.
+//
+// - When the mixer crashes/restarts, all quota values are erased.
+//   This means this isn't good for allocation quotas although
+//   it works well enough for rate limits quotas.
+//
+// - Since the data is all memory-resident and there isn't any cross-node
+//   synchronization, this adapter can't be used in an Istio mixer where
+//   a single service can be handled by different mixer instances.
+package memQuota
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"sync"
+	"time"
+
+	"istio.io/mixer/adapter/memQuota/config"
+	"istio.io/mixer/pkg/adapter"
+)
+
+type builder struct{ adapter.DefaultBuilder }
+
+type memQuota struct {
+	sync.Mutex
+
+	// the definitions we know about, immutable
+	definitions map[string]*adapter.QuotaDefinition
+
+	// the counters we track for non-expiring quotas, protected by lock
+	cells map[string]int64
+
+	// the rolling windows we track for expiring quotas, protected by lock
+	windows map[string]*rollingWindow
+
+	// two ping-ponging maps of active dedup ids
+	recentDedup map[string]int64
+	oldDedup    map[string]int64
+
+	// used for reaping dedup ids
+	ticker *time.Ticker
+
+	// indirection to support fast deterministic tests
+	getTick func() int32
+
+	logger adapter.Logger
+}
+
+// we maintain a pool of these for use by the makeKey function
+type keyWorkspace struct {
+	keys   []string
+	buffer []byte
+}
+
+// pool of reusable keyWorkspace structs
+var keyWorkspacePool = sync.Pool{New: func() interface{} { return &keyWorkspace{} }}
+
+var (
+	name = "istio/memQuota"
+	desc = "Simple volatile memory-based quotas."
+	conf = &config.Params{MinDeduplicationWindowSeconds: 1}
+)
+
+const (
+	// See the rollingWindow comment for a description of what this is for.
+	ticksPerSecond = 10
+)
+
+// Register records the builders exposed by this adapter.
+func Register(r adapter.Registrar) {
+	r.RegisterQuota(newBuilder())
+}
+
+func newBuilder() builder {
+	return builder{adapter.NewDefaultBuilder(name, desc, conf)}
+}
+
+func (builder) NewQuota(env adapter.Env, c adapter.AspectConfig, d map[string]*adapter.QuotaDefinition) (adapter.QuotaAspect, error) {
+	return newAspect(env, c.(*config.Params), d)
+}
+
+// newAspect returns a new aspect.
+func newAspect(env adapter.Env, c *config.Params, definitions map[string]*adapter.QuotaDefinition) (adapter.QuotaAspect, error) {
+	mq := &memQuota{
+		definitions: definitions,
+		cells:       make(map[string]int64),
+		windows:     make(map[string]*rollingWindow),
+		recentDedup: make(map[string]int64),
+		oldDedup:    make(map[string]int64),
+		ticker:      time.NewTicker(time.Duration(c.MinDeduplicationWindowSeconds) * time.Second),
+		getTick:     func() int32 { return int32(time.Now().UnixNano() / (1000000000 / ticksPerSecond)) },
+		logger:      env.Logger(),
+	}
+
+	go func() {
+		for range mq.ticker.C {
+			mq.reapDedup()
+		}
+	}()
+
+	return mq, nil
+}
+
+func (mq *memQuota) Close() error {
+	mq.ticker.Stop()
+	return nil
+}
+
+func (mq *memQuota) Alloc(args adapter.QuotaArgs) (int64, error) {
+	return mq.alloc(args, false)
+}
+
+func (mq *memQuota) AllocBestEffort(args adapter.QuotaArgs) (int64, error) {
+	return mq.alloc(args, true)
+}
+
+func (mq *memQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (int64, error) {
+	return mq.commonWrapper(args, func(d *adapter.QuotaDefinition, key string) int64 {
+		result := args.QuotaAmount
+
+		// we optimize storage for non-expiring quotas
+		if d.Window == 0 {
+			inUse := mq.cells[key]
+
+			if result > d.MaxAmount-inUse {
+				if !bestEffort {
+					return 0
+				}
+
+				// grab as much as we can
+				result = d.MaxAmount - inUse
+			}
+			mq.cells[key] = inUse + result
+			return result
+		}
+
+		window, ok := mq.windows[key]
+		if !ok {
+			seconds := int32((d.Window + time.Second - 1) / time.Second)
+			window = newRollingWindow(d.MaxAmount, seconds*ticksPerSecond)
+			mq.windows[key] = window
+		}
+
+		currentTick := mq.getTick()
+		if !window.alloc(result, currentTick) {
+			if !bestEffort {
+				return 0
+			}
+
+			// grab as much as we can
+			result = window.available()
+			_ = window.alloc(result, currentTick)
+		}
+
+		return result
+	})
+}
+
+func (mq *memQuota) ReleaseBestEffort(args adapter.QuotaArgs) (int64, error) {
+	return mq.commonWrapper(args, func(d *adapter.QuotaDefinition, key string) int64 {
+		result := args.QuotaAmount
+
+		if d.Window == 0 {
+			inUse := mq.cells[key]
+
+			if result >= inUse {
+				// delete the cell since it contains no useful state
+				delete(mq.cells, key)
+				return inUse
+			}
+
+			mq.cells[key] = inUse - result
+			return result
+		}
+
+		window, ok := mq.windows[key]
+		if !ok {
+			return 0
+		}
+
+		currentTick := mq.getTick()
+		result = window.release(result, currentTick)
+
+		if window.available() == d.MaxAmount {
+			// delete the cell since it contains no useful state
+			delete(mq.windows, key)
+		}
+
+		return result
+	})
+}
+
+type quotaFunc func(d *adapter.QuotaDefinition, key string) int64
+
+func (mq *memQuota) commonWrapper(args adapter.QuotaArgs, qf quotaFunc) (int64, error) {
+	d, ok := mq.definitions[args.Name]
+	if !ok {
+		return 0, fmt.Errorf("request for unknown quota '%s' received", args.Name)
+	}
+
+	if args.QuotaAmount < 0 {
+		return 0, fmt.Errorf("negative quota amount %d received", args.QuotaAmount)
+	}
+
+	if args.QuotaAmount == 0 {
+		return 0, nil
+	}
+
+	key := makeKey(args)
+
+	mq.Lock()
+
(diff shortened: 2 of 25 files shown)


adapter/memQuota/memQuota_test.go
```diff
@@ -0,0 +1,286 @@
...
+		} else {
+			amount, err = a.Alloc(qa)
+		}
+
+		if err != nil {
+			t.Errorf("Expecting success, got %v, case %d", err, i)
+		}
+
+		if amount != c.allocResult {
+			t.Errorf("Expecting %d, got %d, case %d", c.allocResult, amount, i)
```
douglas-reid: how about `("Alloc(%v) => %d, want %d", amount, c.allocResult)` ?

matches expected order from: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

same thing for the below messages.

adapter/memQuota/memQuota_test.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+
+		qa = adapter.QuotaArgs{
```
douglas-reid: can we split this out into a separate TestMemQuota_Release() or similar?

adapter/memQuota/memQuota_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+	qa := adapter.QuotaArgs{
+		Name:        "Q1",
+		QuotaAmount: 10,
+	}
+
+	qa.DeduplicationID = "0"
+	amount, _ := asp.Alloc(qa)
+	if amount != 10 {
+		t.Errorf("Expecting 10, got %d", amount)
```
douglas-reid: as above, suggest something like: `"Alloc() => %d, want 10"` here and below.

adapter/memQuota/rollingWindow.go
```diff
@@ -0,0 +1,85 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package memQuota
+
+import (
```
douglas-reid: can probably be collapsed to just `import "sync"`

adapter/memQuota/rollingWindow.go
```diff
@@ -0,0 +1,85 @@
...
+	last *alloc
+
+	// total amount left in the window
+	avail int64
+
+	// determines how many ticks individual allocs survive before being considered dead
+	ticksInWindow int32
+}
+
+func newRollingWindow(limit int64, ticksInWindow int32) *rollingWindow {
```
douglas-reid: nit: just `ticks`. inWindow is implied by method name.

adapter/memQuota/rollingWindow.go
```diff
@@ -0,0 +1,85 @@
...
+	return &rollingWindow{
+		avail:         limit,
+		ticksInWindow: ticksInWindow,
+	}
+}
+
+func (w *rollingWindow) alloc(amount int64, currentTick int32) bool {
+	// reap old entries
+	for n := w.first; n != nil && n.expirationTick <= currentTick; n = n.next {
```
douglas-reid: any reason not to use "container/list" (or similar) for the window/allocs here (as opposed to having the allocs maintain the pointers themselves)?

adapter/memQuota/rollingWindow.go
```diff
@@ -0,0 +1,85 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package memQuota
+
+import (
+	"sync"
+)
+
+// A single allocation, carrying its amount and expiration time
```
douglas-reid: did we look at https://godoc.org/golang.org/x/time/rate for rate limiting? is there something our implementation provides that is better than the existing rate limiter?

adapter/memQuota/config/config.proto
```diff
@@ -0,0 +1,24 @@
...
+
+syntax = "proto3";
+
+package adapter.memQuota.config;
+
+option go_package="config";
+
+message Params {
+	// Minimum number of seconds that deduplication is possible for a given operation.
+	int32 min_deduplication_window = 1;
```
geeknoid: Done

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,309 @@
...
+	if !ok {
+		return 0, fmt.Errorf("unknown quota request %s received", args.Name)
+	}
+
+	key := makeKey(args)
+	amount := args.QuotaAmount
+
+	mq.Lock()
+
+	if d.Kind == adapter.AllocationKind {
```
geeknoid: Done

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,309 @@
...
+// reapDedup cleans up dedup entries from the oldDedup map and moves all entries from
+// the recentDedup map into the oldDedup map, making those next in line for deletion.
+//
+// This is normally called on a regular basis via a go routine. It's also used directly
+// from tests to inject specific behaviors.
+func (mq *memQuota) reapDedup() {
+	mq.dedupLock.Lock()
+	t := mq.oldDedup
+	mq.oldDedup = mq.recentDedup
+	mq.recentDedup = t
```
geeknoid: Trying to save on garbage.

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// produce a unique key representing the cell
+	var buffer bytes.Buffer
+	buffer.WriteString(args.Name)
+	for _, k := range keys {
+		buffer.WriteString(";")
+		buffer.WriteString(k)
+		buffer.WriteString("=")
+
+		// isn't it amazing that the primitive types like int64 and bool don't implement the Stringer interface?
```
geeknoid: Done.

adapter/memQuota/memQuota_test.go
```diff
@@ -0,0 +1,286 @@
...
+		var err error
+
+		if c.allocBestEffort {
+			amount, err = a.AllocBestEffort(qa)
+		} else {
+			amount, err = a.Alloc(qa)
+		}
+
+		if err != nil {
+			t.Errorf("Expecting success, got %v, case %d", err, i)
```
geeknoid: Done.

Although I did do these test output changes, I find this somewhat busy work and not particularly useful to anyone. When you have a test failure, the exact line of the failure is printed. The first thing you do when you see such a failure is load the test file and go to that line. And on top of that, the fact it's highly likely that the majority of these test failures will never ever be seen.

A large test corpus I worked with previous used a framework that encourage minimal failure messages. The parameters involved in the test case would be printed in a structured way, but no additional prose was included. It made writing tests faster and was just as diagnosable.

Anyway, that's my perspective on these...

adapter/memQuota/memQuota_test.go
geeknoid: Done

adapter/memQuota/memQuota_test.go
geeknoid: I find having it all in one place is more convenient, avoiding a lot of redundancy.

adapter/memQuota/memQuota_test.go
geeknoid: Done

adapter/memQuota/rollingWindow.go
geeknoid: Done

adapter/memQuota/rollingWindow.go
geeknoid: Yes, I looked at a few rate limiters. They seem to all suffer from the '11th hour' problem. That is, if you have a QPS limit of 10, it's possible for someone to allocate 10 units at 0.9s and another 10 units at 1.1s.

My approach is a true rolling window, allowing no more than 10 units in any 1 second interval.

On top of that, my version is considerably more efficient in memory & time :-)

adapter/memQuota/rollingWindow.go
geeknoid: Just 'ticks' doesn't do it for me, I think it's not obvious enough.

geeknoid:
I've addressed most issues, fixed a race condition with how I was handling dedup lookup, and substantial reduced memory churn and steady-state use.

PTAL.

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,335 @@
...
+		result = mq.allocCore(d, key, args.QuotaAmount, bestEffort)
+		mq.recentDedup[args.DeduplicationID] = result
+	}
+
+	mq.Unlock()
+
+	return result, nil
+}
+
+func (mq *memQuota) allocCore(d *adapter.QuotaDefinition, key string, amount int64, bestEffort bool) int64 {
```
mandarjog: can we split this into `rateQuota()` and `allocationQuota()`

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,305 @@
...
+type keyWorkspace struct {
+	keys   []string
+	buffer []byte
+}
+
+// pool of reusable keyWorkspace structs
+var keyWorkspacePool = sync.Pool{New: func() interface{} { return &keyWorkspace{} }}
+
+var (
+	name = "istio/memQuota"
```
mandarjog: I have 2 issues with combining these 2 functions into 1.
1. ratelimit and quota are different aspect `kinds` from a user's perspective.
so 
```
r.RegisterQuota(newQuotaBuilder())
r.RegisterRateLimit(newRateLimitBuilder())
```

2. I think there is very little common between AllocationQuota and RateLimit.
Allocation() is different (windowed, and non-windowed) and release is different. 

If there *is* some common code it can be made into a base class or helper funcs.

geeknoid: Why should these be different aspects? This is adding cognitive load for no value.

The only distinction between an allocation quota and a rate quota is that rate quotas expire after a while. Everything else is identical. There is in fact only a single QuotaDescriptor for this reason, we don't have different types.

geeknoid:
Applied changes corresponding to the update to QuotaDescriptor I've submitted to istio/api.

Basically, the line between rate & allocation quota is removed completely from the API surface. We maintain a distinction in the implementation of the adapter strictly for the sake memory efficiency for non-expiring quotas.

PTAL

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,296 @@
...
+		} else {
+			window, ok := mq.windows[key]
+			if !ok {
+				window = newRollingWindow(d.MaxAmount, d.ExpirationSeconds*ticksPerSecond)
+				mq.windows[key] = window
+			}
+
+			currentTick := mq.getTick()
+			if !window.alloc(result, currentTick) {
+				if bestEffort {
```
douglas-reid: what does it mean to be "bestEffort" for a rate limit?  Is this the equivalent of a "burst" rate above and beyond the normal limit? Should we be modeling it in that manner?

I'm not clear what the expectations are for: "allow 10 qps, best effort".  

With this code, won't it be possible to send unlimited QPS, as window.available() == 0 is not a blocker?

It feels like, for rate limiting, we should be establishing a limit with a configurable burst.

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,296 @@
...
+				delete(mq.cells, key)
+			} else {
+				if result > inUse {
+					// free as much as we can
+					result = inUse
+				}
+
+				mq.cells[key] = inUse - result
+			}
+		} else {
```
douglas-reid: can we rewrite this as:

```
if d.ExpirationSeconds > 0 {
   return 0
}
....
```

(dropping the `else` and pulling the no-op condition up front).

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,296 @@
...
+			result = 0
+		}
+
+		return result
+	})
+}
+
+type quotaFunc func(d *adapter.QuotaDefinition, key string) int64
+
+func (mq *memQuota) common(args adapter.QuotaArgs, qf quotaFunc) (int64, error) {
```
douglas-reid: is there a more descriptive name than `common`?

adapter/memQuota/rollingWindow.go
```diff
@@ -0,0 +1,71 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package memQuota
+
+// Implements a rolling window that allows N units to be allocated per rolling time interval.
+// Time is abstracted in terms of ticks, provided by the caller, decoupling the
+// implementation from real-time, enabling much easier testing and more flexibility.
+type rollingWindow struct {
```
douglas-reid: at the risk of annoying repetition, I still feel like we should be using something like: https://godoc.org/golang.org/x/time/rate#Limiter here. It uses the token bucket algorithm and should avoid the 11th hour issue you've raised earlier.

Doing so has several advantages (well-tested, widely-used, part of the official supplementary libs, provides a configurable burst). It would also be easier to read and understand (defined in terms that match the user config, no "ticks" to mentally map).

The one thing it doesn't provide is insight into remaining available tokens.  But I'm not clear how that will be used by rate limiting clients in practice. This relates to my other question about what best-effort means for rate limits.

pkg/adapter/quota.go
```diff
@@ -34,11 +34,26 @@ type (
...
+	// QuotaKind determines the usage semantics
+	QuotaKind int8
+
+	// QuotaDefinition is used to describe an individual quota the aspect will encounter at runtime
+	QuotaDefinition struct {
+		// MaxAmount specifies the upper limit for the quota
+		MaxAmount int64
+
+		// Size of rolling window in seconds
+		ExpirationSeconds int32
```
douglas-reid: any reason not to take a generic time.Duration here?  Do we care about flexibility for sub-second windows?

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,296 @@
...
+var keyWorkspacePool = sync.Pool{New: func() interface{} { return &keyWorkspace{} }}
+
+var (
+	name = "istio/memQuota"
+	desc = "Simple volatile memory-based quotas."
+	conf = &config.Params{MinDeduplicationWindowSeconds: 1}
+)
+
+const (
+	ticksPerSecond = 10
```
geeknoid: Done

adapter/memQuota/memQuota.go
```diff
@@ -0,0 +1,296 @@
...
+					result = d.MaxAmount - inUse
+				} else {
+					result = 0
+				}
+			}
+
+			if result > 0 {
+				mq.cells[key] = inUse + result
+			}
+		} else {
```
geeknoid: Done

adapter/memQuota/memQuota.go
geeknoid: Done

pkg/adapter/quota.go
geeknoid: Done

adapter/memQuota/memQuota.go
geeknoid: This code doesn't return errors when it can't allocate quota; it returns 0 instead. Errors are used only for actual error conditions such as misconfiguration.

Regarding the semantics, it's the semantics are:

- I can support N units per seconds
- Please allocate M units right now, or as much as you can

I think this is a fairly straightforward model. It'll work well for prefetching tokens in particular.

adapter/memQuota/rollingWindow.go
geeknoid: We're talking about 1 26 line function. What's there now is tested, works, has all the semantics we need, and is very efficient. I'd like to leave as-is.

adapter/memQuota/memQuota.go
geeknoid: Regarding bursting, this is done through a simple composition of limits. For example, a QPS limit of 100 and a QPM limit of 1000. This allows bursts of 100QPS in the broader context of 1000QPM.

douglas-reid: What handles enforcement within Istio?  Is the alloc bit not responsible for ensuring rate limits are respected?

If it is, best effort, as I read the code, is the same as having no limit defined.

adapter/memQuota/memQuota.go
geeknoid: The condition is no longer a no-op, but I did reorg the code as suggested.

geeknoid:
Addressed the comments, except for the suggestion to use the existing rate limiter, for reasons stated in my comments. I've also gone ahead and implement rate limit release semantics.

PTAL.

geeknoid:
The aspect manager is where this decision happens.

   - Alloc returns requestedAmount or 0, depending on if totalAvailable >=
   requestedAmount
   - AllocBestEffort returns requestedAmount or totalAvailable, whichever
   is least

The aspect manager will implement the allow/disallow semantics
appropriately. The proxy will depend on Alloc semantics initially, and
leverage AllocBestEffort once we integrate prefetching.

pkg/aspect/quotaManager.go
```diff
@@ -0,0 +1,117 @@
...
+			if val, ok := labels[a]; ok {
+				args.Labels[a] = val
+				continue
+			}
+			if val, found := attribute.Value(attrs, a); found {
+				args.Labels[a] = val
+			}
+		}
+
+		// TODO: handle return value!
```
douglas-reid: just for clarity: this TODO is the bit we expect to do enforcement (based on your earlier comments), correct?

So, for this to work, this needs to be impl'd?

adapter/memQuota/rollingWindow.go
douglas-reid: OK. I don't think this is a blocker.

If we are having the aspect manager do enforcement, then this seems reasonable (as the existing libraries don't return numTokens consumed, etc.).

I do worry, however, in general about reimplementing vetted solutions for problems.

pkg/aspect/quotaManager.go
geeknoid: Righto, I'll plug this in as I wire up Check and Quota to do something real wrt quotas.

geeknoid:
Added a few more tests and fixed a bug as a result. Coverage of rollingWindow.go is 100%, coverage of memQuota.go is missing 2 lines which is around the background reaper goroutine. The reaper itself is tested, just not the tick-based for loop.

Anyway, all ready for checkin :-)

PR #203, opened 2017-01-27
Builders must have unique names

mandarjog:
At present 2 builders cannot have the same name even if their Aspect is different.

This was causing panic while starting mixer.
We need integration tests :-)

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/203)


mandarjog:
opened issue. https://github.com/istio/mixer/issues/204

PR #205, opened 2017-01-27
Move docker push to istio-testing GCP Project.

douglas-reid:
@kyessenov has requested the move from 'istio-test' to 'istio-testing', as 'istio-testing' is the official testing repo.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/205)


docker/gcloud_build.sh
```diff
@@ -2,8 +2,8 @@
 
 set -ex
 
-gcloud config set project istio-test
+gcloud config set project istio-testing
```
mandarjog: can we add 
`PROJECTNAME=istio-testing`
and then just use env var?

kyessenov: gcloud docker --authorize-only
and then you can push to any gcr bucket you have permissions for.

douglas-reid:
PTAL.

Updated to use env var $PROJECT, as requested.

PR #208, opened 2017-01-28
Introduce prometheus adapter skeleton.

douglas-reid:
This PR is the first step towards full prometheus integration. Prometheus, at the moment, is unique for adapters (pull model v. push). This PR represents an initial stab at what setting up a server from within an adapter might look like.

It is expected that in the pending move to go1.8 the server bits will be rewritten to provide for graceful shutdown of the started server.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/208)


adapter/prometheus/config/prometheus.proto
```diff
@@ -0,0 +1,22 @@
+// Copyright 2017 Google Inc.
```
geeknoid: For the other adapters, we've called this file just config.proto. We should be consistent throughout IMO.

pkg/adapter/registrar.go
```diff
@@ -30,6 +30,9 @@ type Registrar interface {
 
 	// RegisterQuota registers a new Quota builder.
 	RegisterQuota(QuotaBuilder)
+
+	// RegisterMetrics registers a new Metrics factory.
```
geeknoid: factory -> builder. Let's keep the names consistent until/if we refactor again...

pkg/adapterManager/registry.go
```diff
@@ -77,6 +77,11 @@ func (r *registry) RegisterQuota(quota adapter.QuotaBuilder) {
 	r.insert(quota)
 }
 
+// RegisterMetrics registers a new Metrics factory.
```
geeknoid: factory -> builder

adapter/prometheus/prometheus.go
```diff
@@ -0,0 +1,64 @@
...
+		adapter.DefaultBuilder
+
+		srv server
+	}
+
+	prom struct{}
+)
+
+var (
+	name = "prometheus.io/exporter" // TODO: better name?
```
geeknoid: Shouldn't this just be istio/prometheus?

adapter/prometheus/prometheus.go
```diff
@@ -0,0 +1,64 @@
...
+func Register(r adapter.Registrar) {
+	r.RegisterMetrics(newFactory())
+}
+
+func newFactory() factory {
+	return factory{adapter.NewDefaultBuilder(name, desc, conf), newServer(defaultAddr)}
+}
+
+func (f factory) Close() error {
+	if f.srv != nil {
```
geeknoid: Is this conditional needed? When you srv be nil?

adapter/prometheus/server.go
```diff
@@ -0,0 +1,77 @@
...
+func newServer(addr string) server {
+	return &serverInst{addr: addr}
+}
+
+func (s *serverInst) Start() error {
+	var listener net.Listener
+	srv := &http.Server{Addr: s.addr}
+	listener, err := net.Listen("tcp", s.addr)
+	if err != nil {
+		return err
```
geeknoid: This would probably benefit from a custom error message and a log message.

adapter/prometheus/server.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	http.Handle(metricsPath, promhttp.Handler())
+
+	go func() {
+		err := srv.Serve(listener.(*net.TCPListener))
+		if err != nil {
+			// TODO: update and fix
+			log.Println("Prometheus HTTP Server Error - ", err)
```
geeknoid: Please use Env.Logger for logging from adapters.

adapter/prometheus/prometheus.go
douglas-reid: I'm not convinced that third-party integrations should be in the istio namespace. We wouldn't expect istio/gcp or something would we? To me istio/* are for bits where istio is doing everything (list checking, deny, writing to stdio).

Thoughts?

geeknoid:
I've never understood the naming we're using. I would personally just drop
any kind of prefix and just have the Name() method return the coe name of
the adapter proper (ipListChecker, prometheus, etc)

adapter/prometheus/server.go
douglas-reid: This is a bit problematic given how the server is built at the moment. The Env is only provided at NewMetrics() time; the server is built at Factory creation time.

Building the server with NewMetrics() is an option, but comes with other complications, etc.  Worth pursuing?

adapter/prometheus/server.go
douglas-reid: Updated with custom error message.

I don't think we want to log and return an error.  Wherever the error is handled can log appropriately.

adapter/prometheus/prometheus.go
douglas-reid: Removed.

adapter/prometheus/prometheus.go
douglas-reid: Updated to just prometheus.

pkg/adapterManager/registry.go
douglas-reid: done.

pkg/adapter/registrar.go
douglas-reid: done.

adapter/prometheus/config/prometheus.proto
douglas-reid: done.

douglas-reid:
PTAL. Addressed most of the existing comments. Added a bit around Start() that panic()s if Start() returns an error.

adapter/prometheus/server.go
geeknoid: I definitely want all adapter output to be funneled through the Env.Logger such that we can annotate the log data and redirect it appropriately. So I do think it's worth pursuing.

Maybe we should change the Register method to accept Env and not pass it in the NewXXX methods?

douglas-reid:
PTAL.

Modified `Start()` to take an `adapter.Logger`.  I did not attempt to refactor `Register()` to provide an Env, but added a `sync.Once` to the factory.

PR #210, opened 2017-01-31
add ability to read from config maps

mandarjog:
--serviceConfigFile and --globalConfigFile can be specified as
configmap://{namespace}/mapname 
kubeconfig it automatically detected, but it can also be specified.

Since this PR brings in k8s client-go, it adds many dependencies to WORKSPACE.
Some dependencies require auto build file generation. This may fail on a Mac if your file system is `case insensitive`.
To fix it create a `case sensitive` partition on the Mac and point bazel to it.
```--output_base=/Volumes/casesentitive/BAZEL_OUT```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/210)

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -50,6 +50,7 @@ type serverArgs struct {
 	// mixer manager args
 	serviceConfigFile      string
 	globalConfigFile       string
+	kubeConfig             string
 	configFetchIntervalSec uint
 }
 
@@ -81,6 +82,8 @@ func serverCmd(errorf errorFn) *cobra.Command {
 
 	serverCmd.PersistentFlags().StringVarP(&sa.serviceConfigFile, "serviceConfigFile", "", "serviceConfig.yml", "Combined Service Config")
 	serverCmd.PersistentFlags().StringVarP(&sa.globalConfigFile, "globalConfigFile", "", "globalConfig.yml", "Global Config")
+	serverCmd.PersistentFlags().StringVarP(&sa.kubeConfig, "kubeConfig", "", "", "kube Config")
+
 	serverCmd.PersistentFlags().UintVarP(&sa.configFetchIntervalSec, "configFetchInterval", "", 5, "Config fetch interval in seconds")
 
 	return &serverCmd
@@ -132,7 +135,7 @@ func runServer(sa *serverArgs) error {
 	eval := expr.NewIdentityEvaluator()
 	adapterMgr := adapterManager.NewManager(adapter.Inventory(), aspect.Inventory(), eval)
 	configManager := config.NewManager(eval, adapterMgr.AspectValidatorFinder(), adapterMgr.BuilderValidatorFinder(),
-		sa.globalConfigFile, sa.serviceConfigFile, time.Second*time.Duration(sa.configFetchIntervalSec))
+		sa.globalConfigFile, sa.serviceConfigFile, sa.kubeConfig, time.Second*time.Duration(sa.configFetchIntervalSec))
 	handler := api.NewHandler(adapterMgr, adapterMgr.AspectMap())
 
 	grpcServerOptions, err := serverOpts(sa, handler)
--- a/pkg/config/manager.go
+++ b/pkg/config/manager.go
@@ -22,6 +22,8 @@ import (
 
 	"github.com/golang/glog"
 
+	"strings"
+
 	"istio.io/mixer/pkg/adapter"
 	"istio.io/mixer/pkg/attribute"
 	"istio.io/mixer/pkg/expr"
@@ -49,11 +51,13 @@ type Manager struct {
 	loopDelay     time.Duration
 	globalConfig  string
 	serviceConfig string
+	kubeConfig    string
 
-	cl      []ChangeListener
-	closing chan bool
-	scSHA   [sha1.Size]byte
-	gcSHA   [sha1.Size]byte
+	cl        []ChangeListener
+	closing   chan bool
+	scSHA     [sha1.Size]byte
+	gcSHA     [sha1.Size]byte
+	configMap *K8sResourceFetcher
 
 	sync.RWMutex
 	lastError error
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: cmd/server/server.go [hunk 1]
```diff
@@ -50,6 +50,7 @@ type serverArgs struct {
 	// mixer manager args
 	serviceConfigFile      string
 	globalConfigFile       string
+	kubeConfig             string
```
geeknoid: I really don't like to see k8s stuff creeping into the core parts of our design.

Can we encapsulate all details of config maps into a completely separate binary that communicates with mixer's generic API? It'd be a bit more complicated, but it would maintain isolation between our core and k8s.

pkg/config/manager.go
```diff
@@ -182,6 +200,6 @@ func (c *Manager) Start() {
 	// If it is not successful, we will continue to watch for changes.
 	go c.loop()
 	if err != nil {
-		glog.Warning("Unable to process config", err)
+		glog.Warning("Unable to process config ", err)
```
geeknoid: Should have a : in there

cmd/server/server.go
kyessenov: I can help with this binary in Manager repo. It was our original design goal not to couple Mixer with platform specific mechanisms. You need to provide a way to reload configuration in Mixer. There are two generic methods to achieve this:
1) you expose an HTTP endpoint that accepts a config JSON as a POST body
2) you listen for SIG HUP in your process, atomically displace files, and trigger a reload.
Please implement one of these methods.

douglas-reid: Agree.  Can we quickly create an admin API that receives a full config file (or similar) and leave the polling to the manager?

kyessenov:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[pkg/config/k8s.go, line 72 at r1](https://reviewable.io:443/reviews/istio/mixer/210#-Kbq-wRQaOSz5PjWkSqI:-Kbq-wRQaOSz5PjWkSqJ:bvuqi1p) ([raw file](https://github.com/istio/mixer/blob/b3c6c40f7530b8788400d4e56799d58b81ca8958/pkg/config/k8s.go#L72)):*
> ```Go
> 
> 	f.readMap = func(ns string, mapName string) (*v1.ConfigMap, error) {
> 		return k8sClient.ConfigMaps(ns).Get(mapName, meta_v1.GetOptions{})
> ```

What about watching for changes?
k8s doesn't recommend changing config map files, so you need another layer of indirection, where you send the name of a new config map file to mixer to fetch.

---

*[pkg/config/manager.go, line 100 at r1](https://reviewable.io:443/reviews/istio/mixer/210#-Kbq-GwAafM-PfuzqZNY:-Kbq-GwB7sVtT8APqpxQ:b-r9eqps) ([raw file](https://github.com/istio/mixer/blob/b3c6c40f7530b8788400d4e56799d58b81ca8958/pkg/config/manager.go#L100)):*
> ConfigMapScheme+"://"
you should use FQ k8s URL for config maps (API object URI)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/210)*

mandarjog:
*[cmd/server/server.go, line 53 at r1](https://reviewable.io:443/reviews/istio/mixer/210#-Kbnx2H0WYdX9ksXpJ7H-r1-53:-Kbq4Qn7gq-i92jhVXaj:b-i4zqks) ([raw file](https://github.com/istio/mixer/blob/b3c6c40f7530b8788400d4e56799d58b81ca8958/cmd/server/server.go#L53)):*

Agree.  Can we quickly create an admin API that receives a full config file (or similar) and leave the polling to the manager?
</blockquote></details>

most of pkg/config is going to become the mixer manager. We can (and plan to do) all of the above. I don't think these changes can be made today for the hackathon. 
I can abandon this PR. hackathon will mount configmaps and use filename args that  mixer already supports.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/210)*

mandarjog:
*[pkg/config/k8s.go, line 72 at r1](https://reviewable.io:443/reviews/istio/mixer/210#-Kbq-wRQaOSz5PjWkSqI:-Kbq5JabWnBj8J6cu0rG:b-bn8oua) ([raw file](https://github.com/istio/mixer/blob/b3c6c40f7530b8788400d4e56799d58b81ca8958/pkg/config/k8s.go#L72)):*
<details><summary><i>Previously, kyessenov (Kuat) wrote…</i></summary><blockquote>

What about watching for changes?
k8s doesn't recommend changing config map files, so you need another layer of indirection, where you send the name of a new config map file to mixer to fetch.
</blockquote></details>

It is polling for changes.

Do they recommend that config maps be immutable ? Can you point me to that discussion ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/210)*

kyessenov:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[pkg/config/k8s.go, line 72 at r1](https://reviewable.io:443/reviews/istio/mixer/210#-Kbq-wRQaOSz5PjWkSqI:-KbqD3x_0eigbm7tbVrI:b-efr07k) ([raw file](https://github.com/istio/mixer/blob/b3c6c40f7530b8788400d4e56799d58b81ca8958/pkg/config/k8s.go#L72)):*

It is polling for changes.

Do they recommend that config maps be immutable ? Can you point me to that discussion ?
</blockquote></details>

You should use Watch interface, not Get. It uses a long-living connection that streams updates.
Discussion is spread-out around attempts to use config maps to use as the dynamic update mechanism. There are all sorts of things you need to think about related to pod life cycle, ownership, rollout once you use config maps for this purpose.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/210)*

PR #211, opened 2017-01-31
Wire up quotas to mixer's check call.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/211)

### DIFF
--- a/adapter/memQuota/memQuota.go
+++ b/adapter/memQuota/memQuota.go
@@ -93,26 +93,43 @@ func newBuilder() builder {
 	return builder{adapter.NewDefaultBuilder(name, desc, conf)}
 }
 
+func (builder) ValidateConfig(cfg adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	c := cfg.(*config.Params)
+
+	if c.MinDeduplicationWindowSeconds <= 0 {
+		ce = ce.Appendf("MinDeduplicationWindowSeconds", "deduplication window of %d is invalid, must be >= 0", c.MinDeduplicationWindowSeconds)
+	}
+
+	return
+}
+
 func (builder) NewQuota(env adapter.Env, c adapter.AspectConfig, d map[string]*adapter.QuotaDefinition) (adapter.QuotaAspect, error) {
 	return newAspect(env, c.(*config.Params), d)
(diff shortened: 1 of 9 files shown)


adapter/memQuota/memQuota_test.go
```diff
@@ -335,6 +345,54 @@ func TestReaper(t *testing.T) {
 	}
 }
 
+func TestReaperTicker(t *testing.T) {
```
douglas-reid: do we need this test?  I think we can assume that the Ticker works as expected.  The sleep() approach to testing it seems fragile -- and adding a callback or something just for testing seems wasteful.

testdata/globalconfig.yml
```diff
@@ -1,9 +1,7 @@
 subject: "namespace:ns"
 revision: "2022"
 adapters:
-  - name: denychecker.1
-    kind: istio/denyChecker
-    impl: istio/denyChecker
+  - name: memQuota
```
douglas-reid: should we call this something other than memQuota to avoid potential confusion with the implementation?

adapter/memQuota/memQuota.go
```diff
@@ -93,19 +93,34 @@ func newBuilder() builder {
 	return builder{adapter.NewDefaultBuilder(name, desc, conf)}
 }
 
+func (builder) ValidateConfig(cfg adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	c := cfg.(*config.Params)
+
+	if c.MinDeduplicationWindowSeconds <= 0 {
+		ce = ce.Appendf("MinDeduplicationWindow", "deduplication window of %d is invalid, must be >= 0", c.MinDeduplicationWindowSeconds)
```
geeknoid: Done

adapter/memQuota/memQuota_test.go
geeknoid: I'm not a fan of the sleep, but I *am* a big fan of 100% coverage. I'm happy to entertain additional options ;-)

testdata/globalconfig.yml
geeknoid: Changed to 'default'

PR #217, opened 2017-02-01
Add support for simple evalPredicate

mandarjog:
Single attribute is evaluated for equality
'$' is used as an attribute designator

selector: $service_name == abcd
or
selector: $request_status_code == 200

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/217)

### DIFF
--- a/pkg/expr/identity.go
+++ b/pkg/expr/identity.go
@@ -22,11 +22,13 @@ import (
 	"istio.io/mixer/pkg/attribute"
 )
 
-// identity is an evaluator that expects mapExpression to be a key in the attribute bag. It does no evaluation, only lookup.
+// identity is an evaluator that expects mapExpression to be a key in the attribute bag.
+// It does equality-only evaluation in EvalPredicate.
+// Otherwise it does no evaluation, only lookup.
 type identity struct{}
 
-// NewIdentityEvaluator returns an evaluator that performs no evaluations; instead it uses the provided mapExpression
-// as the key into the attribute bag.
+// NewIdentityEvaluator returns an evaluator that performs equality-only evaluation in EvalPredicate;
+// Otherwise it uses the provided mapExpression as the key into the attribute bag.
 func NewIdentityEvaluator() Evaluator {
 	return identity{}
 }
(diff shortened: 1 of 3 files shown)


pkg/expr/identity_test.go
```diff
@@ -143,38 +145,73 @@ func TestEvalPredicate(t *testing.T) {
...
 			t.Errorf("Failed to create bag for case: %v; err: %v", c, err)
 		}
 		id := NewIdentityEvaluator()
+		if err = id.Validate(c.in); err != nil {
+			t.Errorf("[%d] Unexpected error %s", idx, err.Error())
+		}
 		val, err := id.EvalPredicate(c.in, bag)
-		if c.err && err == nil {
-			t.Errorf("Expected err on case '%+v' but got none.", c)
+		if c.err != "" && err == nil {
```
douglas-reid: i think this might be clearer with some slight organization changes. something like:

```
if c.err == "" {
    if err != nil {
       t.Errorf("unexpected error...")
       continue
    }
    if val != c.out {
       t.Errorf("got %v, want %v"...)
    }
    continue
}
if err == nil {
    t.Errorf("expected error"...)
    continue
}
if !strings.Contains(err.Error(), c.err) {
    t.Errorf("got %v, want %v"...)
}
```

Following the `c.err ==` and `c.err !=` as they weave is a bit tricky.

pkg/expr/identity.go
```diff
@@ -45,12 +47,53 @@ func (identity) EvalString(mapExpression string, bag attribute.Bag) (string, err
 	return "", fmt.Errorf("%s not in attribute bag", mapExpression)
```
geeknoid: Is this a message that can make it out to operators? If so, it shouldn't use the term "attribute bag"

pkg/expr/identity_test.go
```diff
@@ -143,38 +145,73 @@ func TestEvalPredicate(t *testing.T) {
 		attrs *mixerpb.Attributes
 		in    string
 		out   interface{}
-		err   bool
+		err   string
```
mandarjog: We do want to know what expr thought the error was. If it is unresolved attribute, or invalid expression etc.

mandarjog:
Updates made. PTAL.

PR #218, opened 2017-02-01
Add a delay between closing gRPC streams and closing the underlying s…

geeknoid:
…ocket.

This works around a bug in gRPC server-side code. A bug has been filed for
this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/218)

### DIFF
--- a/cmd/client/util.go
+++ b/cmd/client/util.go
@@ -57,6 +57,10 @@ func createAPIClient(port string, enableTracing bool) (*clientState, error) {
 }
 
 func deleteAPIClient(cs *clientState) {
+	// TODO: This is to compensate for this bug: https://github.com/grpc/grpc-go/issues/1059
+	//       Remove this delay once that bug is fixed.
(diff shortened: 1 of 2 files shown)


cmd/client/util.go
```diff
@@ -57,6 +57,10 @@ func createAPIClient(port string, enableTracing bool) (*clientState, error) {
 }
 
 func deleteAPIClient(cs *clientState) {
+	// TODO: This is to compensate for this bug: https://github.com/grpc/grpc-go/issues/1059
+	//       Remove this delay once that bug is fixed.
+	time.Sleep(time.Duration(50) * time.Millisecond)
```
ZackButcher: nit: time.Duration is an alias for an int64, this can be simplified to `time.Sleep(50 * time.Millisecond)` and the compiler will handle the type coercion.

geeknoid: Indeed. Done.

PR #220, opened 2017-02-01
For now, quota is handled by the Check call.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/220)

### DIFF
--- a/pkg/aspect/registry.go
+++ b/pkg/aspect/registry.go
@@ -39,6 +39,7 @@ func Inventory() []APIBinding {
 		{NewListCheckerManager(), config.CheckMethod},
 		{NewLoggerManager(), config.ReportMethod},
 		{NewAccessLoggerManager(), config.ReportMethod},
+		{NewQuotaManager(), config.CheckMethod},
 		{NewQuotaManager(), config.QuotaMethod},
 	}
 }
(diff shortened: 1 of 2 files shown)


qiwzhang:
Does that mean istio/proxy not need to make a separate Quota() call?

geeknoid:
For this week only. We're doing this so we can get end-to-end functionality
up and running without having to add the quota call into the proxy.
Ultimately, the proxy will need to make the call though.

pkg/aspect/registry.go
```diff
@@ -39,6 +39,6 @@ func Inventory() []APIBinding {
 		{NewListCheckerManager(), config.CheckMethod},
 		{NewLoggerManager(), config.ReportMethod},
 		{NewAccessLoggerManager(), config.ReportMethod},
-		{NewQuotaManager(), config.QuotaMethod},
+		{NewQuotaManager(), config.CheckMethod},
```
douglas-reid: should we add instead of replace? something like:

```
{NewQuotaManager(), config.QuotaMethod},
{NewQuotaManager(), config.CheckMethod},
```

I think that would allow us to test the transition.

geeknoid: Done.

geeknoid:
Addressed feedback, and fixed some issues in the canonical adapter testing code.

PR #223, opened 2017-02-02
Streamline log entry types.

douglas-reid:
AccessLogger can re-use `adapter.LogEntry`: AccessLog -> TextPayload. This will reduce redundant types in `//pkg/adapter`.

This PR preserves the separation of `accessLogger` and `logger`, as the two aspects have different configurations (and user expectations), but enables generic logging adapters to handle the entries generated by both aspects in a common fashion, if so desired.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/223)


douglas-reid:
Yeah, this may be rearranging deck chairs, but the code for stdioLogger bothered me, because it was so simple -- and yet required duplication of the serialization code just to support extremely similar entry types. I imagine any logging backend will support both aspects, so they should be able to share as much code as possible.

PR #226, opened 2017-02-02
Create a generic testing Env for use in adapter tests.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/226)

### DIFF
--- a/adapter/memQuota/memQuota_test.go
+++ b/adapter/memQuota/memQuota_test.go
@@ -15,7 +15,6 @@
 package memQuota
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


adapter/memQuota/memQuota_test.go
```diff
@@ -25,28 +24,7 @@ import (
 	at "istio.io/mixer/pkg/adapter/testing"
```
douglas-reid: suggestion: we rename the `//pkg/adapter/testing` package to avoid the collision with the standard `testing` package. Since they will only ever be used together, it would be nice to not have to rename it on every import. 

How about `//pkg/adapter/test` ?  That would give us code that read like `test.NewEnv(t)`.

PR #227, opened 2017-02-02
Allow multiple builders by the same name, per aspect kind.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/227)

### DIFF
--- a/pkg/adapterManager/registry.go
+++ b/pkg/adapterManager/registry.go
@@ -20,20 +20,25 @@ import (
 	"github.com/golang/glog"
 	"istio.io/mixer/pkg/adapter"
 	"istio.io/mixer/pkg/aspect"
-	"istio.io/mixer/pkg/config"
 )
+
+// BuildersByName holds a set of builders of the same aspect kind, indexed by their name.
+type BuildersByName map[string]adapter.Builder
+
+// BuildersPerKind holds a set of builders, indexed by aspect kind.
+type BuildersPerKind map[string]BuildersByName
 
 // registry implements pkg/adapter/Registrar.
 // registry is initialized in the constructor and is immutable thereafter.
-// All registered builders must have unique names.
+// All registered builders must have unique names per aspect kind.
 // It also implements builderFinder that manager uses.
 type registry struct {
-	buildersByName map[string]adapter.Builder
+	builders BuildersPerKind
 }
 
 // newRegistry returns a new Builder registry.
 func newRegistry(builders []adapter.RegisterFn) *registry {
-	r := &registry{buildersByName: make(map[string]adapter.Builder)}
+	r := &registry{make(BuildersPerKind)}
 	for idx, builder := range builders {
 		glog.V(2).Infof("Registering [%d] %v", idx, builder)
 		builder(r)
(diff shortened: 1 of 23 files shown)


pkg/adapterManager/registry.go
```diff
@@ -20,20 +20,19 @@ import (
 	"github.com/golang/glog"
 	"istio.io/mixer/pkg/adapter"
 	"istio.io/mixer/pkg/aspect"
-	"istio.io/mixer/pkg/config"
 )
 
 // registry implements pkg/adapter/Registrar.
 // registry is initialized in the constructor and is immutable thereafter.
 // All registered builders must have unique names.
```
ZackButcher: Update this comment -> "All registered builders must have a unique name per aspect kind"

pkg/adapterManager/registry.go
```diff
@@ -20,20 +20,19 @@ import (
...
 // It also implements builderFinder that manager uses.
 type registry struct {
-	buildersByName map[string]adapter.Builder
+	builders map[string]map[string]adapter.Builder
```
ZackButcher: Can you add a comment noting that the outer map is by kind and the inner map is by name? The types aren't helpful here.

douglas-reid: Maybe we could have a type named KindNameBuilderMap or something to document it here and the other places in the code?

pkg/aspect/inventory.go
```diff
@@ -18,18 +18,22 @@ import (
...
-
 // APIBinding associates an aspect with an API method
 type APIBinding struct {
 	Aspect Manager
 	Method config.APIMethod
 }
 
+// Name of all supported aspect kinds.
+const (
+	AccessLogKind = "access-logs"
```
douglas-reid: i feel like we need/want the tautological comments for these exported consts here.

pkg/adapterManager/registry.go
geeknoid: Done

pkg/adapterManager/registry.go
geeknoid: Done

pkg/aspect/inventory.go
geeknoid: What do you mean?

douglas-reid: I was guessing that it was a lint issue (not documenting an exported element). I _feel_ like that used to be a golint issue, but maybe not now? or maybe my memory is hazy?

geeknoid:
I didn't have a comment at first, and one of the linters said I needed a
comment either on each element or on the block as a whole. Putting the
comment on the block as a whole keeps all the linters happy.

testdata/globalconfig.yml
```diff
@@ -2,16 +2,16 @@ subject: "namespace:ns"
 revision: "2022"
 adapters:
   - name: default
-    kind: istio/quota
-    impl: istio/memQuota
+    kind: quotas
```
mandarjog: With an adapter providing multiple aspects, kind field should be removed from the proto.

Issue #201, opened 2017-01-27
Delineate config and request state by adding Request Args


mandarjog:
It results in a more understandable configuration that can be validated early.

It creates a clean separation between constructor configuration and request state.

Based on 
https://docs.google.com/document/d/1PkGbjYeiNzcx8l43E0diXoXKhXHHRo8_QA1_ghyFdMk/edit#heading=h.s01ziym7eywb

We need to have different messages to represent config parameters vs. request parameters. The current listChecker configuration would go from:

```proto
message ListCheckerParams {
  bool blacklist = 1;
  string check_attribute = 2;
}

to:

// configuration, consistent for a deployment
message ListCheckerConfig {
  bool blacklist = 1;
}

// per-request state, changes for each request flowing through the mixer
message ListCheckerRequestArgs {
  string symbol = 2;
}
```

Following is an example of how one would do this metric and quota descriptors.

```yml
 - kind: ratelimiter
    params:
      limits:  # imposes 2 limits, 100/s per source and destination
      - limit: "100/s"
        labels:
          - label: src.user_id
          - label: target.service_id
       - limit: "1000/s"  # every destination service gets 1000/s
        labels:
          - label: target.service_id

metric_descriptor:
 name: "test_metric"
 type: DELTA / INT64
 label: "user_name", string
 label: "url", Url   

-----
- kind: metric
  params:
      ["test_metric", "test_metric1"]
  requestArgs:
   metric:    --> map[string]string
     - value: $request_count
       descriptor: "test_metric"
       user_name: ($foo | $bar)
       url: ($fooUrl | $barUrl) 
```
      
```python    
 AspectManager.Execute(cfg, bag):
         me := MetricEntry()
         for ra in cfg.RequestArgs:
                 md = lookupMetricDesc(ra.descriptor)
                 me.user_name = evalString(ra.user_name, bag)
                 me.url = (md[user_name].type) evalUrl(ra.url, bag)

          w.asp.send(me)
```

mandarjog:
@geeknoid In the `iplistchecker` the `configArgs` and `requestArgs` are orthogonal.
We could therefore define requestArgs as a typed message.
requestArgs is indeed what is being passed down to the aspect.

In the metrics above, it appears that the shape of `requestArgs` falls out of the metric descriptor. Which means requestArgs cannot be more prescriptive than a `list of maps[string]string`

The keys of the `RequestArgs` map depend on the specific values of the metric descriptor.
[...]

PR #231, opened 2017-02-02
Add mechanism for passing metrics descriptor info to adapters.

douglas-reid:
This PR creates a basic MetricDefinition type and alters the NewMetrics() signature to pass in a list of definitions at build time.

Metrics backends will likely need to know what metrics they will be receiving values for before those values come streaming in. Prometheus, for example, will want to register metrics once up front, and then create observations against that registry when values are passed in.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/231)

### DIFF
--- a/pkg/adapter/metrics.go
+++ b/pkg/adapter/metrics.go
@@ -19,11 +19,21 @@ import (
 	"time"
 )
 
+// Metric kinds supported by mixer.
 const (
-	// GAUGE is used to record instantaneous (non-cumulative) measurement
-	GAUGE Kind = iota
-	// COUNTER is used to record increasing cumulative values.
-	COUNTER
+	Gauge   Kind = iota // records instantaneous (non-cumulative) measurements
+	Counter             // records increasing cumulative values
+)
+
+// Label kinds supported by mixer.
+const (
+	String LabelKind = iota
+	Int64
+	Float64
+	Bool
+	Time
+	IPAddress
+	URL
 )
 
 type (
@@ -61,7 +71,7 @@ type (
 	}
 
 	// Kind defines the set of known metrics types that can be generated
-	// by istio.
+	// by the mixer.
 	Kind int
 
 	// MetricsBuilder builds instances of the Metrics aspect.
@@ -80,9 +90,13 @@ type (
 		// Kind provides type information about the metric.
 		Kind Kind
 		// Labels are the names of keys for dimensional data that will
-		// generated at runtime and passed along with metric values.
-		Labels []string
+		// be generated at runtime and passed along with metric values.
+		Labels map[string]LabelKind
 	}
+
+	// LabelKind defines the set of known label types that can be generated
+	// by the mixer.
+	LabelKind int
 )
 
 // String returns the string-valued metric value for a metrics.Value.
--- a/pkg/adapter/metrics.go
+++ b/pkg/adapter/metrics.go
@@ -69,7 +69,19 @@ type (
 		Builder
 
 		// NewMetricsAspect returns a new instance of the Metrics aspect.
-		NewMetricsAspect(env Env, config AspectConfig) (MetricsAspect, error)
+		NewMetricsAspect(env Env, config AspectConfig, metrics []MetricDefinition) (MetricsAspect, error)
+	}
+
+	// MetricDefinition provides the basic description of a metric schema
+	// for which metrics adapters will be sent Values at runtime.
+	MetricDefinition struct {
+		// Name is the canonical name of the metric.
+		Name string
+		// Kind provides type information about the metric.
+		Kind Kind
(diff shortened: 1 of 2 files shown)


pkg/adapter/metrics.go
```diff
@@ -69,7 +69,19 @@ type (
...
+	// MetricDefinition provides the basic description of a metric schema
+	// for which metrics adapters will be sent Values at runtime.
+	MetricDefinition struct {
+		// Name is the canonical name of the metric.
+		Name string
+		// Kind provides type information about the metric.
+		Kind Kind
+		// Labels are the names of keys for dimensional data that will
+		// generated at runtime and passed along with metric values.
+		Labels []string
```
geeknoid: will BE generated...

Wouldn't it be desirable to know the type of the label values? When talking to GCM for example, I think it likes to know when things are integers/floats rather than just strings. This could be a map[string]config.AttributeDescriptor.ValueType (or whatever that enum is called in Go...). Maybe that map deserves to be its own type (LabelMap?] and then we can use it elsewhere like in QuotaDefinition.

ZackButcher: +1, I think having the label's type on hand is valuable. Also, have we spent any time thinking about how labels are created/defined? This is relevant for the statsd adapter I'm working on: statsd metrics are dimensionless so the common pattern is to encode dimensions in the metric's name. Given the current interface, these dimensions would have to be passed into the aspect impl as labels.

douglas-reid: I thought about passing in a map to types briefly but had forgotten that we also had attribute descriptors and so skipped it. Thanks for the reminder!

I'll update to send a type map. I think i'll probably add a layer of indirection between the descriptor enum and what gets down, but yes, this seems like ultimately something we should support.

douglas-reid: @ZackButcher labels are defined in the metric descriptor. they are a different word for attribute name/value pairs. so, the manager would read the descriptor, get a list of attr names, look them up in the attr.Bag and put them in the Labels map and pass them along.

geeknoid: Zack, labels are created by the user via descriptors. Individual descriptors specify the set of labels in terms of name/type (so "user_name" or type "string", or "provider_url" of type "Url").

See issue #201 for more context.

ZackButcher: Ah, of course: I remember us talking about it now. What threw me off was that we haven't updated the protos in the api repo to reflect that decision yet.

douglas-reid:
PTAL. I added a type: `LabelKind`. I was a bit torn on whether to include things like DNS and Email, as there aren't any (that I know) corresponding golang types for those. I left them out for now, so that there would be a 1:1 correspondence between the LabelKinds and the types of values in the map[string]interface{} delivered to the backends.

pkg/adapter/metrics.go
```diff
@@ -19,11 +19,21 @@ import (
...
-	// GAUGE is used to record instantaneous (non-cumulative) measurement
-	GAUGE Kind = iota
-	// COUNTER is used to record increasing cumulative values.
-	COUNTER
+	Gauge   Kind = iota // records instantaneous (non-cumulative) measurements
+	Counter             // records increasing cumulative values
+)
+
+// Label kinds supported by mixer.
+const (
```
geeknoid: Any reason why this doesn't include all the types defined in AttributeDescriptor.ValueType?

ZackButcher: Would it be bad to just use that enum directly as the values for the map?

douglas-reid: I added my justification in the comments above.

I'm not convinced there is value in supporting types that aren't supported by the language, at the moment.  Even IPAddr and URL seem a bit overkill for our initial integration pieces. Prometheus supports string labels only. Statsd doesn't even support labels. Stackdriver only supports String, Bool, and Int64.

I don't think we want to include that enum directly. We certainly haven't been passing descriptors down directly in the rest of the code. And I think we want to insulate backend adapters from any API changes, as best we can.

douglas-reid: One more data point: AWS CloudWatch Metric Dimensions are string valued.

geeknoid:
I think we should be consistent in our API surface. Our attributes are
defined with all these types, and it should be possible for the user to
define labels using all the same types. The adapter can trivially convert
whatever it doesn't understand to string.

Otherwise, we'll have a mismash of support for different types in different
places and that'll just be confusing.

Note that I'd like to use the same definitions for quotas, security
principals, and other descriptors too.

douglas-reid:
OK. I've added the full set. I expect all metrics adapters will ignore most of these types, but this now matches the API surface, and can be used for other aspects, etc.

PR #239, opened 2017-02-03
Statsd Adapter

ZackButcher:
We introduce an adapter implementing the metrics aspect which exports metrics to statsd.

Lots of open questions in this PR around the error handling and config validation, these are recorded as TODOs inline.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/239)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -210,6 +210,6 @@ new_go_repository(
 
 new_go_repository(
     name = "com_github_cactus_statsd_client",
-    commit = "91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d",
+    commit = "91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d", # release 3.1.0
     importpath = "github.com/cactus/go-statsd-client",
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -210,6 +210,6 @@ new_go_repository(
 
 new_go_repository(
     name = "com_github_cactus_statsd_client",
-    commit = "91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d", # release 3.1.0
+    commit = "91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d", # release 3.1.0, 5/30/2016
     importpath = "github.com/cactus/go-statsd-client",
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -207,3 +207,9 @@ new_go_repository(
     commit = "fa8ad6fec33561be4280a8f0514318c79d7f6cb6", # Feb 12, 2015 (only release too old)
     importpath = "github.com/prometheus/client_model",
 )
+
+new_go_repository(
+    name = "com_github_cactus_statsd_client",
+    commit = "91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d",
+    importpath = "github.com/cactus/go-statsd-client",
+)
--- a/adapter/statsd/config/BUILD
+++ b/adapter/statsd/config/BUILD
@@ -1,0 +1,19 @@
+load("@org_pubref_rules_protobuf//go:rules.bzl", "go_proto_library")
+
+go_proto_library(
+    name = "go_default_library",
+    imports = [
+        "external/com_github_google_protobuf/src",
+    ],
+    inputs = [
+        "@com_github_google_protobuf//:well_known_protos",
+    ],
+    protos = [
+        "config.proto",
+    ],
+    verbose = 0,
+    visibility = ["//adapter/statsd:__pkg__"],
+    deps = [
+        "@com_github_golang_protobuf//ptypes/duration:go_default_library",
+    ],
+)
--- a/adapter/statsd/config/BUILD
+++ b/adapter/statsd/config/BUILD
@@ -1,7 +1,10 @@
-load("@org_pubref_rules_protobuf//gogo:rules.bzl", "gogo_proto_library")
+load("@org_pubref_rules_protobuf//gogo:rules.bzl", "gogoslick_proto_library")
 
-gogo_proto_library(
+gogoslick_proto_library(
     name = "go_default_library",
+    importmap = {
+        "google/protobuf/duration.proto": "github.com/gogo/protobuf/types",
+    },
     imports = [
         "external/com_github_google_protobuf/src",
     ],
@@ -14,6 +17,7 @@ gogo_proto_library(
     verbose = 0,
     visibility = ["//adapter/statsd:__pkg__"],
     deps = [
-        "@com_github_golang_protobuf//ptypes/duration:go_default_library",
+        "@com_github_gogo_protobuf//sortkeys:go_default_library",
+        "@com_github_gogo_protobuf//types:go_default_library",
     ],
 )
--- a/adapter/statsd/config/BUILD
+++ b/adapter/statsd/config/BUILD
@@ -1,6 +1,6 @@
-load("@org_pubref_rules_protobuf//go:rules.bzl", "go_proto_library")
+load("@org_pubref_rules_protobuf//gogo:rules.bzl", "gogo_proto_library")
 
-go_proto_library(
+gogo_proto_library(
     name = "go_default_library",
     imports = [
         "external/com_github_google_protobuf/src",
--- a/adapter/statsd/config/config.proto
+++ b/adapter/statsd/config/config.proto
@@ -29,11 +29,11 @@ message Params {
 
     // Flush Interval controls the maximum amount of time between sending metrics to the statsd collection server.
     // Metrics are reported when either flush_bytes is full or flush_interval time has elapsed since the last report.
-    google.protobuf.Duration flush_interval = 3;
+    google.protobuf.Duration flush_duration = 3;
 
     // Maximum UDP packet size to send; if not specified defaults to 512 bytes. If the statsd server is running on the
     // same (private) network 1432 bytes is recommended for better performance.
-    int64 flush_bytes = 4;
+    int32 flush_bytes = 4;
 
     // Chance that any particular metric is sampled when incremented; can take the range [0, 1], defaults to 1 if unspecified.
     float sampling_rate = 5;
@@ -43,5 +43,7 @@ message Params {
     // `action_name-response_code`. The template strings must conform to go's text/template syntax. For the example of
     // `action_name-response_code`, we use the template:
     //    `{{.apiMethod}}-{{.responseCode}}`
+    //
+    // If a metric's name is not in the map then the exported statsd metric name will be exactly the metric's name.
     map<string, string> metric_name_template_strings = 6;
 }
--- a/adapter/statsd/config/config.proto
+++ b/adapter/statsd/config/config.proto
@@ -45,5 +45,7 @@ message Params {
     //    `{{.apiMethod}}-{{.responseCode}}`
     //
     // If a metric's name is not in the map then the exported statsd metric name will be exactly the metric's name.
+    //
+    // TODO: unify templates to use our expression language when its finalized
     map<string, string> metric_name_template_strings = 6;
 }
--- a/adapter/statsd/config/config.proto
+++ b/adapter/statsd/config/config.proto
@@ -1,0 +1,47 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package adapter.statsd.config;
+
+option go_package = "config";
+
+import "google/protobuf/duration.proto";
+
+message Params {
+    // Address of the statsd server, e.g. localhost:8125
+    string address = 1;
+
+    // Metric prefix, do not specify for no prefix
+    string prefix = 2;
+
+    // Flush Interval controls the maximum amount of time between sending metrics to the statsd collection server.
+    // Metrics are reported when either flush_bytes is full or flush_interval time has elapsed since the last report.
+    google.protobuf.Duration flush_interval = 3;
+
+    // Maximum UDP packet size to send; if not specified defaults to 512 bytes. If the statsd server is running on the
+    // same (private) network 1432 bytes is recommended for better performance.
+    int64 flush_bytes = 4;
+
+    // Chance that any particular metric is sampled when incremented; can take the range [0, 1], defaults to 1 if unspecified.
+    float sampling_rate = 5;
+
+    // Map of metric name -> template string; the template will be filled with values from the metric's labels and the
+    // resulting string will be used as the statsd metric name. This allows easier creation of statsd metrics like
+    // `action_name-response_code`. The template strings must conform to go's text/template syntax. For the example of
+    // `action_name-response_code`, we use the template:
+    //    `{{.apiMethod}}-{{.responseCode}}`
+    map<string, string> metric_name_template_strings = 6;
+}
--- a/adapter/statsd/statsd.go
+++ b/adapter/statsd/statsd.go
@@ -1,0 +1,161 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package statsd
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+	"time"
+
+	"github.com/cactus/go-statsd-client/statsd"
+	"github.com/golang/protobuf/ptypes/duration"
+
+	"istio.io/mixer/adapter/statsd/config"
+	"istio.io/mixer/pkg/adapter"
+)
+
+type (
+	// builder holds a connection to the statsd server which is shared by all aspects, as well as the parsed
+	// templates for constructing statsd metric names.
+	builder struct {
+		adapter.DefaultBuilder
+
+		client    statsd.Statter
+		templates map[string]*template.Template // metric name -> template
+	}
+
+	aspect struct {
+		rate      float32
+		client    statsd.Statter
+		templates map[string]*template.Template // metric name -> template
+	}
+)
+
+var (
+	name = "statsd"
+	desc = "Pushes statsd metrics"
+	// conf encodes the default configuration for this adapter
+	conf = &config.Params{
+		Address:                   "localhost:8125",
+		Prefix:                    "",
+		FlushInterval:             &duration.Duration{Seconds: 0, Nanos: int32(300 * time.Millisecond)},
+		FlushBytes:                512,
+		SamplingRate:              1.0,
+		MetricNameTemplateStrings: make(map[string]string),
+	}
+)
+
+// Register records the builders exposed by this adapter.
+func Register(r adapter.Registrar) {
+	r.RegisterMetricsBuilder(newBuilder())
+}
+
+func newBuilder() *builder {
+	client, err := statsd.NewBufferedClient(conf.Address, conf.Prefix, toTime(conf.FlushInterval), int(conf.FlushBytes))
+	if err != nil {
+		// TODO: something other than panic?
+		panic("failed to construct statsd client with err: " + err.Error())
+	}
+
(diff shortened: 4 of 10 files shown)


WORKSPACE
```diff
@@ -206,3 +206,9 @@ new_go_repository(
     commit = "fa8ad6fec33561be4280a8f0514318c79d7f6cb6",
     importpath = "github.com/prometheus/client_model",
 )
+
+new_go_repository(
+    name = "com_github_cactus_statsd_client",
+    commit = "91c326c3f7bd20f0226d3d1c289dd9f8ce28d33d", # release 3.1.0
```
douglas-reid: let's add a date to the comment here too.

adapter/statsd/config/BUILD
```diff
@@ -0,0 +1,19 @@
+load("@org_pubref_rules_protobuf//go:rules.bzl", "go_proto_library")
```
douglas-reid: please change everything to gogoprotobuf.

so: gogo:rules.bzl, gogo_proto_library.

adapter/statsd/config/config.proto
```diff
@@ -0,0 +1,47 @@
...
+message Params {
+    // Address of the statsd server, e.g. localhost:8125
+    string address = 1;
+
+    // Metric prefix, do not specify for no prefix
+    string prefix = 2;
+
+    // Flush Interval controls the maximum amount of time between sending metrics to the statsd collection server.
+    // Metrics are reported when either flush_bytes is full or flush_interval time has elapsed since the last report.
+    google.protobuf.Duration flush_interval = 3;
```
douglas-reid: let's call this `flush_duration` to match api style guidelines.

adapter/statsd/config/config.proto
```diff
@@ -0,0 +1,47 @@
...
+
+    // Chance that any particular metric is sampled when incremented; can take the range [0, 1], defaults to 1 if unspecified.
+    float sampling_rate = 5;
+
+    // Map of metric name -> template string; the template will be filled with values from the metric's labels and the
+    // resulting string will be used as the statsd metric name. This allows easier creation of statsd metrics like
+    // `action_name-response_code`. The template strings must conform to go's text/template syntax. For the example of
+    // `action_name-response_code`, we use the template:
+    //    `{{.apiMethod}}-{{.responseCode}}`
+    map<string, string> metric_name_template_strings = 6;
```
douglas-reid: what is default behavior for metric name not in map?  can we specify in comment?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,161 @@
...
+
+func newBuilder() *builder {
+	client, err := statsd.NewBufferedClient(conf.Address, conf.Prefix, toTime(conf.FlushInterval), int(conf.FlushBytes))
+	if err != nil {
+		// TODO: something other than panic?
+		panic("failed to construct statsd client with err: " + err.Error())
+	}
+
+	templates := make(map[string]*template.Template)
+	for metricName, s := range conf.MetricNameTemplateStrings {
```
douglas-reid: why do this in `newBuilder`?  do we care about parsing templates from our default conf for any reason up front?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	params, ok := c.(*config.Params)
+	if !ok {
+		// TODO: should this just report an error?
+		panic(fmt.Errorf("invalid config type, want adapter/statsd/config/config.proto, got: %v", params))
+	}
+
+	// TODO: we need to validate that the provided template strings are satisfiable by the mixer's attributes/metric's labels
```
douglas-reid: alternatively, we could adjust the way "missing" values are handled in this adapter (or even in the template itself `{{or ".apiMethod" "default"}}`, etc.)

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,161 @@
...
+		if err := t.Execute(buf, value.Labels); err != nil {
+			// TODO: we should be able to verify that this will never happen at config time, should this panic?
+			return fmt.Errorf("failed to create metric name with template '%s' and labels '%v'; got err: %s", t.Name(), value.Labels, err)
+		}
+		name = buf.String()
+	}
+
+	switch value.Kind {
+	case adapter.Gauge:
+		v, err := value.Int64()
```
douglas-reid: do we need handling for float64()s here (or below)? i feel like latency numbers may arrive as float64s (though the prototype may not be a good source to go by on this, as it did client-side aggregation).

also, i don't see anything related to client.Timing.  does that imply we aren't currently interested in generating distributions from observations?  if we get a bunch of latency reports, i feel like we want to generate a distribution... but I'm not sure how that is done.

Thoughts?

adapter/statsd/config/config.proto
```diff
@@ -0,0 +1,47 @@
...
+    // Metric prefix, do not specify for no prefix
+    string prefix = 2;
+
+    // Flush Interval controls the maximum amount of time between sending metrics to the statsd collection server.
+    // Metrics are reported when either flush_bytes is full or flush_interval time has elapsed since the last report.
+    google.protobuf.Duration flush_interval = 3;
+
+    // Maximum UDP packet size to send; if not specified defaults to 512 bytes. If the statsd server is running on the
+    // same (private) network 1432 bytes is recommended for better performance.
+    int64 flush_bytes = 4;
```
geeknoid: int64 seems a bit excessive for a packet size :-)

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,161 @@
...
+			templates[metricName] = t
+		}
+	}
+
+	return &builder{adapter.NewDefaultBuilder(name, desc, conf), client, templates}
+}
+
+func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	params, ok := c.(*config.Params)
+	if !ok {
```
geeknoid: No need for this check, we guarantee the right type of config is delivered to the adapters.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,161 @@
...
+		// TODO: should this just report an error?
+		panic(fmt.Errorf("invalid config type, want adapter/statsd/config/config.proto, got: %v", params))
+	}
+
+	// TODO: we need to validate that the provided template strings are satisfiable by the mixer's attributes/metric's labels
+	for metricName, s := range params.MetricNameTemplateStrings {
+		if _, err := template.New(metricName).Parse(s); err != nil {
+			ce = ce.Append("template "+metricName, fmt.Errorf("failed to parse template '%s' for metric '%s' with err: %s", s, metricName, err))
```
geeknoid: The first parameter to Append should just be the field name in the config where the error occurred. This will be used for error messages to the user.

adapter/statsd/statsd.go
ZackButcher: The templates are a user input, though. We could provide a library of "standard" statsd metric types, like `api_method-response_code` that handles the missing attribute cases, but I don't think that can be the full solution. The access logger has the same problem when a user provides a custom log template, right?

adapter/statsd/config/config.proto
ZackButcher: To be fair, int32 is too :P

adapter/statsd/statsd.go
ZackButcher: Reading the statsd docs, it looks like gauges can support floats. However, the statsd client library I'm using (which appeared to be the best/most popular based on searching around) accepts only int64 inputs, no floats.

I can see if they'd be amendable to a PR that added support for floats for gauges.

ZackButcher: And based on my reading, statsd distributions (histograms) are configured on the collection side, not the publishing side. So the user can configure their server to retain metric "foo" as a histogram, and configure Istio to report a time for metric "foo", which their collection server will retain as a histogram.

As far as supporting timing and other metric types, we just need to add more metric kinds; I didn't want to conflate adding more kinds with the PR to add the adapter.

ZackButcher:
Updated, PTAL.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/duration"
+
+	"istio.io/mixer/adapter/statsd/config"
+	"istio.io/mixer/pkg/adapter"
+)
+
+type (
+	// builder holds a connection to the statsd server which is shared by all aspects, as well as the parsed
+	// templates for constructing statsd metric names.
```
geeknoid: Comment is out of date.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,168 @@
...
+	params := cfg.(*config.Params)
+	templates := make(map[string]*template.Template)
+	for metricName, s := range params.MetricNameTemplateStrings {
+		def, found := findMetric(metrics, metricName)
+		if !found {
+			continue // we don't have a metric that corresponds to this template, keep on going
+		}
+
+		t, err := template.New(metricName).Parse(s)
+		if err != nil {
```
geeknoid: Trust in ValidateConfig, no need for this redundant check.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func newBuilder() *builder {
+	flushDuration, err := ptypes.Duration(conf.FlushDuration)
+	if err != nil {
+		// we panic because we're reading in our static default config
+		panic(fmt.Errorf("failed to parse default duration, %v, as a time.Duration", conf.FlushDuration))
+	}
+
+	client, err := statsd.NewBufferedClient(conf.Address, conf.Prefix, flushDuration, int(conf.FlushBytes))
```
mandarjog: This always creates a connection to "localhost" during Registration, seems incorrect.
If your intent is to internally share  `client statsd.Statter` across all aspect that use the same `conf.Address`, then you will need to implement a statd.Statter cache.
Connection should only be opened during call to `newaspect.`

adapter/statsd/statsd_test.go
```diff
@@ -0,0 +1,274 @@
...
+func TestNewBuilder(t *testing.T) {
+	conf = &config.Params{
+		Address:                   "localhost:8125",
+		Prefix:                    "",
+		FlushDuration:             &duration.Duration{Seconds: 0, Nanos: int32(300 * time.Millisecond)},
+		FlushBytes:                512,
+		SamplingRate:              1.0,
+		MetricNameTemplateStrings: map[string]string{"a": `{{ .apiMethod "-" .responseCode }}`},
+	}
+	b := newBuilder()
```
mandarjog: At present we have no way to pass parameters to newBuilder during registration. I don;t think these params belong here.

adapter/statsd/statsd.go
ZackButcher: Yup, totally forgot to move this along with all the other setup; it's in the correct spot now.

adapter/statsd/config/config.proto
```diff
@@ -0,0 +1,49 @@
...
+    int32 flush_bytes = 4;
+
+    // Chance that any particular metric is sampled when incremented; can take the range [0, 1], defaults to 1 if unspecified.
+    float sampling_rate = 5;
+
+    // Map of metric name -> template string; the template will be filled with values from the metric's labels and the
+    // resulting string will be used as the statsd metric name. This allows easier creation of statsd metrics like
+    // `action_name-response_code`. The template strings must conform to go's text/template syntax. For the example of
+    // `action_name-response_code`, we use the template:
+    //    `{{.apiMethod}}-{{.responseCode}}`
```
ZackButcher: Added a TODO.

ZackButcher:
Updated, PTAL. Totally forgot to move client creation down with the rest, it's done in `NewMetricsAspect` now.

ZackButcher:
Also, open question about the implementation of `Record`: as written it bails on the first error when attempting to record a value. Would it be better to attempt to record all values, aggregating any errors we get, rather than stopping at the first error?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+
+func newBuilder() *builder {
+	return &builder{adapter.NewDefaultBuilder(name, desc, defaultConf)}
+}
+
+func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	params := c.(*config.Params)
+
+	flushDuration, err := ptypes.Duration(params.FlushDuration)
+	if err != nil {
```
geeknoid: Probably should check for flushDuration <= 0

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"text/template"
+	"time"
+
+	"github.com/cactus/go-statsd-client/statsd"
+	"github.com/golang/protobuf/ptypes"
+
+	"github.com/golang/protobuf/ptypes/duration"
```
geeknoid: Move up into previous block?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+		rate      float32
+		client    statsd.Statter
+		templates map[string]*template.Template // metric name -> template
+	}
+)
+
+var (
+	name        = "statsd"
+	desc        = "Pushes statsd metrics"
+	defaultConf = &config.Params{
```
geeknoid: Do we need a default address? Anything make sense for that, like localhost?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+func newBuilder() *builder {
+	return &builder{adapter.NewDefaultBuilder(name, desc, defaultConf)}
+}
+
+func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	params := c.(*config.Params)
+
+	flushDuration, err := ptypes.Duration(params.FlushDuration)
+	if err != nil {
+		ce = ce.Append("FlushDuration", fmt.Errorf("could not parse as time.Duration with err: %s", err))
```
geeknoid: I think you can just use Append("FlushDuration", err)

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if params.SamplingRate < 0 {
+		ce = ce.Append("SamplingRate", fmt.Errorf("sampling rate must be >= 0"))
+	}
+
+	if _, err := statsd.NewBufferedClient(params.Address, params.Prefix, flushDuration, int(params.FlushBytes)); err != nil {
```
geeknoid: Does this call hit the network? 'Cause ValidateConfig needs to be a parse-level thing, not a runtime thing. So it shouldn't be hitting the network.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+// Register records the builders exposed by this adapter.
+func Register(r adapter.Registrar) {
+	r.RegisterMetricsBuilder(newBuilder())
+}
+
+func newBuilder() *builder {
+	return &builder{adapter.NewDefaultBuilder(name, desc, defaultConf)}
+}
+
+func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
```
geeknoid: Disallow negative flushBytes?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		templates[metricName] = t
+	}
+	return &aspect{params.SamplingRate, client, templates}, nil
+}
+
+func (a *aspect) Record(values []adapter.Value) error {
+	for _, v := range values {
+		if err := a.record(v); err != nil {
+			// TODO: should we keep track of all of the errors and return a composite, rather than aborting at the first problem?
```
geeknoid: Yes, best effort and return a composite failure. This is a runtime environment and we want to succeed as much as we can.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,171 @@
...
+	return nil
+}
+
+func (a *aspect) record(value adapter.Value) error {
+	name := value.Name
+	if t, found := a.templates[value.Name]; found {
+		buf := new(bytes.Buffer)
+		if err := t.Execute(buf, value.Labels); err != nil {
+			// TODO: should this panic? this indicates that the value we were given wasn't filled in completely,
+			// since we validate in NewMetricsAspect that all templates are satisfiable with the metric's labels.
```
geeknoid: Yes, panic. This is not a recoverable error.

adapter/statsd/statsd.go
ZackButcher: Both this and flushBytes are checked by the statsd constructor. If flushDuration <= 0 they set it to 300 milliseconds, if flushBytes <= 0 they set it to 1432 (which is why we check that one: so we can default to a smaller packet size).

geeknoid: I'd rather if == 0 was the default and < 0 were error cases.

adapter/statsd/statsd.go
ZackButcher: It does - it listens on the address. Updated to just attempt to resolve the address.

adapter/statsd/statsd.go
ZackButcher: AFAIK the default would/should be "localhost:8125", but I don't know if we'll ever actually want to use that. Thoughts?

geeknoid: If there's a non-zero chance someone would just prop up a default statsd service and a default mixer to use it and things just work, then why not make that the default address?

adapter/statsd/statsd.go
ZackButcher: Done

adapter/statsd/statsd.go
ZackButcher: Added reasonable defaults for all fields of the config.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,178 @@
...
+	if flushDuration < time.Duration(0) {
+		ce = ce.Append("FlushDuration", fmt.Errorf("flush duration must be >= 0"))
+	}
+	if params.FlushBytes < 0 {
+		ce = ce.Append("FlushBytes", fmt.Errorf("flush bytes must be >= 0"))
+	}
+	if params.SamplingRate < 0 {
+		ce = ce.Append("SamplingRate", fmt.Errorf("sampling rate must be >= 0"))
+	}
+	if _, err := net.ResolveUDPAddr("udp", params.Address); err != nil {
```
geeknoid: As before, I think ValidateConfig is about format verification and not about runtime verification. In other words, ValidateConfig should be fully deterministic. Checking if an address can resolve is non-deterministic and should therefore be treated as a runtime error instead.

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,178 @@
...
+			result = multierror.Append(result, err)
+		}
+	}
+	return result.ErrorOrNil()
+}
+
+func (a *aspect) record(value adapter.Value) error {
+	name := value.Name
+	if t, found := a.templates[value.Name]; found {
+		buf := new(bytes.Buffer)
```
geeknoid: Should this be a reusable pool of buffers?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,178 @@
...
+		if err := t.Execute(buf, value.Labels); err != nil {
+			panic(fmt.Errorf("failed to create metric name with template '%s' and labels '%v'; got err: %s", t.Name(), value.Labels, err))
```
geeknoid: Only untested line, can we cover it?

adapter/statsd/statsd.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
+	params := c.(*config.Params)
+	flushDuration, err := ptypes.Duration(params.FlushDuration)
+	if err != nil {
+		ce = ce.Append("FlushDuration", err)
+	}
+	if flushDuration < time.Duration(0) {
+		ce = ce.Append("FlushDuration", fmt.Errorf("flush duration must be >= 0"))
```
ZackButcher: Append requires that the second arg be an `error`, not just a `string`.

ZackButcher: Oh, misread your comment. Yes, we can swap to `Appendf`.

adapter/statsd/statsd.go
ZackButcher: Done

adapter/statsd/statsd.go
ZackButcher: Turns out my understanding of the template code was a bit wrong: `Execute` only throws due to invalid templates (which we check for in ValidateConfig), not due to not being able to satisfy the template with the input.

So for the example of the template `{{.apiMethod}}-{{.responseCode}}`, trying to `Execute` it with a value that has no labels results in the string `<no value>-<no value>`. I'm pretty convinced that this panic is unreachable given validated config. Should we keep it around, just ignore that error, or change it to a log statement and default to the metric's original name?

geeknoid:
In the last few days, I've pushed coverage to 100% in a few different
modules. On every occasion, I've found actual problems in the last chunk of
uncovered code :-)

I'd say remove the panic and just leave _ = t.Execute() and maybe a comment
that validation ensures correct operation.

ZackButcher:
Removed the panic at execute and added a comment.

douglas-reid:
With switch to gogoslick, the proto config bits should be updated. denyChecker is a good source of a similar switch.

PR #242, opened 2017-02-03
Fix a few bugs in the ipListChecker adapter.

geeknoid:
This boosts ipListChecker coverage to 100%, simplifies the logic, and fixes a few
race bugs found as a result of the new tests.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/242)

### DIFF
--- a/adapter/ipListChecker/config/config.proto
+++ b/adapter/ipListChecker/config/config.proto
@@ -16,6 +16,8 @@ syntax = "proto3";
 
 package adapter.ipListChecker.config;
 
+import "google/protobuf/duration.proto";
+
 option go_package="config";
 
 message Params {
@@ -24,13 +26,11 @@ message Params {
 
     // Determines how often the provider is polled for
     // an updated list
-    // unit: seconds
-    int64 refresh_interval = 2;
+    google.protobuf.Duration refresh_interval = 2;
 
     // Indicates how long to keep a list before discarding it.
     // Typically, the TTL value should be set to noticeably longer (> 2x) than the
     // refresh interval to ensure continued operation in the face of transient
     // server outages.
-    // unit: seconds
-    int64 ttl = 3;
+    google.protobuf.Duration ttl = 3;
 }
--- a/adapter/ipListChecker/ipListChecker.go
+++ b/adapter/ipListChecker/ipListChecker.go
@@ -17,6 +17,7 @@ package ipListChecker
 import (
 	"crypto/sha1"
 	"errors"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -25,7 +26,10 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/duration"
 	"gopkg.in/yaml.v2"
+
 	"istio.io/mixer/adapter/ipListChecker/config"
 	"istio.io/mixer/pkg/adapter"
 )
@@ -34,18 +38,19 @@ type (
 	builder struct{ adapter.DefaultBuilder }
 
 	listChecker struct {
-		log             adapter.Logger
-		providerURL     string
-		atomicList      atomic.Value
-		refreshInterval time.Duration
-		ttl             time.Duration
-		closing         chan bool
-		client          http.Client
+		log           adapter.Logger
+		providerURL   string
+		atomicList    atomic.Value
+		client        http.Client
+		closing       chan bool
+		refreshTicker *time.Ticker
+		purgeTimer    *time.Timer
+		ttl           time.Duration
 	}
 
 	listState struct {
-		payload    []*net.IPNet
-		payloadSha [sha1.Size]byte
+		entries    []*net.IPNet
+		entriesSHA [sha1.Size]byte
 		fetchError error
 	}
 )
@@ -55,8 +60,8 @@ var (
 	desc = "Checks whether an IP address is present in an IP address list."
 	conf = &config.Params{
 		ProviderUrl:     "http://localhost",
-		RefreshInterval: 60,
-		Ttl:             120,
+		RefreshInterval: &duration.Duration{Seconds: 60},
+		Ttl:             &duration.Duration{Seconds: 300},
 	}
 )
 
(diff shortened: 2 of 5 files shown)


adapter/ipListChecker/ipListChecker_test.go
```diff
@@ -197,6 +214,75 @@ func TestValidateConfig(t *testing.T) {
...
+
+	// cause a refetch
+	refreshChan <- time.Now()
+	<-fetched
+
+	// cause a purge
+	purgeChan <- time.Now()
+
+	// wait for the purge
+	for len(a.getListState().entries) == 0 {
```
ZackButcher: Should this be `len(a.getListState().entries) > 0`?

adapter/ipListChecker/ipListChecker.go
```diff
@@ -89,27 +91,52 @@ func (builder) ValidateConfig(cfg adapter.AspectConfig) (ce *adapter.ConfigError
...
-	go l.listRefresher()
+	_ = l.fetchList()
+	purgeTimer.Reset(l.ttl)
+
+	// go routine to periodically refresh the list
+	go func() {
+		for range refreshTicker.C {
+			if err := l.fetchList(); err == nil {
+				// reset the next purge time
+				purgeTimer.Reset(l.ttl)
```
douglas-reid: do we need to check if the Timer is active before resetting here?

https://golang.org/pkg/time/#Timer.Reset

adapter/ipListChecker/ipListChecker.go
```diff
@@ -137,87 +164,63 @@ func (l *listChecker) CheckList(symbol string) (bool, error) {
...
-	list := l.getList()
-
 	resp, err := l.client.Get(l.providerURL)
 	if err != nil {
-		list.fetchError = err
-		l.setList(list)
+		ls := l.getListState()
+		ls.fetchError = err
+		l.setListState(ls)
 		l.log.Warningf("Could not connect to %s: %v", l.providerURL, err)
```
douglas-reid: (here and elsewhere) i think we should try to avoid logging an error and returning it as well.  if we want to log the error, we should do it somewhere where it is actually handled (rather than passed up).

adapter/ipListChecker/ipListChecker.go
```diff
@@ -89,27 +91,52 @@ func (builder) ValidateConfig(cfg adapter.AspectConfig) (ce *adapter.ConfigError
...
+	_ = l.fetchList()
+	purgeTimer.Reset(l.ttl)
+
+	// go routine to periodically refresh the list
+	go func() {
+		for range refreshTicker.C {
+			if err := l.fetchList(); err == nil {
+				// reset the next purge time
+				purgeTimer.Reset(l.ttl)
+			}
```
douglas-reid: i think this is where we should log failures to fetchList() with the Warningf() bits.

adapter/ipListChecker/ipListChecker_test.go
```diff
@@ -109,14 +107,33 @@ func TestBasic(t *testing.T) {
...
-	if len(list.payload) != 2 {
-		t.Errorf("Expecting %d, got %d entries", len(badList.WhiteList)-1, len(list.payload))
+	// now try to parse a list with errors
+	listToUse = 1
+	if err := lc.fetchList(); err != nil {
+		t.Errorf("Unable to refresh list: %v", err)
+	}
+	list := lc.getListState()
+	if len(list.entries) != 2 {
+		t.Errorf("Expecting %d, got %d entries", len(list1.WhiteList)-1, len(list.entries))
```
douglas-reid: i know you disagree, but I will mention again that go convention is ordering this [got, want] v. [want, got].

`Note that the order here is actual != expected, and the message uses that order too. Some test frameworks encourage writing these backwards: 0 != x, "expected 0, got x", and so on. Go does not.`

https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

adapter/ipListChecker/ipListChecker_test.go
geeknoid: Done.

adapter/ipListChecker/ipListChecker.go
```diff
@@ -25,6 +25,8 @@ import (
 	"sync/atomic"
 	"time"
```
geeknoid: Done.

adapter/ipListChecker/ipListChecker.go
```diff
@@ -34,18 +36,18 @@ type (
...
+		refreshTicker *time.Ticker
+		purgeTimer    *time.Timer
+		ttl           time.Duration
 	}
 
 	listState struct {
-		payload    []*net.IPNet
-		payloadSha [sha1.Size]byte
+		entries    []*net.IPNet
+		entriesSha [sha1.Size]byte
```
geeknoid: Done.

adapter/ipListChecker/ipListChecker.go
```diff
@@ -89,27 +91,52 @@ func (builder) ValidateConfig(cfg adapter.AspectConfig) (ce *adapter.ConfigError
...
-	l.client = http.Client{Timeout: l.ttl}
-	l.setList(listState{})
+	l.setListState(listState{})
 
 	// load up the list synchronously so we're ready to accept traffic immediately
-	l.refreshList()
-
-	// crank up the async list refresher
-	go l.listRefresher()
+	_ = l.fetchList()
```
geeknoid: Not particularly. The problem will be logged, like other failures. I'll add a comment to that effect.

adapter/ipListChecker/ipListChecker.go
geeknoid: Good catch.

adapter/ipListChecker/ipListChecker_test.go
geeknoid: I definitely disagree with the importance of such formatting conventions. It really adds nothing to the debugging of the test failure. It's not easier to read, not easier to understand, and above all, it will be seen by probably no one.

It's interesting to note that the particular bit of advice about useful test failures in the code review doc immediately precedes the worse piece of style advice I've ever seen, namely "Variable names in Go should be short rather than long". I point this out to highlight the fact it's OK to ignore suggestions that don't help.

adapter/ipListChecker/ipListChecker.go
geeknoid: I added the error return value so that I can have a conditional statement to decide what to do with the purge timer. If I move logging out of this, then I'll need to put it at both call sites of fetchList. Seems suboptimal.

adapter/ipListChecker/ipListChecker_test.go
geeknoid: Consider this from test code:

	if err != nil {
		t.Errorf("Unable to create aspect: %v", err)
	}

Based on the recommendation, this should be written as:

	if nil != err {
		t.Errorf("Creating an aspect failed with %v, expecting success", err)
	}

The if statement doesn't appear consistent with go conventions.

geeknoid:
Comments addressed.

PTAL.

adapter/ipListChecker/ipListChecker_test.go
```diff
@@ -22,44 +22,42 @@ import (
 
 	"gopkg.in/yaml.v2"
 
+	"time"
```
douglas-reid: this should be in the top block.

mandarjog: we really need a better `goimports` that does these things. should not be a code review thing.

ZackButcher: +1, super annoying that goimports won't touch whitespace.

adapter/ipListChecker/config/config.proto
```diff
@@ -25,12 +25,12 @@ message Params {
     // Determines how often the provider is polled for
     // an updated list
     // unit: seconds
-    int64 refresh_interval = 2;
+    int32 refresh_interval = 2;
```
douglas-reid: nit: append `_seconds` to field names

ZackButcher: Why not use a duration proto so the unit is not ambiguous?

geeknoid: Switched to Duration.

adapter/ipListChecker/ipListChecker_test.go
geeknoid: Done.

geeknoid:
Switched time quantities to the Duration type. I'll do this in memQuota too.

PTAL.

geeknoid:
Ping...

PR #243, opened 2017-02-04
Annotate SHAs in WORKSPACE with date and release info.

douglas-reid:
I've examined each of the SHAs in the workspace and tried to annotate them appropriately. For some, I rolled back to the most recent releases. For deps with no releases, I moved forward to the latest commit, excepting deps that I had to peg the SHAs at what was a known good SHA.

Tested with `bazel clean` and `bazel test ...`, as well as starting `mixs` and sending a ReportRequest via `mixc`.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/243)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,7 +2,7 @@ workspace(name = "com_github_istio_mixer")
 
 git_repository(
     name = "io_bazel_rules_go",
-    commit = "76c63b5cd0d47c1f2b47ab4953db96c574af1c1d",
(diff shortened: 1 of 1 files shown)


mandarjog:
Do we need to do this manually every so often?

douglas-reid:
Every time we update or add a dependency, we should annotate the date and release information.  I'm not sure about how often we should update our dependencies, but it might be nice to track grpc updates, etc.

PR #246, opened 2017-02-07
Add Report functionality to prometheus adapter.

douglas-reid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/246)


douglas-reid:
Hold on review, please. There are some things I want to rearrange after sleeping on it.

douglas-reid:
OK. Fixed up registry/gatherer stuff. Please take a look whenever you have a moment. Thanks.

adapter/prometheus/prometheus.go
```diff
@@ -32,40 +38,185 @@ type (
...
 	if serverErr != nil {
 		return nil, fmt.Errorf("could not start prometheus server: %v", serverErr)
 	}
 
-	return &prom{}, nil
+	metricsMap := make(map[string]prometheus.Collector, len(metrics))
+	for _, m := range metrics {
+		switch m.Kind {
+		case adapter.Gauge:
+			metricsMap[m.Name] = mustRegisterOrGet(newGaugeVec(m.Name, m.Description, m.Labels))
```
ZackButcher: We have the ability to return an error here, why not change `mustRegisterOrGet` to `registerOrGet` and return an error instead of panicking?

adapter/prometheus/prometheus.go
```diff
@@ -32,40 +38,185 @@ type (
...
+	return c
+}
+
+func safeName(n string) string {
+	s := strings.TrimPrefix(n, "/")
+	return charReplacer.Replace(s)
+}
+
+func promValue(val adapter.Value) (float64, error) {
+	switch i := val.MetricValue.(type) {
```
ZackButcher: I feel like we can do a better job here given that we're the ones constructing the value (in the aspect manager). AFAIK the value must be produced from eval'ing some expression over the attributes, so I think the only types you need to worry about are those that can be produced from an evaluation: int64, float64, and maybe string.

adapter/prometheus/prometheus.go
douglas-reid: Done.

adapter/prometheus/prometheus.go
douglas-reid: OK. I was trying to cover any future changes to the contract. I've removed all but the suggested types in latest push.

douglas-reid:
Updated based on initial comments. PTAL.

adapter/prometheus/prometheus.go
```diff
@@ -32,40 +38,174 @@ type (
...
+		case adapter.Gauge:
+			c, err := registerOrGet(newGaugeVec(m.Name, m.Description, m.Labels))
+			if err != nil {
+				metricErr = multierror.Append(metricErr, fmt.Errorf("could not register metric: %v", err))
+			}
+			metricsMap[m.Name] = c
+		case adapter.Counter:
+			c, err := registerOrGet(newCounterVec(m.Name, m.Description, m.Labels))
+			if err != nil {
+				metricErr = multierror.Append(metricErr, fmt.Errorf("could not register metric: %v", err))
```
ZackButcher: I think we want a `continue` here and above, right?

douglas-reid: Yes. Done.

adapter/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,294 @@
...
+		{"One Gauge", []adapter.MetricDefinition{gaugeNoLabels}},
+		{"One Counter", []adapter.MetricDefinition{counterNoLabels}},
+		{"Multiple Metrics", []adapter.MetricDefinition{counterNoLabels, gaugeNoLabels}},
+		{"With Labels", []adapter.MetricDefinition{counter}},
+	}
+
+	for _, v := range tests {
+		t.Run(v.name, func(t *testing.T) {
+			if _, err := f.NewMetricsAspect(test.NewEnv(t), &config.Params{}, v.metrics); err != nil {
+				t.Errorf("NewMetricsAspect() => unexpected error: %v", err)
```
ZackButcher: Small nit: I like @mandarjog's style of including the test index in the error so I know immediately which case failed. Can you add that to the error strings here and elsewhere? Atm there's no way for me to know which of the cases is broken without manually debugging.

douglas-reid: Actually, the test.Run with the name of the test should do just that. I'm using the support for subtests here.

https://blog.golang.org/subtests

ZackButcher: Oh I missed that, nice.

PR #250, opened 2017-02-08
Switch to gogoslick for protobuf compilation in mixer.

douglas-reid:
This PR moves us from the base gogoprotobuf to the gogoslick plugin.

More details on the implications fo gogoslick can be found [here](https://github.com/gogo/protobuf).

In particular, the following changes apply:
- imports gogoprotobuf
- no XXX_unrecognized, fewer pointer fields
- generated string, gostring and equal methods

Any additional options can be added to the protos themselves.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/250)


douglas-reid:
Checked with @ZackButcher, confirmed that this merge will go before the statsd merge.

PR #251, opened 2017-02-09
Tag docker images with date (and build number).

douglas-reid:
Since bazel strips out the date information (to give us repeatable builds!), this
will add a helpful way to distinguish between images, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/251)

### DIFF
--- a/docker/gcloud_build.sh
+++ b/docker/gcloud_build.sh
@@ -16,4 +16,7 @@ else
     bazel --output_base=$BAZEL_OUTBASE run //docker:mixer gcr.io/$PROJECT/mixer:$DOCKER_TAG
 fi
 
+docker tag gcr.io/$PROJECT/mixer:$DOCKER_TAG gcr.io/$PROJECT/mixer:latest
+
 gcloud docker -- push gcr.io/$PROJECT/mixer:$DOCKER_TAG
+gcloud docker -- push gcr.io/$PROJECT/mixer:latest
--- a/docker/gcloud_build.sh
+++ b/docker/gcloud_build.sh
@@ -4,11 +4,16 @@ set -ex
 
 gcloud docker --authorize-only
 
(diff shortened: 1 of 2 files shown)


docker/gcloud_build.sh
```diff
@@ -4,11 +4,16 @@ set -ex
...
+fi
+
 if [ -z $BAZEL_OUTBASE ]
 then
-    bazel run //docker:mixer gcr.io/$PROJECT/mixer:experiment
+    bazel run //docker:mixer gcr.io/$PROJECT/mixer:$DOCKER_TAG
 else
-    bazel --output_base=$BAZEL_OUTBASE run //docker:mixer gcr.io/$PROJECT/mixer:experiment
+    bazel --output_base=$BAZEL_OUTBASE run //docker:mixer gcr.io/$PROJECT/mixer:$DOCKER_TAG
 fi
```
mandarjog: I think while pulling the build, we still want to be able to say

gcr.io/$PROJECT/mixer:latest   otherwise I have to lookup sha everytime.

douglas-reid: Added support for latest.

PTAL.

docker/gcloud_build.sh
```diff
@@ -4,11 +4,19 @@ set -ex
...
 
-gcloud docker -- push gcr.io/$PROJECT/mixer:experiment
+docker tag gcr.io/$PROJECT/mixer:$DOCKER_TAG gcr.io/$PROJECT/mixer:latest
+
+gcloud docker -- push gcr.io/$PROJECT/mixer:$DOCKER_TAG
+gcloud docker -- push gcr.io/$PROJECT/mixer:latest
```
mandarjog: nice :-) that was simple.

douglas-reid: and hopefully it will work :) !

PR #252, opened 2017-02-09
Do a variety of code cleanup improvements suggested by Gogland code inspection

geeknoid:
- Remove unnecessary import aliases.

- Remove redundant use of type names in declarations.

- Rename variables that conflict with imported package names.

- Stop assigning to method receivers.

- Sort imports.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/252)


pkg/aspect/applicationLogsManager.go
```diff
@@ -114,8 +114,8 @@ func (e *applicationLogsWrapper) Execute(attrs attribute.Bag, mapper expr.Evalua
 	// TODO: would be nice if we could use a mutable.Bag here and could pass it around
 	// labels holds the generated attributes from mapper
 	labels := make(map[string]interface{})
-	for attr, expr := range e.inputs {
-		if val, err := mapper.Eval(expr, attrs); err == nil {
+	for attr, e := range e.inputs {
```
mandarjog: Reuse of 'e' as loop var and wrapper

pkg/adapter/configError.go
```diff
@@ -50,25 +50,27 @@ type ConfigError struct {
 // in configuration. The field parameter indicates the name of the
 // specific configuration field name that is problematic.
 func (e *ConfigErrors) Appendf(field, format string, args ...interface{}) *ConfigErrors {
-	if e == nil {
-		e = &ConfigErrors{}
+	ce := e
```
mandarjog: What purpose does 'ce' serve? Multi error and proto code do assign to receiver.

geeknoid: Assigning to the variable 'e' has no impact on the caller of the method. Assigning to the receiver is generally bad form, like assigning to incoming method parameters.

pkg/aspect/applicationLogsManager.go
geeknoid: Renamed to exp

PR #253, opened 2017-02-09
Various aspect-related improvements.

geeknoid:
- Introduce the type aspect.Kind and implement its use. aspect.Kind replaces
strings to represent aspect kinds with an enum, providing better usabiity
and robustness in the code.

- Moved the definition of APIMethods into the aspect
package, instead of the config package. The config package is
agnostic to the API methods and should stay that way.

- Eliminate the APIBinding type, and replace
with the ManagerInventory type which provides
a more direct API-centric access to the
managers.

- Simplify ValidatorFinder pattern by using a function pointer instead of interfaces.

- Support aspects with no adapter name specified, defaults to 'default'
commit 962c1394f626c3822f45ab4550cfd55968d73460

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/253)


pkg/config/runtime.go
```diff
@@ -105,19 +104,15 @@ func (r *Runtime) resolveRules(bag attribute.Bag, aspectSet AspectSet, rules []*
...
 
 	var adp *pb.Adapter
 	// find matching adapter
 	// assume that config references are correct
-	if aa.GetAdapter() != "" {
-		adp = r.adapterByName[aa.GetAdapter()]
-	} else {
-		adp = r.adapterByKind[aa.GetKind()][0]
-	}
+	adp = r.adapterByName[aa.Adapter]
```
mandarjog: aspect specifies kind or adapter.  If adapter is specified by should lookup the adapter by name. If no adapter name is specified we use the 1st adapter available of the correct kind.
```
  - kind: application-logs
    params:
      logName: "mixer_log"
      logEntryDescriptorNames: ["default"]
```
The above config should select the 1st adapter of kind = `application-logs`


```
  - name: stderr-logs
    kind: application-logs
    impl: stdioLogger
    params:
      logStream: 0 # STDERR
```

```
  - kind: application-logs
    adapter: stderr-logs
    params:
      logName: "mixer_log"
      logEntryDescriptorNames: ["default"]
```
This should specifically select the named adapter `stderr-logs`

pkg/config/validator.go
```diff
@@ -68,10 +66,7 @@ type (
 	// Validated store validated configuration.
 	// It has been validated as internally consistent and correct.
 	Validated struct {
-		// Names are given to specific adapter configurations: unique
 		adapterByName map[string]*pb.Adapter
```
mandarjog: Looks like we do need by `kind` unless we always want to specify one by name.

pkg/aspect/inventory.go
```diff
@@ -14,36 +14,74 @@
...
+	ListsKindName:           ListsKind,
+	MetricsKindName:         MetricsKind,
+	QuotasKindName:          QuotasKind,
+}
+
+// ManagerInventory holds a set of aspect managers.
+type ManagerInventory map[APIMethod][]Manager
+
+// Inventory is the authoritative set of aspect managers used by the mixer.
+var Inventory = ManagerInventory{
```
geeknoid: Why not use global constants? Makes for nicer code at the call site.

Using a function instead makes it so the caller can't change the actual variable, that's true. But since the function returns a reference to some underlying storage, the caller is still free to mutate the state. So unless we'd be copying the data before returning the function, there is no real difference between a function and a global constant.

pkg/config/runtime.go
geeknoid: This doesn't seem like a good idea in general. This makes the order in which adapters are declared to be meaningful, which would be surprising to the operator.

This is also an example of design decisions made to try to make hand-editing of the config state easier by avoiding the need to type in a line of config text.

Let's make this explicit, it has to be specified.

pkg/aspect/inventory.go
geeknoid: Yeah, I was being foolish. Replaced all these exposed vars with functions.

pkg/config/runtime.go
mandarjog: if would like to specify which metrics I want to collect, I use kind=metrics and configure it.

I do not care where these metrics go, SRE has decided that. Why would you force me to care and specify. 

So we need to assign an adapter for use with an aspect when none is specified. It can de that special adapter named 'default' if you prefer.

geeknoid: OK, added support for 'default'.

PTAL.

PR #254, opened 2017-02-10
Implement an aspect manager for the metric aspect.

ZackButcher:
We also touch up some of the code dealing with metrics, renaming LabelKind to LabelType and adding a few utility methods to pkg/adapter/metrics.go.

Still a few open questions denoted by TODOs in the code.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/254)

### DIFF
--- a/adapter/prometheus/prometheus_test.go
+++ b/adapter/prometheus/prometheus_test.go
@@ -44,21 +44,21 @@ var (
 		Name:        "/funky::gauge",
 		Description: "funky all the time",
 		Kind:        adapter.Gauge,
-		Labels:      map[string]adapter.LabelKind{},
+		Labels:      map[string]adapter.LabelType{},
 	}
 
 	counterNoLabels = adapter.MetricDefinition{
 		Name:        "the.counter",
 		Description: "count all the tests",
 		Kind:        adapter.Counter,
-		Labels:      map[string]adapter.LabelKind{},
+		Labels:      map[string]adapter.LabelType{},
 	}
 
 	counter = adapter.MetricDefinition{
 		Name:        "special_counter",
 		Description: "count all the special tests",
 		Kind:        adapter.Counter,
-		Labels: map[string]adapter.LabelKind{
+		Labels: map[string]adapter.LabelType{
 			"bool":   adapter.Bool,
 			"string": adapter.String,
 			"email":  adapter.EmailAddress,
@@ -69,7 +69,7 @@ var (
 		Name:        "unknown",
 		Description: "unknown",
 		Kind:        adapter.Gauge - 2,
-		Labels:      map[string]adapter.LabelKind{},
+		Labels:      map[string]adapter.LabelType{},
 	}
 
 	counterVal = adapter.Value{
@@ -127,7 +127,7 @@ func TestFactory_NewMetricsAspectMetricDefinitionErrors(t *testing.T) {
 		Name:        "/funky::gauge",
 		Description: "funky all the time",
 		Kind:        adapter.Gauge,
-		Labels: map[string]adapter.LabelKind{
+		Labels: map[string]adapter.LabelType{
 			"test": adapter.String,
 		},
 	}
@@ -136,7 +136,7 @@ func TestFactory_NewMetricsAspectMetricDefinitionErrors(t *testing.T) {
 		Name:        "special_counter",
 		Description: "count all the special tests",
 		Kind:        adapter.Counter,
-		Labels: map[string]adapter.LabelKind{
+		Labels: map[string]adapter.LabelType{
 			"email": adapter.EmailAddress,
 		},
 	}
--- a/pkg/adapter/metrics.go
+++ b/pkg/adapter/metrics.go
@@ -17,6 +17,8 @@ package adapter
 import (
 	"errors"
 	"time"
+
+	dpb "istio.io/api/mixer/v1/config/descriptor"
 )
 
 // Metric kinds supported by mixer.
@@ -27,11 +29,12 @@ const (
 
 // Label kinds supported by mixer.
 const (
-	String LabelKind = iota
+	String LabelType = iota
 	Int64
 	Float64
 	Bool
 	Time
+	Duration
 	IPAddress
 	EmailAddress
 	URI
@@ -96,12 +99,12 @@ type (
 		Kind MetricKind
 		// Labels are the names of keys for dimensional data that will
 		// be generated at runtime and passed along with metric values.
-		Labels map[string]LabelKind
+		Labels map[string]LabelType
 	}
 
-	// LabelKind defines the set of known label types that can be generated
+	// LabelType defines the set of known label types that can be generated
 	// by the mixer.
-	LabelKind int
+	LabelType int
 )
 
 // String returns the string-valued metric value for a metrics.Value.
@@ -134,4 +137,44 @@ func (v Value) Float64() (float64, error) {
 		return v, nil
 	}
 	return 0, errors.New("metric value is not a float64")
+}
+
+// FromPbMetricKind translates from MetricDescriptor_MetricKind to Metric Kind.
+func FromPbMetricKind(pbk dpb.MetricDescriptor_MetricKind) MetricKind {
+	switch pbk {
+	case dpb.GAUGE:
+		return Gauge
+	case dpb.COUNTER:
+		return Counter
+	default:
+		return Counter // TODO: unclear what to return in the error case
+	}
+}
+
+// FromPbType translates from ValueType to LabelType.
+func FromPbType(pbk dpb.ValueType) LabelType {
+	switch pbk {
+	case dpb.STRING:
+		return String
+	case dpb.BOOL:
+		return Bool
+	case dpb.INT64:
+		return Int64
+	case dpb.DOUBLE:
+		return Float64
+	case dpb.TIMESTAMP:
+		return Time
+	case dpb.DNS_NAME:
+		return DNSName
+	case dpb.DURATION:
+		return Duration
+	case dpb.EMAIL_ADDRESS:
+		return EmailAddress
+	case dpb.IP_ADDRESS:
+		return IPAddress
+	case dpb.URI:
+		return URI
+	default:
+		return String // TODO: unclear what to return in the error case
+	}
 }
--- a/pkg/adapter/metrics.go
+++ b/pkg/adapter/metrics.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Google Inc.
+// Copyright 2017 The Istio Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/adapter/metrics.go
+++ b/pkg/adapter/metrics.go
@@ -16,6 +16,7 @@ package adapter
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	dpb "istio.io/api/mixer/v1/config/descriptor"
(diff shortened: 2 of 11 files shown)


pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// NewAspect creates a metric aspect.
+func (m *metricsManager) NewAspect(c *config.Combined, a adapter.Builder, env adapter.Env) (Wrapper, error) {
+	params := c.Aspect.Params.(*aconfig.MetricsParams)
+
+	// TODO: get descriptors from config
+	desc := []*dpb.MetricDescriptor{
+		{
+			Name:  "api_responses",
```
douglas-reid: not a blocker for this review, but...

we should pick the three metrics that @mandarjog wants to use for the service graph here.  something like `http_requests`, `errors`, and `total_latencies`.

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,176 @@
...
+			Kind:  dpb.COUNTER,
+			Value: dpb.INT64,
+			Labels: []*dpb.LabelDescriptor{
```
douglas-reid: please add a TODO to sync these schemas with the internal effort to standardize metric schemas. i can loop you in on the docs.

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,176 @@
...
+		out = &Output{Code: code.Code_OK}
+	}
+	return out, result.ErrorOrNil()
+}
+
+func (w *metricsWrapper) Close() error {
+	return w.aspect.Close()
+}
+
+func toDefinition(desc *dpb.MetricDescriptor) adapter.MetricDefinition {
```
douglas-reid: nit: to match the other methods, maybe `definitionFromProto` ?

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+
+		// TODO: investigate either pooling these, or keeping a set around that has only its field's values updated.
+		// we could keep a map[metric name]value, iterate over the it updating only the fields in each value
+		values = append(values, adapter.Value{
+			Name:   name,
+			Kind:   md.metricKind,
+			Labels: labels,
+			// TODO: how do we get times?
+			StartTime:   time.Now(),
```
douglas-reid: at some point, there should be a timestamp attribute that is passed with data. how that gets translated, etc., is an open question.

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,176 @@
...
+func (w *metricsWrapper) Execute(attrs attribute.Bag, mapper expr.Evaluator) (*Output, error) {
+	result := &multierror.Error{}
+	var values []adapter.Value
+
+metadataLoop:
+	for name, md := range w.metadata {
+		metricValue, err := mapper.Eval(md.value, attrs)
+		if err != nil {
+			result = multierror.Append(result, fmt.Errorf("failed to eval metric value for metric '%s' with err: %s", name, err))
+			continue metadataLoop // we can't satisfy this metric because we have the wrong types, so skip to the next
```
douglas-reid: do we need the label for the loop? 

it seems like, here, we could just use an unlabeled continue.

below, if we had a func that was `func mapLabels(expr.Evaluator, map[string]string) (map[string]interface{}, error)`, we could just have a simple:
```
labels, err := mapLabels(mapper, md.labels)
if err != nil {
    continue
}
```

That would be more readable, imho.

Even better if we added a method like: `func (*expr.Evaluator) EvalAll(map[string]string) (map[string]interface{}, error)`, so our code could be:
```
labels, err := mapper.EvalAll(md.labels)
if err != nil {
    continue
}
```

pkg/aspect/metricsManager_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"istio.io/mixer/pkg/adapter"
+	"istio.io/mixer/pkg/adapter/test"
+	aconfig "istio.io/mixer/pkg/aspect/config"
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/config"
+	pb "istio.io/mixer/pkg/config/proto"
+	"istio.io/mixer/pkg/expr"
+)
+
+type fakeBag struct {
```
douglas-reid: why not use the `Bag` and `Eval` in `pkg/aspect/test`?  if those are insufficient, can we flesh them out to meet our needs here?

adapter/prometheus/prometheus_test.go
```diff
@@ -69,7 +69,7 @@ var (
 		Name:        "unknown",
 		Description: "unknown",
 		Kind:        adapter.Gauge - 2,
```
geeknoid: Given we use Kind in config to reference aspects, maybe we should rename this Kind field to Flavor or somesuch?

pkg/aspect/config/metrics.proto
```diff
@@ -0,0 +1,64 @@
...
+    //        api_method: $apiMethod | "unknown"
+    //        # either the attribute named 'responseCode' or the literal int64 500; must eval to an int64
+    //        response_code: $responseCode | 500
+    message Metric {
+        // Must match the name of some metric_descriptor in the global config.
+        string descriptor = 1;
+
+        // Attribute expression to evaluate to determine the value for this metric;
+        // the result of the evaluation must match the value ValueType of the metric_descriptor.
+        string value = 2;
```
geeknoid: I think we want something better than string here.

We can either use type Any, or we can invent our own version of Any that's more constrained. Something like

message AttributeValue {
  oneof value {
    string string_value = 1;
    int64 int64_value = 2;
    ...

adapter/prometheus/prometheus_test.go
ZackButcher: The type of the field is MetricKind, given it lives in a MetricDefinition I think it's overkill to call it MetricKind, but I like Kind more than Flavor. We could call the enum MetricType and this field Type, I suppose.

pkg/aspect/config/metrics.proto
ZackButcher: We chatted offline, to record that here: for now we'll keep stringly typed expressions. In the future we'll address the higher level idea of managers receiving values hydrated via the expression language where the configs can be typed at the proto level. Then, with golang reflection and an extended expression evaluation system we'll be able to populate the strongly typed values correctly. Then the managers won't have to deal with any of this and will be greatly simplified.

pkg/adapter/metrics.go
```diff
@@ -27,11 +30,12 @@ const (
 
 // Label kinds supported by mixer.
 const (
-	String LabelKind = iota
+	String LabelType = iota
```
mandarjog: Are these not a subset of mixer.ValueTypes ?
Do we need to redefine it here?

ZackButcher: They are the same, we can use the proto types directly I suppose. However, I think we want a separation between our API layer and our internal representation of that API for an aspect, which is why this enum exists.

If we want to remove this enum lets do it in a follow-on PR, since it requires updating both of the metric aspects too.

douglas-reid: Agree that I think we will ultimately want a separation between the representations, especially as we add a broader set of adapters.

pkg/aspect/config/metrics.proto
```diff
@@ -0,0 +1,64 @@
...
+    //      descriptor: "response_code" # must match metric_descriptor.name
+    //      value: $requestCount        # Istio expression syntax for the attribute named "requestCount"
+    //      labels:
+    //        # either the attribute named 'apiMethod' or the literal string 'unknown'; must eval to a string
+    //        api_method: $apiMethod | "unknown"
+    //        # either the attribute named 'responseCode' or the literal int64 500; must eval to an int64
+    //        response_code: $responseCode | 500
+    message Metric {
+        // Must match the name of some metric_descriptor in the global config.
+        string descriptor = 1;
```
douglas-reid: i think we probably want to call this `descriptor_name`.

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,227 @@
...
+
+	"github.com/hashicorp/go-multierror"
+	"google.golang.org/genproto/googleapis/rpc/code"
+
+	"istio.io/mixer/pkg/adapter"
+	aconfig "istio.io/mixer/pkg/aspect/config"
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/config"
+	"istio.io/mixer/pkg/expr"
+
```
douglas-reid: nit: eliminate this empty line... `dpb` should be treated the same as `aconfig`

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,227 @@
...
+			Kind:  dpb.COUNTER,
+			Value: dpb.INT64,
+			Labels: []*dpb.LabelDescriptor{
+				{Name: "source", ValueType: dpb.STRING},
+				{Name: "target", ValueType: dpb.STRING},
+				{Name: "service", ValueType: dpb.STRING},
+			},
+		},
+		{
+			Name:  "error_count",
```
douglas-reid: @mandarjog do we have enough in the expression language to suss out error_count directly?  or should we add a label to `request_count` to track response_code (and derive error counts from that)?

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,227 @@
...
+
+	if err := w.aspect.Record(values); err != nil {
+		result = multierror.Append(result, fmt.Errorf("failed to record all values with err: %s", err))
+	}
+
+	// We may accumulate errors while still being able to record some metrics; if we have any values at all we'll
+	// return an OK alongside our errors (since presumably we're able to record at least a few).
+	var out *Output
+	if len(values) > 0 {
+		out = &Output{Code: code.Code_OK}
```
douglas-reid: nit: should we set a default `*Output` with  `Code_OK` and then have this check actually override with something like `Code_INVALID_ARGUMENT` for when `len(values) == 0`?

pkg/aspect/metricsManager.go
```diff
@@ -0,0 +1,227 @@
...
+			desc.Name, desc.Kind, err)
+	}
+	return &adapter.MetricDefinition{
+		Name:   desc.Name,
+		Kind:   kind,
+		Labels: labels,
+	}, nil
+}
+
+func find(defs []*aconfig.MetricsParams_Metric, name string) (*aconfig.MetricsParams_Metric, bool) {
```
douglas-reid: aside: at some point, i imagine we will want a version of this that takes a fn and can be used more broadly across other slices, etc.

pkg/aspect/config/metrics.proto
mandarjog: can we call this `valueExpression` and  `labelExpression` ?
names are useful when people don't  RTFM. 

Another thing to note here is that the expression evaluator is only aware of `ValueTypes`. So it can correctly process and produces the supported value types.

I don't think labelTypes indirection is giving us a whole lot. We will need to map it down multiple times. 
I think it is ok to has   
`func isValidlableType(ValueType) {}` to subset it.

pkg/aspect/metricsManager_test.go
```diff
@@ -0,0 +1,336 @@
...
+	"testing"
+
+	"istio.io/mixer/pkg/adapter"
+	"istio.io/mixer/pkg/adapter/test"
+	aconfig "istio.io/mixer/pkg/aspect/config"
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/config"
+	pb "istio.io/mixer/pkg/config/proto"
+	"istio.io/mixer/pkg/expr"
+
```
douglas-reid: nit: no real reason to have this separate line...

pkg/aspect/metricsManager_test.go
```diff
@@ -0,0 +1,336 @@
...
+	dpb "istio.io/api/mixer/v1/config/descriptor"
+)
+
+type fakeBag struct {
```
douglas-reid: let's consolidate these types, as discussed, in a follow-on CL.

pkg/aspect/metricsManager.go
ZackButcher: I think INVALID_ARGUMENT isn't quite right: the goal here is to see that we passed _something_ to the aspect, and we got some response. The problem is that due to the interface we don't have insight into what an err return from `Record` means, so we can't really tell if we should return an INTERNAL, an OK, or something else. I use the fact that we had any values at all as a proxy for us doing some work here. I think we need a better error story here.

pkg/aspect/metricsManager.go
mandarjog: we do have an attribute `response_code`. so you can do 
`response_code != "200"`  

cannot do `response_code >= 400` yet, but that's just implementation of a Func.

That said, error_count is request_count filtered on a dimension.
It seems we should not make the user write this in the expression language.

pkg/aspect/metricsManager.go
douglas-reid: OK. is there an issue filed to get a better error story (here and elsewhere)?  if not, let's file one and move on.

pkg/aspect/metricsManager.go
douglas-reid: OK. @ZackButcher, let's remove the error_count metric and add a `response_code` field to `request_count`.

pkg/aspect/config/metrics.proto
douglas-reid: defer this for now, based on out-of-band convo.

pkg/aspect/metricsManager.go
ZackButcher: Done

pkg/aspect/metricsManager.go
ZackButcher: Done

pkg/aspect/metricsManager.go
ZackButcher: filed https://github.com/istio/mixer/issues/260

ZackButcher:
PTAL

pkg/adapter/metrics.go
mandarjog: Attributes are *used* in the API layer, but their definitions have wider use.
They are used in the policy language, which is orthogonal to the API.

The expression language is only aware of config.Value_Types, so we will perpetually have to maintain this mapping.

pkg/adapter/metrics.go
```diff
@@ -135,3 +139,43 @@ func (v Value) Float64() (float64, error) {
...
+		return Gauge, nil
+	case dpb.COUNTER:
+		return Counter, nil
+	default:
+		return 0, fmt.Errorf("invalid proto MetricKind %v", pbk)
+	}
+}
+
+// LabelTypeFromProto translates from ValueType to LabelType.
+func LabelTypeFromProto(pbt dpb.ValueType) (LabelType, error) {
```
mandarjog: This could just be lookup in `[]LabelType`

PR #255, opened 2017-02-11
Expression language

mandarjog:
Introducing expression language.

Front end uses golang parser.
Internal AST only has 3 nodes: `variables, constants, functions`
All of them are directly or indirectly typed. All operators are functions. 
`substring(a, 5) == "abc"  --> EQ(substring($a, 5), "abc")`

Supports following types of expressions.  Supports predicates and other expression in a uniform way.

Predicates   returns bool
```
service.name == "*.ns1.cluster" && service.user == "admin"
request.user == "user1"
(x == 20 && y == 10) || x == 30

request.header["X-FORWARDED-HOST"] == "aaa"`
```
Mapping Expressions
```
request.user2| request.user3 | "user1"     --> returns string
`request.size| 200`                                    --> returns int
```

Functions are implemented using `Func` interface
```
type Func interface {
        // Name uniquely identified the function.
        Name() string
        // ReturnType specifies the return type of this function.
        ReturnType() config.ValueType
        // ArgTypes specifies the argument types in order expected by the function.
        ArgTypes() []config.ValueType
        // NullArgs specifies if the function accepts null args
        NullArgs() bool
        // Call performs the function call. It is guaranteed
        // that call will be made with correct types and arity.
        // may panic.
        Call([]interface{}) interface{}
}
```

All expression can be type checked during validation.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/255)

### DIFF
--- a/bin/fmt.sh
+++ b/bin/fmt.sh
@@ -14,12 +14,11 @@ UX=$(uname)
 #remove blank lines so gofmt / goimports can do their job
 for fl in ${GO_FILES}; do
 	if [[ ${UX} == "Darwin" ]];then
-		sed -i '' -e '/^import\s*(/,/)/{ /^\s*$/d;}' $fl
+		sed -i '' -e "/^import[[:space:]]*(/,/)/{ /^\s*$/d;}" $fl
 	else
-		sed -i -e '/^import\s*(/,/)/{ /^\s*$/d;}' $fl
+		sed -i -e "/^import[[:space:]]*(/,/)/{ /^\s*$/d;}" $fl
 	fi
 done
-
 gofmt -s -w ${GO_FILES}
 goimports -w -local istio.io ${GO_FILES}
 buildifier -mode=fix $(find adapter cmd pkg -name BUILD -type f)
--- a/bin/fmt.sh
+++ b/bin/fmt.sh
@@ -3,13 +3,25 @@
 # Applies requisite code formatters to the source tree
 
 set -e
-
 SCRIPTPATH=$( cd "$(dirname "$0")" ; pwd -P )
 ROOTDIR=$SCRIPTPATH/..
 cd $ROOTDIR
 
-gofmt -s -w adapter cmd pkg
-goimports -w adapter cmd pkg
+GO_FILES=$(find adapter cmd pkg -type f -name '*.go')
+
+UX=$(uname)
+
+#remove blank lines so gofmt / goimports can do their job
+for fl in ${GO_FILES}; do
+	if [[ ${UX} == "Darwin" ]];then
+		sed -i '' -e '/^import\s*(/,/)/{ /^\s*$/d;}' $fl
+	else
+		sed -i -e '/^import\s*(/,/)/{ /^\s*$/d;}' $fl
+	fi
+done
+
+gofmt -s -w ${GO_FILES}
+goimports -w -local istio.io ${GO_FILES}
 buildifier -mode=fix $(find adapter cmd pkg -name BUILD -type f)
 buildifier -mode=fix ./BUILD
 buildifier -mode=fix ./BUILD.api
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -134,7 +134,7 @@ func serverOpts(sa *serverArgs, handlers api.Handler) (*api.GRPCServerOptions, e
 
 func runServer(sa *serverArgs) error {
 	// get aspect registry with proper aspect --> api mappings
-	eval := expr.NewCexlEvaluator()
+	eval := expr.NewCEXLEvaluator()
 	adapterMgr := adapterManager.NewManager(adapter.Inventory(), aspect.Inventory(), eval)
 	configManager := config.NewManager(eval, adapterMgr.AspectValidatorFinder(), adapterMgr.BuilderValidatorFinder(),
 		adapterMgr.AdapterToAspectMapperFunc(),
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -134,7 +134,7 @@ func serverOpts(sa *serverArgs, handlers api.Handler) (*api.GRPCServerOptions, e
 
 func runServer(sa *serverArgs) error {
 	// get aspect registry with proper aspect --> api mappings
-	eval := expr.NewIdentityEvaluator()
+	eval := expr.NewCexlEvaluator()
 	adapterMgr := adapterManager.NewManager(adapter.Inventory(), aspect.Inventory(), eval)
 	configManager := config.NewManager(eval, adapterMgr.AspectValidatorFinder(), adapterMgr.BuilderValidatorFinder(),
 		adapterMgr.AdapterToAspectMapperFunc(),
--- a/pkg/attribute/definitionFinder.go
+++ b/pkg/attribute/definitionFinder.go
@@ -1,0 +1,25 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attribute
+
+import (
+	dpb "istio.io/api/mixer/v1/config/descriptor"
+)
+
+// DefinitionFinder finds attribute definitions.
+type DefinitionFinder interface {
+	// FindAttribute finds attribute descriptor in vocabulary. returns nil if not found.
+	FindAttribute(name string) *dpb.AttributeDescriptor
+}
--- a/pkg/attribute/definitionFinder.go
+++ b/pkg/attribute/definitionFinder.go
@@ -18,8 +18,8 @@ import (
 	dpb "istio.io/api/mixer/v1/config/descriptor"
 )
 
-// DefinitionFinder finds attribute definitions.
-type DefinitionFinder interface {
-	// FindAttribute finds attribute descriptor in vocabulary. returns nil if not found.
-	FindAttribute(name string) *dpb.AttributeDescriptor
+// DescriptorFinder finds attribute definitions.
+type DescriptorFinder interface {
+	// FindDescriptor finds attribute descriptor in the vocabulary. returns nil if not found.
+	FindDescriptor(name string) *dpb.AttributeDescriptor
 }
--- a/pkg/config/runtime.go
+++ b/pkg/config/runtime.go
@@ -17,6 +17,7 @@ package config
 import (
 	"github.com/golang/glog"
 	multierror "github.com/hashicorp/go-multierror"
+
 	"istio.io/mixer/pkg/attribute"
 	pb "istio.io/mixer/pkg/config/proto"
 	"istio.io/mixer/pkg/expr"
--- a/pkg/config/runtime.go
+++ b/pkg/config/runtime.go
@@ -70,8 +70,7 @@ func (r *Runtime) Resolve(bag attribute.Bag, aspectSet AspectSet) (dlist []*Comb
 		defer func() { glog.Infof("resolved (err=%v): %s", err, dlist) }()
 	}
 	dlist = make([]*Combined, 0, r.numAspects)
-	err = r.resolveRules(bag, aspectSet, r.serviceConfig.GetRules(), "/", &dlist)
-	return dlist, err
+	return r.resolveRules(bag, aspectSet, r.serviceConfig.GetRules(), "/", dlist)
 }
 
 func (r *Runtime) evalPredicate(selector string, bag attribute.Bag) (bool, error) {
@@ -83,14 +82,14 @@ func (r *Runtime) evalPredicate(selector string, bag attribute.Bag) (bool, error
 }
 
 // resolveRules recurses through the config struct and returns a list of combined aspects
-func (r *Runtime) resolveRules(bag attribute.Bag, aspectSet AspectSet, rules []*pb.AspectRule, path string, dlist *[]*Combined) (err error) {
+func (r *Runtime) resolveRules(bag attribute.Bag, aspectSet AspectSet, rules []*pb.AspectRule, path string, dlist []*Combined) ([]*Combined, error) {
 	var selected bool
 	var lerr error
+	var err error
 
 	for _, rule := range rules {
-		if glog.V(3) {
-			glog.Infof("resolveRules (%v) ==> %v ", rule, path)
-		}
+		glog.V(3).Infof("resolveRules (%v) ==> %v ", rule, path)
+
 		sel := rule.GetSelector()
 		if selected, lerr = r.evalPredicate(sel, bag); lerr != nil {
 			err = multierror.Append(err, lerr)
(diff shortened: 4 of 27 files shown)


pkg/expr/expr.go
```diff
@@ -0,0 +1,365 @@
...
+	default: // string
+		if typedVal, err = strconv.Unquote(v); err != nil {
+			return nil, err
+		}
+	}
+	return &Constant{Value: v, Kind: vType, TypedValue: typedVal}, nil
+}
+
+// Constant models a typed constant.
+type Constant struct {
```
ZackButcher: The field's names feel off here; I think it'd make more sense as:

    struct Constant struct {
        Str String
        Value interface{}
        Type config.ValueType
    }

pkg/expr/expr.go
```diff
@@ -0,0 +1,365 @@
...
+		w.WriteString(arg.String())
+	}
+	w.WriteString(")")
+	return w.String()
+}
+
+// Eval evaluate function.
+func (f *Function) Eval(attrs attribute.Bag, fMap map[string]Func) (interface{}, error) {
+	fn := fMap[f.Name]
+	if fn == nil {
```
ZackButcher: fn, found := fMap[f.Name]
    if !found {
        ...
    }

is more idiomatic

pkg/expr/expr.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil, fmt.Errorf("unknown function: %s", f.Name)
+	}
+	// can throw NPE if config is not consistent.
```
ZackButcher: -> will return an error for nil args if config is not consistent

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+	return res
+}
+
+func (f *eqFunc) call(args []interface{}) bool {
+	var s0 string
+	var s1 string
+	var ok bool
+
+	if s0, ok = args[0].(string); !ok {
+		return args[0] == args[1]
```
ZackButcher: This doesn't work for cases where args are things like `[]byte`; I think you might need to use [`reflect.DeepEqual`](https://golang.org/pkg/reflect/#DeepEqual) here.

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+func (f *baseFunc) NullArgs() bool               { return f.nullArgs }
+
+type eqFunc struct {
+	*baseFunc
+	invert bool
+}
+
+// newEQFunc returns dynamically typed equality fn.
+// treats strings specially with prefix and suffix '*' matches
+func newEQFunc() Func {
```
ZackButcher: nit: I don't think the `Func` suffix is necessary on all of these functions. `newEQ`, `newLOR`, etc. feel fine to me.

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+	*baseFunc
+}
+
+// newIndexFunc returns a map accessor fn.
+func newIndexFunc() Func {
+	return &indexFunc{
+		baseFunc: &baseFunc{
+			name:     "INDEX",
+			retType:  config.VALUE_TYPE_UNSPECIFIED,
+			argTypes: []config.ValueType{config.VALUE_TYPE_UNSPECIFIED, config.STRING},
```
ZackButcher: We should just add a `MAP_...` type to the `ValueType` proto if we're doing this.

pkg/expr/expr.go
```diff
@@ -0,0 +1,365 @@
...
+	return nil
+}
+
+// Parse parses a given expression to ast.Expression.
+func Parse(src string) (ex *Expression, err error) {
+	a, err := parser.ParseExpr(src)
+	if err != nil {
+		return nil, fmt.Errorf("parse error: %s %s", src, err)
+	}
+	glog.V(2).Infof("\n\n%s : %#v\n", src, a)
```
ZackButcher: `"\n\n%s : %#v\n"` => `parsed '%s' into '%#v'`

glog handles newlines for us and the line as written doesn't provide any context

pkg/expr/func.go
mandarjog: good catch.

pkg/expr/func.go
mandarjog: I think it has already been done.

ZackButcher: Shouldn't it be used here then? The first arg must be a map, not an unspecified value.

mandarjog:
It is a subset of golang expressions that is common to other c like languages. (javascript etc.)
The only thing that it forbids is a python like  `in` operator.
If we really need that we will implement a new parser.

The backend will still implement it as `in (needle, haystack ...)`

pkg/expr/func.go
mandarjog: ok, done.

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+	Name() string
+
+	// ReturnType specifies the return type of this function.
+	ReturnType() config.ValueType
+
+	// ArgTypes specifies the argument types in order expected by the function.
+	ArgTypes() []config.ValueType
+
+	// NullArgs specifies if the function accepts null args
+	NullArgs() bool
```
douglas-reid: nit: would `AcceptsNulls` be a more descriptive name?

pkg/expr/expr.go
```diff
@@ -0,0 +1,375 @@
...
+	return false, fmt.Errorf("typeError: got %s, expected bool", reflect.TypeOf(uret).String())
+}
+
+// Validate validates expression for syntactic correctness.
+// TODO check if all functions and attributes in the expression are defined.
+// at present this violates the contract with Func.Call that ensures
+// arity and arg types. It is upto the policy author to write correct policies.
+func (e *cexl) Validate(s string) (err error) {
+	var ex *Expression
+	if ex, err = Parse(s); err != nil {
```
douglas-reid: nit: this might read easier without the named return param...

```
ex, err := Parse(s)
if err != nil {
    return err
}
...
```

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+		return strings.HasSuffix(s0, s1[1:])
+	}
+	return s0 == s1
+}
+
+type lAndFunc struct {
+	*baseFunc
+}
+
+// newLANDFunc returns a binary logical AND fn.
```
douglas-reid: nit: looks like this is now called just `newLAND`

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+func (f *lAndFunc) Call(args []interface{}) interface{} {
+	return args[0].(bool) && args[1].(bool)
+}
+
+type lOrFunc struct {
+	*baseFunc
+}
+
+// newLORFunc returns logical OR fn.
+func newLOR() Func {
```
douglas-reid: nit: just `newLOR` in comment.

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+		&baseFunc{
+			name:     "LOR",
+			retType:  config.BOOL,
+			argTypes: []config.ValueType{config.BOOL, config.BOOL},
+			nullArgs: false,
+		},
+	}
+}
+
+// Call should return thru if at least one element is true
```
douglas-reid: should return `true` ?

pkg/expr/func.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// applies to non bools.
+type orFunc struct {
+	*baseFunc
+}
+
+// newORFunc returns an OR fn that selects first non empty argument.
+// types can be anything, but all args must be of the same type.
+func newOR() Func {
```
douglas-reid: comment -> just `newOR`

pkg/expr/eval_test.go
```diff
@@ -0,0 +1,375 @@
...
+			},
+			true, "unresolved attribute",
+		},
+	}
+
+	for idx, tst := range tests {
+		attrs := &bag{attrs: tst.tmap}
+		exp, err := Parse(tst.src)
+		if err != nil {
+			t.Errorf("[%d] unexpected error: %s", idx, err)
```
mandarjog: good point, done.
It seems to have caught up with Gomega tests.

pkg/expr/func.go
mandarjog: done

pkg/expr/func.go
mandarjog: one.

pkg/expr/func.go
mandarjog: done

pkg/expr/func.go
mandarjog: done.

mandarjog:
PTAL.
Expression language is in and type validation is also implemented.

pkg/expr/expr_test.go
```diff
@@ -0,0 +1,195 @@
...
+	config "istio.io/api/mixer/v1/config/descriptor"
+	dpb "istio.io/api/mixer/v1/config/descriptor"
+)
+
+func TestGoodParse(tst *testing.T) {
+	tests := []struct {
+		src     string
+		postfix string
+	}{
+		{"a.b.c == 2", "EQ($a.b.c, 2)"},
```
wora: I prefer @a.b.c because @ is close to attribute than $.

pkg/expr/expr_test.go
```diff
@@ -0,0 +1,195 @@
...
+		{`!(a.b || b || "c a b" || ( a && b ))`, `NOT(LOR(LOR(LOR($a.b, $b), "c a b"), LAND($a, $b)))`},
```
wora: I wonder if we can use operators directly, such as `a + b` => `+(a, b)`. In this way, we can allow more operators without having to define a large grammar.

bin/fmt.sh
```diff
@@ -3,13 +3,24 @@
...
 
 set -e
-
 SCRIPTPATH=$( cd "$(dirname "$0")" ; pwd -P )
 ROOTDIR=$SCRIPTPATH/..
 cd $ROOTDIR
 
-gofmt -s -w adapter cmd pkg
-goimports -w adapter cmd pkg
+GO_FILES=$(find adapter cmd pkg -type f -name '*.go')
```
douglas-reid: nit: moving forward, let's separate out unrelated changes into distinct PRs (keep them small, reduce load, etc.)

cmd/server/server.go
```diff
@@ -134,7 +134,7 @@ func serverOpts(sa *serverArgs, handlers api.Handler) (*api.GRPCServerOptions, e
 
 func runServer(sa *serverArgs) error {
 	// get aspect registry with proper aspect --> api mappings
-	eval := expr.NewIdentityEvaluator()
+	eval := expr.NewCexlEvaluator()
```
douglas-reid: is Cexl an initialism?  If so, should this be `NewCEXLEvaluator` ?

pkg/attribute/definitionFinder.go
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+package attribute
+
+import (
+	dpb "istio.io/api/mixer/v1/config/descriptor"
+)
+
+// DefinitionFinder finds attribute definitions.
+type DefinitionFinder interface {
```
douglas-reid: Since the method is `FindAttribute`, should this be `AttributeFinder` ? 

Would `DescriptorFinder` and `FindDescriptor` perhaps be more appropriate?

pkg/config/runtime.go
```diff
@@ -87,7 +88,7 @@ func (r *Runtime) resolveRules(bag attribute.Bag, aspectSet AspectSet, rules []*
 	var lerr error
 
 	for _, rule := range rules {
-		if glog.V(2) {
+		if glog.V(3) {
```
douglas-reid: curious: why do we use this form here, and the inline form below?  is rule that much of a pain to stringify in comparison?

pkg/expr/eval_test.go
douglas-reid: just as an FYI: I don't see a change here to the tests. not a blocker, but I'm wondering if there is a missing commit ?

pkg/expr/eval_test.go
```diff
@@ -0,0 +1,375 @@
...
+			continue
+		}
+		if res != tst.result {
+			t.Errorf("[%d] got %s\nwant %s", idx, res, tst.result)
+		}
+	}
+
+}
+
+func TestCexlEval(t *testing.T) {
```
douglas-reid: testCEXLEval ?

pkg/expr/eval_test.go
```diff
@@ -0,0 +1,375 @@
...
+			continue
+		}
+		// check if error is of the correct type
+		if err != nil && !strings.Contains(err.Error(), tt.err) {
+			t.Errorf("[%d] got %s, want %s", idx, err, tt.err)
+		}
+	}
+}
+
+// fake bag
```
douglas-reid: aside: i wonder if we should create an pkg/attribute/bagtest.go or something. we have several test impls, including pkg/aspect/test/bag.go throughout the codebase that we should try to harmonize.

pkg/expr/func.go
```diff
@@ -0,0 +1,239 @@
...
+package expr
+
+import (
+	"reflect"
+	"strings"
+
+	config "istio.io/api/mixer/v1/config/descriptor"
+)
+
+// Func every expression function provider should have this interface.
```
douglas-reid: nit: make this a complete sentence. something like: `Func defines the interface that every expression function provider must implement`.

testdata/globalconfig.yml
```diff
@@ -6,12 +6,8 @@ adapters:
     impl: memQuota
     params:
   - name: default
-    kind: access-logs
```
douglas-reid: are these changes related to the previous refactors?

pkg/expr/eval_test.go
mandarjog: I just used it in new tests in expr_test.go

pkg/attribute/definitionFinder.go
mandarjog: fixed.

cmd/server/server.go
mandarjog: done.

bin/fmt.sh
mandarjog: 👍

pkg/config/runtime.go
mandarjog: I think at one point there was another statement inside glog.V(2). Changes to inline form now.

pkg/expr/eval_test.go
mandarjog: done.

pkg/expr/eval_test.go
mandarjog: https://github.com/istio/mixer/issues/263

pkg/expr/expr_test.go
mandarjog: function names have to be valid identifiers. So that excludes any operators as function names. 
I would like to minimize number of operators beyond the customary ones that we must support.

pkg/expr/func.go
mandarjog: done.

testdata/globalconfig.yml
mandarjog: Yes. `kind` is specified by the adapter implementation now. It is no longer needed  (allowed) in the global config.

mandarjog:
PTAL.

pkg/expr/func.go
```diff
@@ -0,0 +1,239 @@
...
+	// Name uniquely identified the function.
+	Name() string
+
+	// ReturnType specifies the return type of this function.
+	ReturnType() config.ValueType
+
+	// ArgTypes specifies the argument types in order expected by the function.
+	ArgTypes() []config.ValueType
+
+	// NullArgs specifies if the function accepts null args
```
douglas-reid: nit: comment no longer matches method name.

PR #257, opened 2017-02-11
Add example servicegraph service.

douglas-reid:
This PR adds an experimental servicegraph generation service to the
mixer codebase. This addition is meant to be temporary, as it is
expected that such a service will eventually be moved to a proper
repository once it evolves beyond initial prototype status.

A demosvc is also provided a target for prometheus scraping (to gen data
for testing).

No tests are provided, at the moment, as this is prototype (at best)
quality code.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/257)


mandarjog:
This addresses issue #257

examples/servicegraph/cmd/server/BUILD
```diff
@@ -0,0 +1,21 @@
...
+    deps = [
+        "//examples/servicegraph:go_default_library",
+        "//examples/servicegraph/dot:go_default_library",
+        "//examples/servicegraph/promgen:go_default_library",
+    ],
+)
+
+go_binary(
+    name = "servicegraph",
+    data = ["//examples/servicegraph:js/viz-lite.js"],
```
mandarjog: Can we make this a dependency in the WORKSPACE? do we need to check it in ?

douglas-reid: possibly. it isn't really needed anywhere in general (exists entirely for visual debugging purposes in early days). when we settle on the right place to put everything, we can just delete it entirely.

douglas-reid:
Agreed re: standard cmd codebase / execution. Just want to keep things as minimal as possible during initial impl.

Issue #256, opened 2017-02-11
Regression [panic: runtime error:] - Fix adapterByName semantics: 


mandarjog:
pr #253   changed the way adapters lookup is handled.

At present, `validated.adapterByName` is keyed on `adapterName`
1. It should be keyed off `Impl` on config validation path
2. It should be keyed off `kind/name` on config runtime path

original problem
```
I0211 05:20:27.269363   61556 runtime.go:78] resolveRules (aspects:<kind:"quotas" params:&QuotasParams{QuotaDescriptorNames:[],} > aspects:<kind:"access-logs" params:&AccessLogsParams{LogName:access_log,Attributes:[],LogFormat:COMMON,CustomLogTemplate:,} > aspects:<kind:"application-logs" params:&ApplicationLogsParams{LogName:mixer_log,LogEntryDescriptorNames:[default],SeverityAttribute:,TimestampAttribute:,TimestampFormat:2006-01-02T15:04:05Z07:00,} > ) ==> /
I0211 05:20:27.269545   61556 runtime.go:108] Aspect rejected: access-logs not in set [map[denials:true lists:true quotas:true]]
I0211 05:20:27.269558   61556 runtime.go:108] Aspect rejected: application-logs not in set [map[denials:true lists:true quotas:true]]
I0211 05:20:27.269565   61556 handler.go:106] Resolved [1] ==> [0xc4201905f0]
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xb7ed3]

goroutine 11 [running]:
panic(0x5799e0, 0xc4200180a0)
	/Volumes/Untitled/BAZEL_OUT/execroot/mixer/external/io_bazel_rules_go_toolchain/src/runtime/panic.go:500 +0x1a1
istio.io/mixer/pkg/adapterManager.(*Manager).execute(0xc420015e00, 0x8c3540, 0xc420194f00, 0xc4201905f0, 0x8c74c0, 0xc4202e8300, 0x0, 0x0, 0x0)
	/Volumes/Untitled/BAZEL_OUT/execroot/mixer/bazel-out/local-fastbuild/bin/pkg/adapterManager/go_default_library.a.dir/istio.io/mixer/pkg/adapterManager/manager.go:139 +0x123
[...]

Issue #204, opened 2017-01-27
Support for single adapter object providing multiple builders 


mandarjog:
Add support for single Adapter providing multiple Builders.

This involves.
1. Removing global adapterName uniqness. (adapterName, aspectKind) must be unique.
2. Removing "Kind" property from Adapter proto.
    This enables an adapter to assert multiple "Kinds" in the code.

This enables a GCP Adapter to provide builders for multiple GCP Aspects.

The above changes enable code like the following.

`adapter/stdioLogger/stdioLogger.go`

would like to register - 
```
b := builder{adapter.NewDefaultBuilder(name, desc, conf)}
 r.RegisterLogger(b)
 r.RegisterAccessLogger(b)
```
instead of 
```
	r.RegisterLogger(builder{adapter.NewDefaultBuilder(
 		"istio/stdioLogger",
 		"Writes structured log entries to a standard I/O stream",
 		&config.Params{},
 	)})
 	r.RegisterAccessLogger(builder{adapter.NewDefaultBuilder(
 		"istio/stdioAccessLogger",
 		"Writes structured access log entries to a standard I/O stream",
 		&config.Params{},
 	)})
```

mandarjog:
After this issue is resolved update `adapter/stdioLogger/stdioLogger.go#Register()`

geeknoid:
Done.

mandarjog:
Keeping it open for.
Removing "Kind" property from Adapter proto.

geeknoid:
Addressed by new adapter model.

PR #258, opened 2017-02-12
Fix Regression: semantics of single adapter -> multiple aspects

mandarjog:
fixed issue #256
fixes issue #204

add AdapterToAspectMapperFunc which is analogous to ValidatorFinderFunc
AdapterToAspectMapperFunc returns a list of aspects provided by a given
adapter.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/258)


cmd/server/inventory.go
```diff
@@ -78,32 +78,19 @@ func listAspects() error {
...
-		sort.Strings(keys)
+	sort.Strings(keys)
+	for _, impl := range keys {
+		b := builderMap[impl].Builder
 
-		for _, impl := range keys {
-			b := m[impl]
-			fmt.Printf("adapter %s/%s: %s\n", kind, impl, b.Description())
-			printConfigValidator(b)
+		fmt.Printf("adapter %s: %s\n", impl, b.Description())
```
douglas-reid: should we be using glog here?

pkg/adapterManager/registry.go
```diff
@@ -93,18 +106,31 @@ func (r *registry) RegisterMetricsBuilder(b adapter.MetricsBuilder) {
 	r.insert(aspect.MetricsKind, b)
 }
 
-func (r *registry) insert(kind aspect.Kind, b adapter.Builder) {
-	glog.V(2).Infof("Registering %v:%v", kind, b)
+func (r *registry) insert(k aspect.Kind, b adapter.Builder) {
+	kind := k.String()
+	if glog.V(1) {
+		defer glog.Infof("Registered %s / %s", kind, b.Name())
```
douglas-reid: won't this still print "Registered" even if there is a panic? do we care?

pkg/config/runtime.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/golang/glog"
 	multierror "github.com/hashicorp/go-multierror"
 
+	"bytes"
```
douglas-reid: code cleanup issue with imports (everyone's favorite!).

pkg/config/runtime.go
```diff
@@ -42,6 +44,19 @@ type (
 	AspectSet map[string]bool
 )
 
+func (c *Combined) String() string {
+	var b bytes.Buffer
+	if c.Builder != nil {
+		b.WriteString("builder: ")
```
douglas-reid: aside: how is this not a linter issue (not dealing with the errors) ?

cmd/server/inventory.go
mandarjog: Since this is an actual command, I think we *are* using `fmt.Printf` here.

pkg/adapterManager/registry.go
mandarjog: done.

pkg/config/runtime.go
mandarjog: fixed.

cmd/server/inventory.go
douglas-reid: ah. ok. good catch.

pkg/adapterManager/registry_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"testing"
 
+	"reflect"
```
douglas-reid: nit: clean up imports.

PR #261, opened 2017-02-14
Add ability to add 'static' nodes to the servicegraph.

douglas-reid:
There may be a need to add nodes directly, rather depend on them
being discovered through metrics data.

A subsequent PR will address renaming the bits from {static, dynamic}
to {declared, inferred} and updating behavior to match the evolving
design doc.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/261)


examples/servicegraph/cmd/demosvc/main.go
```diff
@@ -30,15 +31,28 @@ var requests = prometheus.NewCounterVec(
 		Name: "request_count",
 		Help: "requests",
 	},
-	[]string{"source", "target", "service"},
+	[]string{"source", "target", "service", "response_code"},
 )
 
+var responseCodes = []int{
```
mandarjog: looks like this is implementing weights.
A comment would be good here.

examples/servicegraph/cmd/server/main.go
```diff
@@ -47,17 +47,45 @@ func (fs justFilesFilesystem) Open(name string) (http.File, error) {
...
+	nodeName := r.URL.Query().Get("name")
+	if nodeName == "" {
+		w.WriteHeader(http.StatusBadRequest)
+		_, err := w.Write([]byte("missing argument 'name'"))
+		if err != nil {
+			log.Print(err)
+		}
+		return
+	}
+	s.staticGraph.Nodes[nodeName] = struct{}{}
```
mandarjog: is nodeName sufficient without any metadata ?

douglas-reid: Right now, nodes don't have any associated metadata.  That can be added later -- at which point this will need to be updated.

examples/servicegraph/cmd/demosvc/main.go
douglas-reid: Yup. Will do.

douglas-reid:
PTAL.

Added comment on response codes list.

PR #262, opened 2017-02-14
Disable the precommits during rebases

ZackButcher:
From https://github.com/nlf/git-validate/issues/14#issuecomment-62408961

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/262)

### DIFF
--- a/bin/pre-commit
+++ b/bin/pre-commit
@@ -8,6 +8,7 @@ WD=$(dirname $0)
 WD=$(cd $WD; pwd)
 ROOT=$(dirname $WD)
 
+
 # Not running under git commit
 # install handler
 if [[ -z ${GIT_AUTHOR_DATE} ]];then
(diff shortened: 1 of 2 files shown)


bin/pre-commit
```diff
@@ -24,14 +24,17 @@ if [[ -f WORKSPACE ]]; then
...
+    cd $ROOT
 
-# While adding more tests here keep in mind that
-# this runs every time a user commits
-# Run all the formatters now
-bin/fmt.sh
+    # While adding more tests here keep in mind that
+    # this runs every time a user commits
+    # Run all the formatters now
+    bin/fmt.sh
```
mandarjog: now that you are touching this file :-- 
we should *only* run bin/fmt.sh  on pre-commit

bazel test //...
bin/linters.sh
bin/racetest.sh 

should all be put in the  pre-push hook.

ZackButcher:
Split things up so that formatting only happens pre-commit and all tests + the linters run pre-push. You might have insight into this, but apparently the pre-push script's working dir is `ISTIO_ROOT/.git`, not `ISTIO_ROOT` like the pre-commit. Not sure why that is, but the scripts are slightly different as a result.

PR #269, opened 2017-02-22
Fix ConvertParams to use proto names instead of go

mandarjog:
Switch from mapstruct to
yml -> obj -> json -> proto
using gogo/protobuf/jsonpb

Now yaml file should use names from proto definition.

replace yaml.v2 with ghodss_yaml because ghodss_yaml processes yaml in a json compatible way.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/269)


pkg/config/validator.go
```diff
@@ -222,42 +224,33 @@ func ConvertParams(find ValidatorFinderFunc, name string, params interface{}, st
...
-		return err
-	}
-
-	if err = d.Decode(src); err != nil {
+// if strict is true returns error on unknown fields.
+func Decode(src interface{}, dst adapter.AspectConfig, strict bool) (err error) {
+	var ba []byte
+	// src is potentially map[interface{}] interface{}
+	// unmarshaled by yaml. json expects map[string] interface{}
+	ba, err = json.Marshal(convert(src))
```
kyessenov: or you can just pass strings from YAMLtoJSON to jsonpb since you serialize keys back to string anyways.

kyessenov: e.g. https://github.com/istio/manager/blob/master/cmd/manager/config.go#L69

PR #271, opened 2017-02-22
Add support for STRING_MAP attributes.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/271)

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -47,6 +47,9 @@ type rootArgs struct {
 	// bytesAttributes is the list of name/value pairs of bytes attributes that will be sent with requests.
 	bytesAttributes string
 
+	// stringMapAttributes is the list of string maps that will be sent with requests
+	stringMapAttributes string
+
 	// mixerAddress is the full address (including port) of a mixer instance to call.
 	mixerAddress string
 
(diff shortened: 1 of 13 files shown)


mandarjog:
@qiwzhang This PR adds support to send attribute of type `StringMap` `map<string,string>` which we expect request.headers / response.headers to be.
The mixer client code will need to support this data type.

qiwzhang:
@mandarjog I will change the client to use it.

cmd/client/main.go
```diff
@@ -47,6 +47,9 @@ type rootArgs struct {
 	// bytesAttributes is the list of name/value pairs of bytes attributes that will be sent with requests.
 	bytesAttributes string
 
+	// stringMapAttributes is the list of string maps that will be sent with requests
+	stringMapAttributes string ``
```
geeknoid: Not sure where that came from... Removed.

cmd/client/util.go
```diff
@@ -99,6 +99,38 @@ func parseBytes(s string) (interface{}, error) {
...
+
+		k := pair[0:colon]
+		v := pair[colon+1:]
+		m[k] = v
+	}
+	return m, nil
+}
+
+// add to dictionary
+func add2Dict(dictionary map[int32]string, s string) int32 {
```
geeknoid: Given the context in which this is used, I didn't care about the redundancy in that case. But just to be anal, I added a little loop to discover and reuse existing dictionary entries.

cmd/client/util_test.go
```diff
@@ -19,110 +19,137 @@ import (
...
+		{"v", 3.0},
+		{"w", true},
+		{"x", time.Date(2006, 1, 2, 15, 4, 5, 0, time.UTC)},
+		{"y", time.Duration(42) * time.Second},
+		{"z", []byte{0x98, 0x76}},
+		{"zz", map[string]string{"k3": "v3"}},
+	}
+
+	for _, r := range results {
+		v, found := attribute.Value(b, r.name)
```
geeknoid: I switched to t.Run in a few tests. Thanks.

geeknoid:
PTAL.

PR #274, opened 2017-02-22
Introduce basic k8s configs for the mixer, prometheus, and grafana

ZackButcher:
We add configs for running the mixer via k8s, as well as prometheus and grafana. These have been tests e2e together. We also add a README that describes running these configs and how to interact with the deployment.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/274)

### DIFF
--- a/deploy/kube/README.md
+++ b/deploy/kube/README.md
@@ -3,7 +3,7 @@
 These configurations stand up an instance of the mixer, a Prometheus instance that scrapes metrics the mixer exposes, and
 a Grafana instance to render those metrics. These are intended for development and local testing, not a real production
 deployment. The easiest way to stand up these deployments is to run (from the Istio mixer root directory):
-```bash
+```shell
 $ kubectl create configmap prometheus-config --from-file=testdata/prometheus.yaml
 $ kubectl create configmap mixer-config --from-file=testdata/globalconfig.yml --from-file=testdata/serviceconfig.yml
 $ kubectl apply -f ./deploy/kube/
@@ -12,7 +12,7 @@ $ kubectl apply -f ./deploy/kube/
 If you're using something like [minikube](https://github.com/kubernetes/minikube) to test locally, you can call the
 mixer server with our test client (`mixc`), providing some attribute values:
 
-```bash
+```shell
 $ MIXS=$(minikube service mixer --url --format "{{.IP}}:{{.Port}}" | head -n 1)
 $ bazel run //cmd/client:mixc -- check -m $MIXS -a source.name=source,target.name=target,api.name=myapi,response.code=200,response.latency=100,client.id=$USER
 ```
@@ -30,12 +30,12 @@ configurations are expected to be mounted in two files at `/etc/opt/mixer/`. We 
 provide these configurations. Usually this configmap is created from the `//testdata/` directory by:
 
 <a name="configmap_command"></a>
-```bash
+```shell
 $ kubectl create configmap mixer-config --from-file=testdata/globalconfig.yml --from-file=testdata/serviceconfig.yml
 ```
 A file can be created to capture this configmap by running:
 
-```bash
+```shell
 $ kubectl get configmap mixer-config -o yaml > ./mixer-config.yaml
 ```
 We do not check in this file because we consider the yaml files in `//testdata` the source of truth for this config.
@@ -49,7 +49,7 @@ TODO: create a sample grafana config and link it here
 
 If you're using minikube to test these deployments locally, get to the UI by running:
 
-```bash
+```shell
 $ minikube service grafana
 ```
 
@@ -61,7 +61,7 @@ source is at `http://prometheus:9090/` (no auth required, access via proxy).
 works with these deployments is stored in `//testdata/`. Prometheus expects its config to be mounted at
 `/etc/opt/mixer/prometheus.yaml`. We use a configmap named `prometheus-config` to provide these configurations. Usually
 this configmap is created from the `//testdata` directory by:
-```bash
+```shell
 $ kubectl create configmap prometheus-config --from-file=testdata/prometheus.yaml
 ```
 
@@ -108,20 +108,20 @@ the mixer server by allowing a developer to push docker images locally rather th
 registry (`kubectl apply -f ./deploy/kube/localregistry.yaml`), and update `mixer.yaml` to use the image
 `localhost:5000/mixer`. After the registry server is running, expose it locally by executing:
 
-```bash
+```shell
 $ kubectl port-forward --namespace kube-system $POD 5000:5000
 ```
 
 If you're testing locally with minikube, `$POD` can be set with:
 
-```bash
+```shell
 $ POD=$(kubectl get pods --namespace kube-system -l k8s-app=kube-registry \
   -o template --template '{{range .items}}{{.metadata.name}} {{.status.phase}}{{"\n"}}{{end}}' \
   | grep Running | head -1 | cut -f1 -d' ')
 ```
 
 Then you can build the mixer, create a docker image, and push it to the local registry, by running:
-```bash
+```shell
 $ bazel build ...:all
 $ bazel run //docker:mixer localhost:5000/mixer
 $ docker push localhost:5000/mixer
--- a/deploy/kube/README.md
+++ b/deploy/kube/README.md
@@ -1,8 +1,8 @@
 # Istio Mixer K8s Configuration
 
-Intended for our demo, these configurations stand up an instance of the mixer, a Prometheus instance that scapes metrics
-the mixer exposes, and a Grafana instance to render those metrics. The easiest way to stand up these deployments is to
-run (from the Istio mixer root directory):
+These configurations stand up an instance of the mixer, a Prometheus instance that scapes metrics the mixer exposes, and
+a Grafana instance to render those metrics. These are intended for a demo, not a real production setup. The easiest way
+to stand up these deployments is to run (from the Istio mixer root directory):
 
     $ kubectl create configmap testdata-config --from-file=testdata/globalconfig.yml --from-file=testdata/serviceconfig.yml --from-file=testdata/prometheus.yaml
     $ kubectl apply -f ./deploy/kube/
--- a/deploy/kube/README.md
+++ b/deploy/kube/README.md
@@ -1,77 +1,128 @@
 # Istio Mixer K8s Configuration
 
-These configurations stand up an instance of the mixer, a Prometheus instance that scapes metrics the mixer exposes, and
-a Grafana instance to render those metrics. These are intended for a demo, not a real production setup. The easiest way
-to stand up these deployments is to run (from the Istio mixer root directory):
-
-    $ kubectl create configmap testdata-config --from-file=testdata/globalconfig.yml --from-file=testdata/serviceconfig.yml --from-file=testdata/prometheus.yaml
-    $ kubectl apply -f ./deploy/kube/
+These configurations stand up an instance of the mixer, a Prometheus instance that scrapes metrics the mixer exposes, and
+a Grafana instance to render those metrics. These are intended for development and local testing, not a real production
+deployment. The easiest way to stand up these deployments is to run (from the Istio mixer root directory):
+```bash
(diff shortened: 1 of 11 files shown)


deploy/kube/README.md
```diff
@@ -0,0 +1,77 @@
+# Istio Mixer K8s Configuration
+
+These configurations stand up an instance of the mixer, a Prometheus instance that scapes metrics the mixer exposes, and
```
kyessenov: scape -> scrape

deploy/kube/README.md
```diff
@@ -0,0 +1,77 @@
...
+to stand up these deployments is to run (from the Istio mixer root directory):
+
+    $ kubectl create configmap testdata-config --from-file=testdata/globalconfig.yml --from-file=testdata/serviceconfig.yml --from-file=testdata/prometheus.yaml
+    $ kubectl apply -f ./deploy/kube/
+    
+If you're using something like [minikube](https://github.com/kubernetes/minikube) to test locally, you can call the
+mixer server with our test client (`mixc`), providing some attribute values:
+
+    $ MIXS=$(minikube service mixer --url --format "{{.IP}}:{{.Port}}")
+    $ bazel run //cmd/client:mixc -- check -m $MIXS -a source.name=source,target.name=target,api.name=myapi,response.code=200,response.latency=100,client.id=$USER
```
kyessenov: no bazel stuff in the demo, please. maybe you should distribute your client as a docker image?

deploy/kube/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## mixer.yaml
+`mixer.yaml` describes a deployment and service for the mixer server binary. The deployment annotates the pods it
+creates with a `prometheus_port` annotation which is required by our Prometheus deployment (see [prometheus.yaml](#prometheus)
+below as well as the prometheus configuration in `//testdata/prometheus.yaml`).
+
+Two pieces of configuration are required to run the mixer: a global configuration and a service configuration. Example
+configurations can be found in the `//testdata` directory (`//` indicates the root of the Istio Mixer project directory).
```
kyessenov: put a link

deploy/kube/README.md
```diff
@@ -0,0 +1,77 @@
...
+access via proxy).
+
+
+## <a name="prometheus"></a> prometheus.yaml
+`prometheus.yaml` describes a deployment and service for a Prometheus collection server. Prometheus expects its config to
+be mounted at `/etc/opt/mixer/prometheus.yaml`. An example configuration that works with these deployments is stored in
+`//testdata/`. For simplicity we shove this into the same configmap the mixer serves from, and the [configmap command above](#configmap_command)
+includes the Prometheus configuration. 
+
+When the mixer is configured with the `prometheus` adapter it hosts metrics on an HTTP server at port `42422` (this can
```
kyessenov: the port numbers I think can be removed from the demo overview.

deploy/kube/README.md
```diff
@@ -0,0 +1,77 @@
...
+be changed in the Prometheus adapter configuration). The `mixer.yaml` deployment annotates all mixer pods with this port
+number, and the Prometheus configuration at `//testdata/prometheus.yaml` uses it to compute the URL to scrape for metrics.
+
+## localregistry.yaml
+`localregistry.yaml` is a copy of Kubernete's local registry configuration, and is included to make it easier to test
```
kyessenov: this is rather strange: why store kubectl get services as a yaml file?

deploy/kube/grafana.yaml
```diff
@@ -0,0 +1,46 @@
...
+        emptyDir: {}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: grafana
+spec:
+  type: NodePort
+  ports:
+  - name: "grafana"
```
kyessenov: we use a pattern to prefix port name with protocol. if it's http, then call it "http"

deploy/kube/localregistry.yaml
```diff
@@ -0,0 +1,88 @@
+# after creating the local registry, to push images to it you need to open up a port to the pod's registry:
```
kyessenov: I don't understand what this service is doing? is it listing services?

deploy/kube/mixer.yaml
```diff
@@ -0,0 +1,41 @@
...
+metadata:
+  name: mixer
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: mixer
+      annotations:
+        prometheus_port: "42422"
```
kyessenov: why not just stick as an arg to mixs?

deploy/kube/mixer.yaml
```diff
@@ -0,0 +1,41 @@
...
+          name: testdata-config
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: mixer
+spec:
+  type: NodePort
+  ports:
+  - name: "api"
```
kyessenov: call it grpc

deploy/kube/prometheus.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - name: prometheus
+        image: prom/prometheus
+        volumeMounts:
+          - mountPath: /prometheus
+            name: data-prom
+          - mountPath: /etc/opt/mixer
+            name: config
+        ports:
+          - containerPort: 9090
+        args: [
```
kyessenov: use array YAML syntax (drop [, ], use dashes)

deploy/kube/localregistry.yaml
mandarjog: This looks like a private docker registry. Is this really needed for the demo ?
I think it detracts from the rest of the flow.

testdata/serviceconfig.yml
```diff
@@ -7,6 +7,23 @@ rules:
   aspects:
   - kind: quotas
     params:
+  - kind: metrics
+    adapter: "prometheus"
+    params:
+      metrics:
+      - descriptor: "request_count"
+        value: "1"
```
mandarjog: if you process value thru the expression language, this should be `1` not `"1"`

deploy/kube/README.md
douglas-reid: this isn't really a demo script, so much as a test deployment.  sending requests directly to mixer is a good test.  i'm not sure how we avoid building the client however.

do you think we should just build the client as part of the same image and use `docker run` scripts to execute the client?  would that be better?

deploy/kube/README.md
```diff
@@ -0,0 +1,77 @@
...
+A file can be created to capture this configmap by running:
+
+    $ kubectl get configmap testdata-config -o yaml > ./testdata-config.yaml
+
+We do not check in this file since we consider the yaml files in `//testdata` the source of truth for this config.
+
+## grafana.yaml
+`grafana.yaml` describes a deployment and service for Grafana, based on the [`grafana/grafana` docker image](https://hub.docker.com/r/grafana/grafana/).
+The Grafana UI is exposed on port `3000`. This configuration stores Grafana's data in the `/data/grafana` dir, which we
+configure to be an ephemeral directory. To persist your Grafana configuration for longer than the lifetime of the `grafana`
```
douglas-reid: you can also export the config for grafana and reimport it. that might be worth calling out here (and we should probably link to a demo config).

deploy/kube/localregistry.yaml
douglas-reid: yeah, i think we can get by with relying on docker for this.

deploy/kube/mixer.yaml
douglas-reid: i think this is using black-magic to get prom to scrape at the right addr:port.

we don't have a good way of ensuring the .yaml matches the mixer adapter configs, etc., yet., so i think this is a reasonable first step.  perhaps we should comment on it here, however?

testdata/serviceconfig.yml
ZackButcher: The `value` field is a string expression, it just so happens that the expression I want evaluated is the literal `1`. I can make it look like `1` since yaml doesn't require quotes for strings, but in reality it's still a string expression that is evaluated to the integer 1.

I agree it's odd, but it's what we want to do: for each (request, service, target, response code) set we want to increment the counter by 1. I'll add comments explaining.

deploy/kube/mixer.yaml
andraxylia: Using an extra annotation is not the correct way to discover the port. Correct way is to expose the 42422 port in the mixer service, then use standard prometheus meta labels in the prometheus job:

"The service role discovers a target for each service port for each service. This is generally useful for blackbox monitoring of a service. The address will be set to the Kubernetes DNS name of the service and respective service port.

Available meta labels:

__meta_kubernetes_namespace: The namespace of the service object.
__meta_kubernetes_service_name: The name of the service object.
__meta_kubernetes_service_label_<labelname>: The label of the service object.
__meta_kubernetes_service_annotation_<annotationname>: The annotation of the service object.
__meta_kubernetes_service_port_name: Name of the service port for the target.
__meta_kubernetes_service_port_number: Number of the service port for the target.
__meta_kubernetes_service_port_portocol: Protocol of the service port for the target.
"

ZackButcher: The prometheus.yaml included in this PR is using meta labels:

````
   kubernetes_sd_configs:
    - role: pod
   relabel_configs:
    # we read the pod annotations "prometheus_port" from kubernetes, and concat it with the address
    - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_port]
      action: replace
      regex: ([^:]+)(?::\d+)?;(\d+)
      replacement: $1:$2
      target_label: __address__
````

I can swap over to the `service` role and rework the relabeling.

deploy/kube/localregistry.yaml
ZackButcher: We definitely can rely on docker hub/gcr - the mixer.yaml in this PR uses gcr. I was using this for my own testing so I didn't have to push random images to dockerhub with my own code changes. I think it's worth keeping around for local debugging but I can remove it if we don't agree.

testdata/serviceconfig.yml
mandarjog: From the expression languages standpoint `"2"` will be evaluated to a string and `2` will be evaluated to an `int64` and looks like for correct operation you __want__ it to be `int64`

kyessenov: What does YAML expect here?

deploy/kube/README.md
ZackButcher: We need a whole separate doc about configuring adapters. I'll add in some config snippets here too.

deploy/kube/README.md
kyessenov: So this is _docker_ registry? I think you should just use testing gcr.io images or internal minikube one.

testdata/serviceconfig.yml
mandarjog: yaml will accept both `1` and `"1"` but in this context `1` is correct.

ZackButcher: They're identical: both `"1"` and `1` will be interpreted as strings => the `1` will be `"1"` at runtime. The proto field backing `value` is of type `string` because it represents an expression, not a literal value. However, in this case I want the expression literal `"1"`, and `"1"` correctly maps to `int64(1)` at runtime.

There is a difference between a string literal _in_ an expression (`"\"1\""`) and an expression represented as a string.

I'll update to `1` instead of `"1"` so its it looks more natural when a user glances over it (though IMO it's actually worse since it makes it less obvious what's actually happening).

ZackButcher: Turns out it won't actually parse if we use a literal `1`, keeping it as is (`"1"`).

kyessenov: YAML mandates integers be quoted if they are used in string fields (e.g. "args")

deploy/kube/README.md
ZackButcher: Added a comment to this effect.

ZackButcher:
Updated per comments. Open questions are:
- Do we want to move the yamls to some other dir?
- Do we want to separate the mixer yaml from the others?
- Do we want to remove the local registry yaml from this PR?

PR #275, opened 2017-02-23
Regularize metric aspect API.

geeknoid:
- A metric value now holds a pointer to a MetricDefinition instead
of merely having its name. This prevents individual adapters having
to maintain a map.

- Remove the Kind field from the metric value type, since it can be
obtained from the MetricDefinition instance.

- Pass a map of MetricDescription into the NewMetricsAspect method.
This enforces the uniqueness of names and avoids some code in some
adapters.

- Pass down the DisplayName and Description values from the
MetricDescriptor such that adapters can use those values when
configuring backends if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/275)

### DIFF
--- a/adapter/statsd/statsd.go
+++ b/adapter/statsd/statsd.go
@@ -97,7 +97,7 @@ func (b *builder) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErro
 	return
 }
 
-func (*builder) NewMetricsAspect(env adapter.Env, cfg adapter.AspectConfig, metrics []adapter.MetricDefinition) (adapter.MetricsAspect, error) {
+func (*builder) NewMetricsAspect(env adapter.Env, cfg adapter.AspectConfig, metrics map[string]*adapter.MetricDefinition) (adapter.MetricsAspect, error) {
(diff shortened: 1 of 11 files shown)


adapter/statsd/statsd.go
```diff
@@ -154,31 +154,19 @@ func (a *aspect) record(value adapter.Value) error {
...
 	}
-	switch value.Kind {
-	case adapter.Gauge:
+	if value.Definition.Kind == adapter.Gauge {
 		v, err := value.Int64()
 		if err != nil {
 			return fmt.Errorf("could not record gauge '%s' with err: %s", mname, err)
 		}
 		return a.client.Gauge(mname, v, a.rate)
-	case adapter.Counter:
```
ZackButcher: I don't think we want to remove this switch. While it's simpler today since we only support two kinds of metrics, in the (near?) future we'll be adding more kinds of metrics - whoever adds support for those is going to have to come in and reimplement the switch.

geeknoid: I added it back in.

PR #278, opened 2017-02-23
Implement the mixer's Quota API.

geeknoid:
- Replace the skeletal implementation of the quota aspect manager
with a fully-functioning version.

- Add the plumbing necessary to support passing API arguments from
the API handler down to the aspect managers.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/278)

### DIFF
--- a/adapter/memQuota/memQuota.go
+++ b/adapter/memQuota/memQuota.go
@@ -27,6 +27,7 @@
 package memQuota
 
 import (
+	"bytes"
 	"fmt"
 	"sort"
 	"strconv"
@@ -43,9 +44,6 @@ type builder struct{ adapter.DefaultBuilder }
 
 type memQuota struct {
 	sync.Mutex
-
-	// the definitions we know about, immutable
-	definitions map[string]*adapter.QuotaDefinition
 
 	// the counters we track for non-expiring quotas, protected by lock
 	cells map[string]int64
@@ -69,7 +67,7 @@ type memQuota struct {
 // we maintain a pool of these for use by the makeKey function
 type keyWorkspace struct {
 	keys   []string
-	buffer []byte
+	buffer bytes.Buffer
 }
 
 // pool of reusable keyWorkspace structs
@@ -128,7 +126,6 @@ func newAspect(env adapter.Env, c *config.Params, definitions map[string]*adapte
 // newAspect returns a new aspect.
 func newAspectWithDedup(env adapter.Env, ticker *time.Ticker, definitions map[string]*adapter.QuotaDefinition) (adapter.QuotasAspect, error) {
 	mq := &memQuota{
-		definitions: definitions,
 		cells:       make(map[string]int64),
 		windows:     make(map[string]*rollingWindow),
 		recentDedup: make(map[string]int64),
@@ -172,7 +169,7 @@ func (mq *memQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (int64, error
 		result := args.QuotaAmount
 
 		// we optimize storage for non-expiring quotas
-		if d.Window == 0 {
+		if d.Expiration == 0 {
 			inUse := mq.cells[key]
 
 			if result > d.MaxAmount-inUse {
@@ -189,7 +186,7 @@ func (mq *memQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (int64, error
 
 		window, ok := mq.windows[key]
 		if !ok {
-			seconds := int32((d.Window + time.Second - 1) / time.Second)
+			seconds := int32((d.Expiration + time.Second - 1) / time.Second)
 			window = newRollingWindow(d.MaxAmount, int64(seconds)*ticksPerSecond)
 			mq.windows[key] = window
 		}
@@ -213,7 +210,7 @@ func (mq *memQuota) ReleaseBestEffort(args adapter.QuotaArgs) (int64, error) {
 	return mq.commonWrapper(args, func(d *adapter.QuotaDefinition, key string) int64 {
 		result := args.QuotaAmount
 
-		if d.Window == 0 {
+		if d.Expiration == 0 {
 			inUse := mq.cells[key]
 
 			if result >= inUse {
@@ -246,11 +243,7 @@ func (mq *memQuota) ReleaseBestEffort(args adapter.QuotaArgs) (int64, error) {
 type quotaFunc func(d *adapter.QuotaDefinition, key string) int64
 
 func (mq *memQuota) commonWrapper(args adapter.QuotaArgs, qf quotaFunc) (int64, error) {
-	d, ok := mq.definitions[args.Name]
-	if !ok {
-		return 0, fmt.Errorf("request for unknown quota '%s' received", args.Name)
-	}
-
+	d := args.Definition
 	if args.QuotaAmount < 0 {
 		return 0, fmt.Errorf("negative quota amount %d received", args.QuotaAmount)
 	}
@@ -259,7 +252,7 @@ func (mq *memQuota) commonWrapper(args adapter.QuotaArgs, qf quotaFunc) (int64, 
 		return 0, nil
 	}
 
-	key := makeKey(args)
+	key := makeKey(args.Definition.Name, args.Labels)
 
 	mq.Lock()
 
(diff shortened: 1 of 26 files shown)


pkg/aspect/quotasManager_test.go
```diff
@@ -14,8 +14,319 @@
...
+	if err == nil {
+		t.Error("NewQuotasManager().NewAspect(conf, builder, test.NewEnv(t)) = _, nil; wanted err")
+	}
+	if !strings.Contains(err.Error(), errString) {
+		t.Errorf("NewAspect(conf, builder, test.NewEnv(t)) = _, %v; wanted err %s", err, errString)
+	}
+}
+
+func TestQuotaWrapper_Execute(t *testing.T) {
+	// TODO: all of these test values are hardcoded to match the metric definitions hardcoded in metricsManager
```
douglas-reid: wait, why are quota tests depending on the hardcoded values of the metrics stuff?

pkg/aspect/quotasManager.go
```diff
@@ -58,94 +74,129 @@ func (m *quotasManager) NewAspect(c *config.Combined, a adapter.Builder, env ada
...
+func findQuota(quotas []*aconfig.QuotasParams_Quota, name string) *aconfig.QuotasParams_Quota {
+	for _, q := range quotas {
+		if q.Descriptor_ == name {
+			return q
+		}
+	}
+	return nil
+}
+
+func quotaDescToDef(desc *dpb.QuotaDescriptor) (*adapter.QuotaDefinition, error) {
```
douglas-reid: similar to above: fwiw, `definitionFromProto` would more closely align with the other methods like this, including `ptypes.DurationFromProto` which is used within this method.

pkg/adapter/quotas.go
```diff
@@ -36,28 +36,38 @@ type (
...
-	QuotaKind int8
-
-	// QuotaDefinition is used to describe an individual quota the aspect will encounter at runtime
+	// QuotaDefinition is used to describe an individual quota that the aspect will encounter at runtime.
 	QuotaDefinition struct {
-		// MaxAmount specifies the upper limit for the quota
+		// Name of this quota definition.
+		Name string
+
+		// Optional user-friendly name for this quota.
```
douglas-reid: nit: go convention is to have comments start with the name of the thing being commented on and to be complete sentences. here, and below, this would change to comments like `// DisplayName is an optional user-friendly name for a quota definition`.

adapter/memQuota/memQuota.go
```diff
@@ -296,27 +288,25 @@ func (mq *memQuota) reapDedup() {
...
+	// ensure stable order
+	for k := range labels {
 		keys = append(keys, k)
 	}
-
-	// ensure stable order
 	sort.Strings(keys)
 
-	buffer = append(buffer, []byte(args.Name)...)
+	buffer = append(buffer, []byte(name)...)
```
douglas-reid: question (curious): why are we not using `bytes.Buffer` for the buffer here?

It has a `Reset()` that retains the underlying storage, and has a `WriteString` method that seems more natural here.

pkg/aspect/apiMethod.go
```diff
@@ -0,0 +1,75 @@
...
+func (a APIMethod) String() string {
+	return apiMethodToString[a]
+}
+
+type (
+	// APIMethodArgs provides additional method-specific parameters
+	APIMethodArgs interface{}
+
+	// CheckMethodArgs is supplied by invocations of the Check method.
+	CheckMethodArgs struct {
```
douglas-reid: do we care about embedding the interface in these structs here (to be explicit and allow sharing of "common" args, should they ever arise)?

adapter/memQuota/memQuota.go
```diff
@@ -296,27 +288,25 @@ func (mq *memQuota) reapDedup() {
 	}
 }
 
-// Produce a unique key representing the cell
-func makeKey(args adapter.QuotaArgs) string {
+// Produce a unique key representing the given labels.
+func makeKey(name string, labels map[string]interface{}) string {
```
geeknoid: Good suggestions. I'm leaving refactoring this out to a separate change.

adapter/memQuota/memQuota.go
geeknoid: I'm going to take the fifth on that question.

I've updated the code to use bytes.Buffer. Looks nice now.

pkg/aspect/quotasManager_test.go
geeknoid: Cut & paste turd, removed.

pkg/aspect/quotasManager_test.go
```diff
@@ -14,8 +14,319 @@
...
+	}
+}
+
+func TestQuotasManager_NewAspect(t *testing.T) {
+	builder := &fakeQuotaBuilder{name: "test", body: func() (adapter.QuotasAspect, error) {
+		return &fakeQuotaAspect{}, nil
+	}}
+
+	conf := &config.Combined{
+		Aspect: &pb.Aspect{
```
geeknoid: Done.

pkg/aspect/metricsManager.go
```diff
@@ -192,7 +192,7 @@ func evalAll(expressions map[string]string, attrs attribute.Bag, eval expr.Evalu
 	return labels, result.ErrorOrNil()
 }
 
-func definitionFromProto(desc *dpb.MetricDescriptor) (*adapter.MetricDefinition, error) {
+func metricDescToDef(desc *dpb.MetricDescriptor) (*adapter.MetricDefinition, error) {
```
geeknoid: Renamed to metricDefinitionFromProto

pkg/aspect/quotasManager.go
geeknoid: Renamed to quotaDefinitionFromProto.

pkg/adapter/quotas.go
geeknoid: Done.

geeknoid:
PTAL.

pkg/aspect/apiMethod.go
geeknoid: How do I embed an interface? 

I wouldn't really want embedding I think. I want to say "CheckMethodArgs is an APIMethodArgs".

ZackButcher: Embedding is how `is-a` relationships are achieved in golang. You embed it by by making it a field on the struct without a fieldname (essentially it makes the members of the embedded type, including methods, members of your type; you can "override" methods this way too):
```golang
type CheckMethodArgs struct {
    ApiMethodArgs
}
```

adapter/memQuota/memQuota.go
```diff
@@ -296,46 +289,65 @@ func (mq *memQuota) reapDedup() {
...
 	for _, k := range keys {
-		buffer = append(buffer, []byte(";")...)
-		buffer = append(buffer, []byte(k)...)
-		buffer = append(buffer, []byte("=")...)
+		buffer.WriteString(";")
+		buffer.WriteString(k)
+		buffer.WriteString("=")
 
-		switch v := args.Labels[k].(type) {
+		switch v := labels[k].(type) {
```
douglas-reid: non-blocker: one thing we might consider here is having the switch set a string var and then having the buffer.WriteString() call in one place (so we could handle the error, etc.).

```
var valStr string
switch v := labels[k].(type) {
case string:
   valStr = v
case int64:
   valStr = strconv.FormatInt(v, 16)
...
}
if _, err := buffer.WriteString(valStr); err != nil {
   // error handling 
}
```

We'd have to use the buffer of the other Pool in `case map[string]string`, so it gets a little awkward... but just food for thought.

geeknoid:
Done.

pkg/aspect/config/quotas.proto
```diff
@@ -20,7 +20,16 @@ option go_package="config";
 
 // Configures a quotas aspect.
 message QuotasParams {
-    // A list of names of istio.mixer.v1.config.descriptor.QuotaDescriptors determining the
-    // set of quotas that will be checked for any Quota call.
-    repeated string quota_descriptor_names = 1;
+    message Quota {
+        // Must match the name of some quota_descriptor in the global config.
+        string descriptor = 1;
```
ZackButcher: I think we want to change the `metric` field to `descriptor_name` rather than `descriptor`, we should do it here too if so. The reasoning is that the protoc generated golang struct renamed `descriptor` to `descriptor_` because protoc thinks of `descriptor` as a reserved word.

pkg/aspect/quotasManager.go
```diff
@@ -58,94 +74,129 @@ func (m *quotasManager) NewAspect(c *config.Combined, a adapter.Builder, env ada
...
+			DeduplicationID: strconv.FormatInt(atomic.AddInt64(&w.manager.dedupCounter, 1), 16),
+			BestEffort:      false,
 		}
 	}
 
-	// TODO: for now we don't support dedup semantics
-	dedupID := strconv.FormatInt(atomic.AddInt64(&w.manager.dedupCounter, 1), 16)
+	info, ok := w.metadata[qma.Quota]
+	if !ok {
+		return &Output{Code: code.Code_INVALID_ARGUMENT}, fmt.Errorf("unknown quota '%s'", qma.Quota)
```
ZackButcher: Is this correct? We should be using gogoprotos which should generate the enum `Code.INVALID_ARGUMENT`

pkg/aspect/config/quotas.proto
douglas-reid: should this be `descriptor_name` ?  this should also help avoid the unsightly `Descriptor_` in the generated code.

pkg/aspect/quotasManager.go
ZackButcher: @douglas-reid Did we miss converting some protos over to gogo?

pkg/aspect/inventory.go
```diff
@@ -108,6 +33,8 @@ func Inventory() ManagerInventory {
 			NewMetricsManager(),
 		},
 
-		QuotaMethod: {},
+		QuotaMethod: {
+			NewQuotasManager(),
```
douglas-reid: should we add a TODO somewhere to remove the NewQuotasManager() from CheckMethod when the proxy is ready?

pkg/aspect/quotasManager.go
douglas-reid: the code import is just an enum and we use the gen'd library directly. we could, of course, change that, if we wanted.  Just another thing to add to the WORKSPACE.  Feel free to file a bug assigned to me to track that down.

pkg/aspect/apiMethod.go
geeknoid: Done.

pkg/aspect/config/quotas.proto
geeknoid: Done.

adapter/memQuota/memQuota.go
geeknoid: The code is currently written to specifically avoid generating temp strings for better perf, so I'd rather not backslide.

pkg/aspect/config/quotas.proto
geeknoid: Done.

pkg/aspect/inventory.go
geeknoid: Done, added to inventory.go.

geeknoid:
PTAL.

PR #281, opened 2017-02-24
Minor updates to the deploy/kube/README.md

ZackButcher:
It was out of date when I submitted it earlier today: we removed metrics from the `Check` path, I found a typo or two, and the prometheus config wasn't very clear so I added a bit more info.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/281)

### DIFF
--- a/deploy/kube/README.md
+++ b/deploy/kube/README.md
@@ -14,7 +14,7 @@ mixer server with our test client (`mixc`), providing some attribute values:
 
 ```shell
 $ MIXS=$(minikube service mixer --url --format "{{.IP}}:{{.Port}}" | head -n 1)
-$ bazel run //cmd/client:mixc -- check -m $MIXS -a source.name=source,target.name=target,api.name=myapi,response.code=200,response.latency=100,client.id=$USER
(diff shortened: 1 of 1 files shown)


deploy/kube/README.md
```diff
@@ -14,7 +14,7 @@ mixer server with our test client (`mixc`), providing some attribute values:
 
 ```shell
 $ MIXS=$(minikube service mixer --url --format "{{.IP}}:{{.Port}}" | head -n 1)
-$ bazel run //cmd/client:mixc -- check -m $MIXS -a source.name=source,target.name=target,api.name=myapi,response.code=200,response.latency=100,client.id=$USER
+$ bazel run //cmd/client:mixc -- report "something happened" -m $MIXS -a source.name=source,target.name=target,api.name=myapi,api.method=v1.foo.bar,response.http.code=200,response.latency=100,client.id=$USER
```
douglas-reid: it doesn't matter at all, but I might make the example `report "http request"...` instead `report "something happened"` .

PR #286, opened 2017-02-27
Fix serviceconfig default example to work with proxy

andraxylia:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/286)


testdata/serviceconfig.yml
```diff
@@ -15,16 +15,16 @@ rules:
...
-          source: source.name
-          target: target.name
+          source: source.name | "unknown"
+          target: target.name | "unknown"
           service: api.name | "unknown"
           method: api.method | "unknown"
-          response_code: response.http.code
+          response_code: response.http.code | "unknown"
       - descriptor_name:  request_latency
         value: response.latency
```
ZackButcher: Add `| "unknown"` here too, this is also an expression that gets evaluated.

testdata/serviceconfig.yml
```diff
@@ -15,16 +15,16 @@ rules:
...
-          target: target.name
+          source: source.name | "unknown"
+          target: target.name | "unknown"
           service: api.name | "unknown"
           method: api.method | "unknown"
-          response_code: response.http.code
+          response_code: response.http.code | "unknown"
       - descriptor_name:  request_latency
-        value: response.latency
+        value: response.latency | "unknown"
```
douglas-reid: do we want "unknown" for latency values?  would "NaN" or something be more appropriate?

ZackButcher: Derp, yup, latency is a different type (it's a duration attribute). I'm not sure what a correctly typed zero value for a duration looks like. @mandarjog, do you know?

kyessenov: Depends on usage I think. If you select high latency requests, do you want ones with no latency? Or if you order requests by latencies, where should those without latency go?

ZackButcher: I mean from a syntactic point of view, I don't know how I write a duration literal that the expression evaluator will understand. Deciding how that zero value is handled is a totally separate issue that we can punt on till after the demo.

PR #288, opened 2017-02-28
Fix latency

andraxylia:
For some reason -1 does not work. There is no zero latency in this world, so we can probably use this value.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/288)


testdata/serviceconfig.yml
```diff
@@ -21,7 +21,7 @@ rules:
           method: api.method | "unknown"
           response_code: response.http.code | "unknown"
       - descriptor_name:  request_latency
-        value: response.latency | "unknown"
+        value: response.latency | 0
```
douglas-reid: can we use `"NaN"` here instead (i've verified that value works in config)? I worry that `0` will be handled incorrectly along the way... and actually be sent to backends.

we need something to signal the manager (or whatever) to not pass this value along here (or something).

@ZackButcher thoughts?

ZackButcher: I think zero is better because we don't do any handling of this value ourselves, we just pass it downstream systems (how does promtheus handle a NaN? I suspect not well). We're slated to come back and look at how we handle missing attributes as a whole, so I see this as a temporary fix.

douglas-reid: Prometheus handles "NaN" correctly (although, the counter permanently becomes NaN afterwards).  0 is fine for now (it will make our averages weird if we aren't careful though).

In the end, we need a histogram/distribution type for latency -- one that can include counts distinct from relying on request_count metrics for calculations, etc.

ZackButcher: Hmm, yeah, and as far as I can tell Prometheus doesn't provide an easy way to filter out values of zero. That's unfortunate. Since request_latency is currently a counter (we really need to add histogram support), we still shouldn't use NaN if it sets the counter to NaN for forever: if the proxy doesn't report latency once we lose all latency data for forever.

douglas-reid: OK. Let's unblock with 0.

Issue #285, opened 2017-02-27
Evaluating an attribute not in the request's context should return a zero value, not an error


ZackButcher:
Today if you try to call the mixer without specifying all of the attributes required to satisfy all expression evaluated at runtime the call will fail. This is because we return errors when we evaluate expressions that contain attributes that are not in the request's attribute bag. Rather than failing the evaluation, we should return a zero value instead so that we can always continue with reasonable values at runtime.

@mandarjog points out that some of the cases aren't straightforward: e.g. what's the zero value of an email?

kyessenov:
I think it would wise to never throw errors in expressions (think mysql, jquery, etc). It's up to you to assign values to undefined expressions such as 1/0 or empty email address.

mandarjog:
I agree that a *user* should never see a 500 because something could not be evaluated.

We could start with -- in case of an error Expression gets its baseTypes' default value.
We could let the user specify an alternate value, in case of an evaluation failure.

ldemailly:
I agree the | "unknown" needed all over isn't right

RoeyMaor:
The problem is that for many VALUE_TYPEs default value literals cannot be created. perhaps the approach should be the same as with the "ip" function defined here: https://github.com/istio/mixer/blob/535eb564667cef6aed334cb4f5e967a104768387/pkg/expr/func.go that allows creating IP_ADDRESS literals from strings. Maybe part of the solution is to define similar functions for the other VALUE_TYPEs that allow to create literals, that way we don't enforce "weird" zero values on the user and give him the option to use whatever values it wishes, 
and he might even choose to use those not as default values, but as the actual values for testing. Another discussion is whether These "new" string->value_type_literal functions should crash when the string given as input doesn't match their logical type format ("istio.io" as email, "42" as a timestamp, etc)
regardless of this, indeed when no default value is stated, there should be a warning rather than an error

mandarjog:
Forcing the user to give default value for an expression is similar to forcing a user to catch an exception. It is explicit and there are no surprises.

`ip` or other type conversion functions let you specify a typed default values, including create a typed literal. We are adding all those shortly, however that does not address the usability issue.

If that is turned into a warning, the

re: Crashing, Mixer should never crash because of user input. However evaluation may fail. 
Typing error of the kind you suggest are caught during validation stage. These do not result in runtime errors.

We can add `missingAttributePolicy` as separate settings for `{report, check} X {predicate, mapping}`. 
[...]

PR #294, opened 2017-02-28
Update the AccessLogManager

ZackButcher:
Update the access log manager to look more like the metric and quota managers. Unlink log templates from attribute names, deferring to a mapping instead. A follow-on PR will introduce using descriptors for both application and access logs at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/294)


ZackButcher:
Downside to this change is that we effectively remove custom access log templates. There's a hack to get them back in, but the final resolution is for us to plumb descriptor config into the managers.

ZackButcher:
Hold off for a bit on this one, it incorrectly conflates `LogEntry.Label`s with the values used to satisfy the log's template, rather than treating labels as dimensions on the log entry. I'll need to update the config proto to add a mapping from `log template variable name` -> `expression` and plumb that through.

ZackButcher:
Fixed the issue about mishandling labels, PTAL.

pkg/aspect/accessLogsManager.go
```diff
@@ -32,28 +33,54 @@ type (
...
 	commonLogFormat = `{{or (.originIp) "-"}} - {{or (.source_user) "-"}} ` +
-		`[{{or (.timestamp.Format "02/Jan/2006:15:04:05 -0700") "-"}}] "{{or (.apiMethod) "-"}} ` +
+		`[{{or (.timestamp.Format "02/Jan/2006:15:04:05 -0700") "-"}}] "{{or (.method) "-"}} ` +
 		`{{or (.url) "-"}} {{or (.protocol) "-"}}" {{or (.responseCode) "-"}} {{or (.responseSize) "-"}}`
 	// TODO: revisit when well-known attributes are defined.
 	combinedLogFormat = commonLogFormat + ` "{{or (.referer) "-"}}" "{{or (.user_agent) "-"}}"`
 )
 
 var (
+	commonLogTemplateExpressions = map[string]string{
```
geeknoid: I really don't like having attribute names baked into our code like that... I'd rather we leave the default blank and require the operator to supply a default block of config that produces the same thing.

In other words, we could only support the custom log format and require the operator configure things.

pkg/aspect/accessLogsManager.go
```diff
@@ -62,68 +89,67 @@ func NewAccessLogsManager() Manager {
...
+		},
 	}
 }
 
 func (accessLogsManager) ValidateConfig(c adapter.AspectConfig) (ce *adapter.ConfigErrors) {
 	cfg := c.(*aconfig.AccessLogsParams)
-	if cfg.LogFormat != aconfig.CUSTOM {
-		return nil
+	if cfg.Log == nil {
+		ce = ce.Appendf("Log", "An AccessLog entry must be provided.")
```
geeknoid: First letter of error messages should be lower case.

pkg/aspect/accessLogsManager.go
```diff
@@ -132,52 +158,31 @@ func (e *accessLogsWrapper) Close() error {
...
-		}
-		if val, found := attribute.Value(attrs, a); found {
-			entry.Labels[a] = val
-		}
-
-		// TODO: do we want to error for attributes that cannot
-		// be found?
+	buf := bufferPool.Get().(*bytes.Buffer)
+	if err := e.template.Execute(buf, templateVals); err != nil {
+		return &Output{Code: rpc.INTERNAL}, err
```
geeknoid: Put the buffer back in the pool?

pkg/aspect/accessLogsManager.go
ZackButcher: SGTM, I'll rip this out and stick it into our config files.

ZackButcher:
PTAL

testdata/serviceconfig.yml
```diff
@@ -30,7 +30,17 @@ rules:
   - kind: access-logs
     params:
       logName: "access_log"
-      logFormat: 0 # Common Log Format
+      log:
+        log_format: 1 # apache common log format
+        template_expressions:
```
douglas-reid: I like the flexibility this provides. I worry a bit about being able to grok the whole bit. As bad as it was to have common/combined hardcoded, it made it a *whole* lot simpler to understand from an operator's perspective (they didn't have to think about it).

Here, we give them half the information -- what the template vars are populated with. We don't declare what the actual template is.  We should think about a way of exposing that that is concise and easy to understand.

For comparison, here is the apache config:
```
LogFormat "%h %l %u %t \"%r\" %>s %b" common
CustomLog logs/access_log common
```

Interestingly, they provide the other half of the information (requiring knowledge of what the percent directives refer to).

I wonder if we should do the mapping in the definition of common and just let this aspect config consist entirely of `logName` and `logFormat`, as it does now.

testdata/serviceconfig.yml
```diff
@@ -30,7 +30,17 @@ rules:
...
+           originIp: origin.ip
+           source_user: origin.user
+           timestamp: request.time
+           method: request.method | ""
```
douglas-reid: why does this need a default value (and none of the others do?)

alternatively, could we put `| "-"` on all of the values here and make the template much simpler (removing all the `{or (...) "-"}` bits)?

pkg/aspect/accessLogsManager.go
```diff
@@ -32,98 +33,92 @@ type (
...
 	combinedLogFormat = commonLogFormat + ` "{{or (.referer) "-"}}" "{{or (.user_agent) "-"}}"`
 )
 
-var (
-	// TODO: revisit when well-known attributes are defined
-	commonLogAttributes = []string{"originIp", "source_user", "timestamp", "apiMethod", "url", "protocol", "responseCode", "responseSize"}
-	// TODO: revisit when well-known attributes are defined
-	combinedLogAttributes = append(commonLogAttributes, "referer", "user_agent")
-)
+var bufferPool = sync.Pool{
```
douglas-reid: i feel like we are reusing this concept (or similar) in a few places. is it time to make this more of a first-class thing within mixer?

pkg/aspect/accessLogsManager.go
```diff
@@ -32,98 +33,92 @@ type (
...
+		return
 	}
-	tmplStr := cfg.CustomLogTemplate
-	if _, err := template.New("test").Parse(tmplStr); err != nil {
-		return ce.Appendf("CustomLogTemplate", "could not parse template string: %v", err)
+	if cfg.Log.LogFormat != aconfig.CUSTOM {
+		// If it's not custom we're using our own configs, so we're fine.
+		return nil
 	}
+	// TODO: validate custom templates when users can provide us with descriptors
```
douglas-reid: i'm confused by this change. why are we removing the validation check on custom templates? Have we lost the ability to specify custom templates?  If so, should we remove the option altogether now (OR have explicit warnings that CUSTOM is NOT supported)?

pkg/aspect/config/accessLogs.proto
```diff
@@ -21,54 +21,61 @@ option go_package="config";
...
+            // Refers to the Apache [Combined Log Format](https://httpd.apache.org/docs/1.3/logs.html#combined).
+            // It is used to generate entries like:
+            // `127.0.0.1 user-identifier frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326 "http://www.example.com/start.html" "Mozilla/4.08 [en] (Win98; I ;Nav)"`
+            COMBINED = 2;
 
-        // Refers to the Apache [Combined Log Format](https://httpd.apache.org/docs/1.3/logs.html#combined).
-        // It is used to generate entries like:
-        // `127.0.0.1 user-identifier frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326 "http://www.example.com/start.html" "Mozilla/4.08 [en] (Win98; I ;Nav)"`
-        COMBINED = 1;
+            // Used to enable operator-specified formats.
```
douglas-reid: Is this usable now?

pkg/aspect/config/accessLogs.proto
```diff
@@ -21,54 +21,61 @@ option go_package="config";
...
+        AccessLogFormat log_format = 1;
 
-        // Used to enable operator-specified formats. This format should be used
-        // in tandem with `custom_log_template` to provide an alternate format
-        // for access log entries.
-        CUSTOM = 99;
-    }
+        // Only used if log_format is CUSTOM. Links this AccessLog to a LogEntryDescriptor
+        // that describes the log's template.
+        string descriptor_name = 2;
```
douglas-reid: i feel like COMMON and COMBINED should be made to be descriptors too.

pkg/aspect/config/accessLogs.proto
```diff
@@ -21,54 +21,61 @@ option go_package="config";
...
 
-    AccessLogFormat log_format = 3;
+        // Map of template variable name to expression for the descriptor's log_template. At
+        // run time each expression will be evaluated, and together they will provide values
+        // for the log's template string.
+        map<string, string> template_expressions = 3;
```
douglas-reid: we need some documentation of how template_expressions and labels interact.  for structured logging, the template behavior is irrelevant, but the the expression values that are generated do matter.

pkg/aspect/accessLogsManager.go
```diff
@@ -132,52 +127,33 @@ func (e *accessLogsWrapper) Close() error {
...
-
-	if len(entry.Labels) == 0 {
-		// don't write empty access logs
-		return &Output{Code: rpc.OK}, nil
-	}
-
-	buf := new(bytes.Buffer)
-	if err := e.template.Execute(buf, entry.Labels); err != nil {
-		return nil, err
+	entry := adapter.LogEntry{
```
douglas-reid: i can't tell from the changes. do the template expressions also make it into the logs as labels now (or not)?

they were before (just that the only actually occurring attribute was timestamp because we set it). example: `{"logName":"access_log","labels":{"timestamp":"2017-02-28T22:44:31.156184134Z"},"textPayload":"- - - [28/Feb/2017:22:44:31 +0000] \"- - -\" - -"}`

testdata/serviceconfig.yml
ZackButcher: I think ultimately the common/combined log templates should not be hard coded but migrated to descriptors. That way their config has both pieces.

pkg/aspect/accessLogsManager.go
ZackButcher: Yes we should. My follow on PR for updating application logs uses the same thing (in fact, when I merge them I'll have to delete one of the two pools since we're in the same package).

pkg/aspect/accessLogsManager.go
ZackButcher: Yeah, sorry I meant to make note of that in a comment on this github thread: we can't support custom templates until we plumb descriptors through our configs. We should probably just remove the variant from the enum.

pkg/aspect/config/accessLogs.proto
ZackButcher: I'm planning on doing that in a follow-on change. The descriptor has already been updated in the istio/api repo but I haven't bumped our SHA for that dependency in istio/mixer yet.

testdata/serviceconfig.yml
ZackButcher: The others should already be reported by the proxy, but request.method (with its new meaning of HTTP method) it not yet implemented.

ZackButcher:
Sorry, I didn't talk with y'all about my plan with these changes. The ultimate goal is make access/application log managers look like the quota and metric managers in that they use descriptors, but also to make them look similar in code structure so that it's easier for us to refactor away the managers later. Since both application and access logs depend on descriptors, it felt like too big a change to update both in one go, so instead I plan on 3 PRs:
1. This PR, so make the access log manager look like metric/quota managers in structure
2. A PR to make the application log manager look like this, metric, and quota managers in structure ([this branch has the WIP change, with I plan to send out for review tomorrow](https://github.com/ZackButcher/mixer/tree/applicationlogs)).
3. A final PR that makes both access and application log managers use descriptors for their logging.

It's possible that 2 and 3 can be combined (maybe even necessary given application logging doesn't make sense when the user can't give us a template to populate).

That's why the template strings for COMBINED and COMMON logs feel weird and out of place, for example: they'll end up being descriptors soon. But, we can't start using the descriptors without depending on the [new descriptor proto](https://github.com/istio/api/commit/f921098f6d62cb3316eee1932f3b405bcaa88cf6), which requires updates to access log manager too.

pkg/aspect/accessLogsManager.go
geeknoid: I've got an outstanding PR that addresses this.

pkg/aspect/accessLogsManager.go
ZackButcher: No, they don't. Labels and template expressions don't mix. My reasoning for this is that the `LogEntryDescriptor` defines a log's shape by dimensioning the payload with labels that are well typed. Separately, we need to construct the payload, which needs to be populated with runtime values, which means we must have expressions for the variables in the template. If we combined template expressions with labels then the `LogEntry` we give the adapter has the wrong shape. This will become more obvious when we actually perform validation: we'll validate that the `label` expressions that the user gives us in the aspect config evaluate to the type of the labels on the `LogEntryDescriptor`, but there's no reason for us to validate types of the template expressions because they'll wind up being strings in the payload.

pkg/aspect/config/accessLogs.proto
ZackButcher: Nope, I was debating just dropping the concept of custom logs in this PR. In theory the concept could be added back in the PR that introduces descriptors, but since we still don't let the user supply descriptors in config custom logs still have to be compiled into the mixer. So we could remove it now and add it back when descriptors are introduces, or even later when we plumb descriptors through config, or just leave it as-is with it not working.

testdata/serviceconfig.yml
douglas-reid: this seems like an odd artifact to surface in the config file (and will quickly become out of date). is this just a workaround for #285 ? If so, can we add a comment about that so future me will remember?

pkg/aspect/accessLogsManager.go
douglas-reid: now that the mentioned PR is merged, can we convert to that?

pkg/aspect/accessLogsManager.go
douglas-reid: We need a way to have access logs reported the constructed string as well as the bits that made up the template (for structured logging).

It sounds like we are saying now that you will have duplicated template_expressions and labels clauses in the case where you want both?

what if we just drop template_expressions all together and require that the expression be written in terms of the labels (if you want an expression)?

pkg/aspect/config/accessLogs.proto
douglas-reid: let's remove it for now.  adding it back in should be easy when we are ready. we can even track it in an issue.

pkg/aspect/config/accessLogs.proto
douglas-reid: SG.

ZackButcher:
PTAL

pkg/aspect/accessLogsManager.go
```diff
@@ -31,153 +31,117 @@ type (
...
 	pool.PutBuffer(buf)
+
+	entry := adapter.LogEntry{
+		LogName:     e.name,
+		Labels:      labels,
+		TextPayload: payload,
+	}
 	if err := e.aspect.LogAccess([]adapter.LogEntry{entry}); err != nil {
-		return Output{Status: status.WithError(err)}
+		return Output{Status: status.WithError(fmt.Errorf("failed to log %s with err: %s", e.name, err))}
```
douglas-reid: nit: do we want `failed to log to %s with...` or `failed to write %s with...` or `failed to append to %s with...` ?

if i have a log named `my_log`, if there is an error here, the current message would be: `failed to log my_log with err...`

pkg/aspect/config/accessLogs.proto
```diff
@@ -21,54 +21,63 @@ option go_package="config";
...
-// kind: access-logs
-// params:
-//    log_name: "logs/access_log"
-//    attributes: "source.ip","host.URL","
-//    log_format: "COMMON"
+//     kind: access-logs
+//     params:
+//        log_name: "logs/access_log"
+//        log:
+//            log_format: CUSTOM
```
douglas-reid: since we don't support CUSTOM at the moment, we should probably use a different example here.

pkg/aspect/config/accessLogs.proto
```diff
@@ -21,54 +21,63 @@ option go_package="config";
...
 
-    // A list of names of attributes that will be used to generate
-    // access log entries.
-    repeated string attributes = 2;
+    // Describes how attributes must be evaluated to produce values for a log message.
+    message AccessLog {
+        // AccessLogFormat specifies the format to use for generating textual log
+        // entries for the accessLogger aspect of the mixer.
+        //
+        // Default value: COMMON.
```
douglas-reid: Default value is no longer COMMON, now that we have added ACCESS_LOG_FORMAT_UNSPECIFIED.

We should update documentation.

testdata/serviceconfig.yml
```diff
@@ -31,7 +31,18 @@ rules:
   - kind: access-logs
     params:
       logName: "access_log"
-      logFormat: 0 # Common Log Format
+      log:
+        log_format: 1 # apache common log format
```
douglas-reid: curious: why does this use `_` here, while `logName` does not?

testdata/serviceconfig.yml
douglas-reid: Can we add a (duplicate) set of labels here to match the template_exprs (for completeness sake)?

testdata/serviceconfig.yml
ZackButcher: Both work, updated to use camel case as more idiomatic yaml.

This is gonna be tricky for our users down the line: `labels` (`map[string]string`) look the same as normal fields in yaml but _must_ match the label names exactly (so underscores matter there).

pkg/aspect/accessLogsManager.go
ZackButcher: How about "failed to report to %s with..."? I don't like the other verbs since we don't know if we're writing or appending or making a telephone call to Bob or something else entirely.

pkg/aspect/config/accessLogs.proto
ZackButcher: Updated.

pkg/aspect/accessLogsManager.go
ZackButcher: Oh meh this is the access manager, the verb isn't `report` like it is for the application log manager. Updating it to "failed to log to".

PR #295, opened 2017-02-28
Remove the golang genproto dependency from WORKSPACE.

douglas-reid:
This CL transitions to using gogoslick for the google/rpc/code.proto
messages. In doing so, it also removes the remaining uses of the
standard golang generated code for protos from the codebase.

Tested with:

```shell
$ bazel clean
$ bazel test ...
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/295)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -62,6 +62,7 @@ gogoslick_proto_library(
     name = "google/rpc",
     protos = [
         "google/rpc/status.proto",
+        "google/rpc/code.proto",
     ],
     importmap = {
         "google/protobuf/any.proto": "github.com/gogo/protobuf/types",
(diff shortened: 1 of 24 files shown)


ZackButcher:
I'm not sure I love collapsing the codes into the rpc namespace along with this. Some of the `code`s read better (e.g. `rpc.FAILED_PRECONDITION`) but others not so much (e.g. `rpc.INTERNAL`; IMO `code.INTERNAL` is more clear). It also makes it less obvious we're working with the canonical error codes. Maybe it's just because I'm used to working `code`.

douglas-reid:
it looks like the `option go` parts of the proto files are ignored by the current protoc setup. Not entirely sure why. I can dig further.

FWIW, `code.proto` declares its package as `google.rpc`. The collapsing then happens, in part, from the gogoslick options.

ZackButcher:
I don't think it's worth blocking this PR over `code` vs `rpc`.

douglas-reid:
OK. Then I'll submit now and maybe next week during a down moment will play with separating.

PR #297, opened 2017-03-01
Add duration handling to prometheus adapter.

douglas-reid:
This is a short-term fix for errors observed with latency testing.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/297)

### DIFF
--- a/adapter/prometheus/prometheus.go
+++ b/adapter/prometheus/prometheus.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


adapter/prometheus/prometheus.go
```diff
@@ -195,6 +196,11 @@ func promValue(val adapter.Value) (float64, error) {
 		return i, nil
 	case int64:
 		return float64(i), nil
+	case time.Duration:
+		d := time.Duration(i)
+		// TODO: what is the right thing here?
+		// convert to millis for now
+		return d.Seconds() * 1e3, nil
```
ZackButcher: No need to create a new duration, in this `case` golang knows that `i` must be a `time.Duration`, so we can just do `return i.Nanoseconds() / int(time.Millisecond), nil`.

douglas-reid: updated to remove `d := time.Duration(i)`.

i.Nanoseconds() is `int64`, so leaving with current construction.

douglas-reid:
PTAL.

PR #306, opened 2017-03-02
Add high-quality logging to diagnose attribute-related issues.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/306)

### DIFF
--- a/pkg/api/handler.go
+++ b/pkg/api/handler.go
@@ -73,7 +73,7 @@ func NewHandler(aspectExecutor Executor, methodMap map[aspect.APIMethod]config.A
 	}
 }
 
-// execute performs common function shared across the api surface.
+// execute performs common functions shared across the api surface.
 func (h *handlerState) execute(ctx context.Context, tracker attribute.Tracker, attrs *mixerpb.Attributes,
 	method aspect.APIMethod, ma aspect.APIMethodArgs) *rpc.Status {
 	ab, err := tracker.StartRequest(attrs)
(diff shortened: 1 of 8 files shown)


pkg/api/handler.go
```diff
@@ -141,6 +141,11 @@ func (h *handlerState) Report(ctx context.Context, tracker attribute.Tracker, re
 
 // Quota performs 'quota' function corresponding to the mixer api.
 func (h *handlerState) Quota(ctx context.Context, tracker attribute.Tracker, request *mixerpb.QuotaRequest, response *mixerpb.QuotaResponse) {
+	if glog.V(2) {
+		glog.Infof("Quota [%x]", request.RequestIndex)
+		defer func() { glog.Infof("Quota [%x] <-- %s", request.RequestIndex, response) }()
```
kyessenov: nit: deferring in logging might cause hard-to-debug memory issues if there're blocking calls inside the body

pkg/attribute/rootBag.go
```diff
@@ -304,52 +309,89 @@ func (rb *rootBag) update(dictionary dictionary, attrs *mixerpb.Attributes) erro
 		return err
 	}
 
+	log := glog.V(2)
+	if log {
+		glog.Infof("RB(%d): updating attribute bag", rb.id)
```
douglas-reid: nit: worth expanding `RB` to `RootBag` for clarity?

pkg/attribute/rootBag.go
```diff
@@ -304,52 +309,89 @@ func (rb *rootBag) update(dictionary dictionary, attrs *mixerpb.Attributes) erro
 		return err
 	}
 
+	log := glog.V(2)
```
douglas-reid: curious: would it be worth putting all of the log statements into a buffer and then issuing a single log statement so that all of the attribute bag updates would be in a single log line?

that way, we could avoid interweaving of updates in the logs across requests?

might make it easier to consume.

pkg/api/handler.go
geeknoid: Fixed.

pkg/attribute/rootBag.go
geeknoid: Done. Works much better that way.

pkg/attribute/rootBag.go
geeknoid: Done.

PR #309, opened 2017-03-02
Introduce a general-purpose pool of bytes.Buffer objects.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/309)

### DIFF
--- a/adapter/memQuota/BUILD
+++ b/adapter/memQuota/BUILD
@@ -11,6 +11,7 @@ go_library(
     deps = [
         "//adapter/memQuota/config:go_default_library",
         "//pkg/adapter:go_default_library",
+        "//pkg/pool:go_default_library",
         "@com_github_gogo_protobuf//types:go_default_library",
     ],
 )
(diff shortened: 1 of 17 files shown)


geeknoid:
Any suggestions for the names of such a combined package? With this PR, we
have:

intern.NewPool
buffer.Get
buffer.Put

Combined could be:

mem.NewStringPool
mem.GetBuffer
mem.PutBuffer

Other names?

geeknoid:
Or maybe the package could be called pool:

pool.InternString
pool.GetBuffer
pool.PutBuffer

This would eliminate having multiple string pools in favor of one global string pool.

ZackButcher:
`package pool` SGTM.

douglas-reid:
pool seems reasonable.

geeknoid:
PTAL.

ZackButcher:
Long term I think we'll want to group buffers by the size of their underlying array, like we do for strings, but that's probably overkill right now.

PR #311, opened 2017-03-02
Update the attribute protocol around stringmaps to enable delta stringmaps

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/311)

### DIFF
--- a/pkg/attribute/bag_test.go
+++ b/pkg/attribute/bag_test.go
@@ -141,6 +141,7 @@ func TestStringMapEdgeCase(t *testing.T) {
 
 	// full to empty
 	sm1 = &mixerpb.StringMap{Map: map[int32]string{}}
+	attrs.DeletedAttributes = []int32{1}
 	attrs.StringMapAttributes = map[int32]*mixerpb.StringMap{1: sm1}
 	_ = rb.update(d, attrs)
 }
--- a/pkg/attribute/rootBag.go
+++ b/pkg/attribute/rootBag.go
@@ -323,6 +323,24 @@ func (rb *rootBag) update(dictionary dictionary, attrs *mixerpb.Attributes) erro
 		rb.reset()
 	}
 
+	// delete requested attributes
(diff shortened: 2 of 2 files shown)


geeknoid:
This can be achieved with the delete feature.

In a single request, you specify you'd like to delete attribute FOO, and
then supply a string map attribute FOO which holds the new value you want.

Supporting these semantics is why I moved the block of delete functionality
up in the rootBag.update method.

Makes sense?

qiwzhang:
Does this feature subject to this "attribute_context" field:

https://github.com/istio/api/blob/master/mixer/v1/attributes.proto#L84

If it is 0,  is it always "replacement"?

If so, if is fine for me.  The issue is: client is NOT ready for delta update yet.

BTW, we may need deleted_keys for string_map. Otherwise, if a key is removed, we have to do whole map replacement.

geeknoid:
I'm not sure what you're asking about regarding attribute_context. An
attribute_context of 0 currently has no special semantics. Each individual
stream can have N parallel attribute contexts and operate on them
independently. There is currently no way to delete attribute contexts.

Yes, if we find that we need to often delete keys from a string map, we'll
need to figure out how to encode that. For the moment, I think we can live
without that feature.

Once I check in this change, you will need to add a delete_attribute item
for each string map you send in order to maintain replacement semantics.
Once the client knows how to do string map deltas, then you can remove the
delete_attribute items.

qiwzhang:
client code is not doing "delta_update" yet.  I through 0 attribute_context means 0 context, all attributes are in the request body.  If mixer is not doing that, we may have to support it.

qiwzhang:
I created issue https://github.com/istio/mixer/issues/313 to request such support.  The current communication between proxy and mixer is broken, unless proxy can quickly implement delta update.

geeknoid:
The client code can just set reset_context to true in every request it
sends. This has the effect of treating the whole request as a replacement
rather than a delta.

PR #328, opened 2017-03-04
Update how aspect managers return results.

geeknoid:
- Aspect managers no longer return both an Output and an error
since that was confusing. They now just return Output structs.

- Have the Output struct hold an rpc.Status instead of an rpc.Code
which allows richer error returns (since rpc.Status has a Message field)

- Make the parallel aspect dispatch code wait for all async work to complete
before returning instead of early-aborting on errors. This is so when, for
example, a Check call is denied, all the failure causes are returned in
one shot.

- Given the ability for Output to hold a Status object, this fixes the
Denials aspect manager to successfully pass-on status message text from
config.

- Make the aspect manager's various NewXXX into newXXX since they don't need
to be exposed out of the aspect package.

- Added the status package with utility functions to manipulate rpc.Status
structs

- Switched some code patterns to use structs rather than pointers in order to
avoid allocs.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/328)

### DIFF
--- a/pkg/adapterManager/parallelManager.go
+++ b/pkg/adapterManager/parallelManager.go
@@ -15,18 +15,23 @@
 package adapterManager
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"sync"
 
+	rpc "github.com/googleapis/googleapis/google/rpc"
+
 	"istio.io/mixer/pkg/aspect"
 	"istio.io/mixer/pkg/attribute"
 	"istio.io/mixer/pkg/config"
+	"istio.io/mixer/pkg/pool"
+	"istio.io/mixer/pkg/status"
 )
 
-// ParallelManager is an istio.io/mixer/pkg/api.Executor which wraps a Manager. Its BulkExecute method calls the
-// wrapped Manager's Execute in parallel across a pool of go routines. The pool of worker go routines is shared by all
-// request go routines, i.e. it's global.
+// ParallelManager is an istio.io/mixer/pkg/api.Executor which wraps a Manager. Its Execute method calls the
+// wrapped Manager's Execute in parallel across a pool of goroutines. The pool of worker goroutines is shared by all
+// request goroutines, i.e. it's global.
 type ParallelManager struct {
 	*Manager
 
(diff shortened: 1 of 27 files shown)


pkg/adapterManager/parallelManager.go
```diff
@@ -57,33 +62,57 @@ func NewParallelManager(manager *Manager, size int) *ParallelManager {
...
 	for i := 0; i < numCfgs; i++ {
 		select {
 		case <-ctx.Done():
-			return nil, fmt.Errorf("deadline exceeded waiting for adapter results with err: %v", ctx.Err())
+			return aspect.Output{Status: status.WithDeadlineExceeded(fmt.Sprintf("deadline exceeded waiting for adapter results with err: %v", ctx.Err()))}
 		case res := <-r:
-			if res.err != nil {
-				// TODO: should we return partial results too when we encounter an err?
-				return nil, fmt.Errorf("%s returned err: %v", res.name, res.err)
+			results[i] = res
```
douglas-reid: should we make fail-fast an option?

geeknoid:
Do you think this would be useful?

The whole operation has a timeout, so it'll only run for so long. And this
change only affects how long it takes to report a check denial or report
failure, which should generally both be rare.

pkg/adapterManager/parallelManager_test.go
```diff
@@ -55,12 +56,12 @@ func TestExecute(t *testing.T) {
 			{&configpb.Adapter{Name: c.name}, &configpb.Aspect{Kind: c.name}},
 		}
 
-		o, err := m.Execute(context.Background(), cfg, nil, nil)
-		if c.inErr != nil && err == nil {
-			t.Errorf("m.Execute(...) = %v; want err: %v", err, c.inErr)
+		o := m.Execute(context.Background(), cfg, nil, nil)
```
mandarjog: This feels unusual for go.
If a function can result in an error, should it no return an error?

pkg/api/handler.go
```diff
@@ -92,31 +93,21 @@ func (h *handlerState) execute(ctx context.Context, tracker attribute.Tracker, a
...
-		return newStatusWithMessage(rpc.INTERNAL, msg)
+		return status.WithInternal(msg)
 	}
 	cfg := untypedCfg.(config.Resolver)
 	cfgs, err := cfg.Resolve(ab, h.methodMap[method])
 	if err != nil {
 		msg := fmt.Sprintf("unable to resolve config: %v", err)
 		glog.Error(msg)
-		return newStatusWithMessage(rpc.INTERNAL, msg)
+		return status.WithMessage(rpc.INTERNAL, msg)
```
mandarjog: `status.WithInternal(msg)` ?

pkg/aspect/manager.go
```diff
@@ -56,6 +57,16 @@ type (
 		io.Closer
 
 		// Execute dispatches to the adapter.
-		Execute(attrs attribute.Bag, mapper expr.Evaluator, ma APIMethodArgs) (*Output, error)
+		Execute(attrs attribute.Bag, mapper expr.Evaluator, ma APIMethodArgs) Output
 	}
 )
+
+// IsOK returns whether the Output represents success or failure
+func (o Output) IsOK() bool {
```
mandarjog: What is the default state of newly created Output Struct?

pkg/adapterManager/parallelManager_test.go
ZackButcher: The problem then becomes "how do I map an err to a status?". I think the better way to think of this is that we're returning an error in the API domain rather than the golang domain. That being said, I think it's still reasonable for our aspect interface's return type to be `(Output, err)` so that adapters can differentiate between "I processed the user's request and rejected it" (API err) and "things are broken, I can't do any work" (golang err).

pkg/api/handler.go
geeknoid: Done

pkg/aspect/manager.go
geeknoid: Default state is OK, since rpc.Code.OK==0

pkg/adapterManager/parallelManager_test.go
geeknoid: Zack captures the rational. These are API errors, not go errors.

Having dual return values is confusing. I had some mismatch in the quota aspect manager for example. With this model, the aspect managers are responsible for producing API-level errors by creating an Output object. Any go-level error they encounter must be handled and/or and turned into an API error.

ZackButcher: I think we could also make the reasonable argument that due to config validation we shouldn't be hitting golang level errors in adapter execution, but I think we'd have to evaluate that statement later when we've written more adapter code.

geeknoid: Completely agree. I've been holding off on making that change for a while, waiting for things to settle more first.

Issue #63, opened 2016-12-14
Define plan for monitoring mixer itself


douglas-reid:
We should figure out how we want to monitor the mixer process itself.  There is probably a way to do it in a slick way using the configured adapters, etc., but we should design/plan for it and start working on it. In particular, we should start defining what (beyond basic request count, latencies, go routines, mem usage, etc.) would be interesting to monitor (most likely, adapter metadata).

wora:
Mixer essentially needs to do self billing/logging/monitoring/quota/etc. We need to design it carefully to avoid infinite loop.

cmluciano:
Are you interested in a PR with promotheus instrumentation?

douglas-reid:
I think we need more than just a prometheus instrumentation PR. For instance, it might be nice if mixer's metrics would be exported through whatever backend metrics adapters were configured for a running instance (could be prometheus, could be statsd, could be cloudwatch, etc.).

If you have prometheus expertise, however, we could always use another set of eyes on the //adapter/prometheus code that currently exists. Any improvements and suggestions there would be more than welcome.

I'd also be interested in a sketch of what metrics we think we should collect about ourselves (which would be independent of prometheus). That level design for instrumentation would be a good first step. If that sounds good to you, then I'd love to work with you on that too.

kyessenov:
I think it would be nice to reuse mixer backend adapters for manager/proxy too. The important bit is to allow direct passthrough so we don't end up monitoring the monitoring path.

geeknoid:
Hey Doug, can this be considered done?

douglas-reid:
Yes. Let's close this. We can always make our existing choices/impl better, but we have monitoring in place now.

PR #338, opened 2017-03-07
Draft proposal for Istio component monitoring.

douglas-reid:
This PR consists of an early strawman draft for how we might
establish monitoring across Istio components. This is based on a
loose set of conversations and emails related to the
problem of having Mixer monitoring itself and providing a mechanism
for the rest of Istio to send its logs and metrics to the backends
configured via Mixer.

In an attempt to have the discussion out in the open -- and for
the development of the proposal to be tracked explicitly
throughout the process -- this PR establishes a `doc/proposal`
directory in istio/mixer and clearly marks the status of this
proposal as DRAFT.

I expect that it will take several PRs for this proposal to reach
a completed, final, broadly-reviewable state, after which a subsequent PR
will be needed to move it out of the `proposal` directory (destination TBD).

The tracking bug for this issue is: #63.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/338)


doc/proposal/monitoring.md
```diff
@@ -0,0 +1,144 @@
...
+Each Istio component will be responsible for the following:
+1. Generating its own set of telemetry data (logs/metrics)
+1. Reporting that data to the Istio Mixer, on a regular basis, via the existing
+   Mixer API (see: [ReportRequest](https://github.com/istio/api/blob/master/mixer/v1/report.proto#L23)).
+1. Publishing the set of schemas for the telemetry data generated by its 
+   component in a *discoverable* fashion.
+
+In order to prevent a never-ending loop of Reports, Mixer will be configured to
+drop ReportRequests received where the source of the traffic being reported is
+Mixer itself.
```
geeknoid: When would this happen?

doc/proposal/monitoring.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Telemetry Generation and Reporting
+
+Schemas for telemetry data are necessary for Mixer to properly convert 
+ReportRequests into logs and metrics for various backend systems (see: 
+[LogEntryDescriptor](https://github.com/istio/api/blob/master/mixer/v1/config/descriptor/log_entry_descriptor.proto#L22),
+[MetricsDescriptor](https://github.com/istio/api/blob/master/mixer/v1/config/descriptor/metric_descriptor.proto#L53)).
+
+It would be extremely onerous to require that these schemas be kept in sync with
```
geeknoid: I thought we'd just treat the proxy, the manager, and mixer in the same way as we do services we manage. That is, the proxy, the manager, and the mixer would all have service config files in the same way services do. These service config files would come alongside the binaries for these components and would be included in the overall istio config for the deployment.

doc/proposal/monitoring.md
wora: where the source => where the origin

doc/proposal/monitoring.md
douglas-reid: That is an option, of course. But it would mean that we would have to generate (and/or manually maintain) the set of metrics also mentioned in the code. Generating the config file from the codebase and including in a config package is certainly an option for publishing.

This feels potentially problematic to me -- and isn't really how we do monitoring internally (not all counters are sent through Chemist for monitoring and alerting -- and certainly aren't all in service config files). For example, consider the difference between the metrics listed in serviceconfig.yaml and the set of metrics created and tracked by the servicecontrol codebase.

doc/proposal/monitoring.md
douglas-reid: this would happen in cases when there was a proxy in front of a logging/metrics backend, and that proxy was also reporting to mixer.

geeknoid: That's what I thought. So this recursion happens when a mixer backend is itself part of the Istio deployment.

One option is to declare that this is not allowed. More specifically, we could require that the mixer can only talk to backends that are not part of the same deployment. So if you want to deploy a backend for use by the mixer, you'd deploy it independently. This would eliminate the recursion.

If we don't take this position and we somehow neuter the telemetry coming from the downstream proxy, then how will this downstream proxy and the backend be monitored? If you deploy the backend independently, it'll be fully manageable in the normal way.

Basically, what this means is that the proxy for the service talks to mixer instance A, while the proxy for the backend talks to mixer instance B.

doc/proposal/monitoring.md
geeknoid: The OP model we have today is that the service provider must produce a service config file for its metrics, independent of what's in the code. Having a model to auto-generate this config based on code would be terrific, but is a separable problem.

Were you imagining a more dynamic setup where the metric schemas are disclosed on-the-fly during runtime?

douglas-reid: I was considering all of the metrics that live in code that OP is not aware of, but are used for monitoring by internal systems. Most of the dashboards I've used for oncall monitoring have not been built from the metrics exposed through OP. Instead, we rely on an internal metrics library (think `/varz` here).

I was considering a dynamic setup, yes. Though, I'm not entirely convinced that this is optimal or even a good idea. It may be necessary, however (depending on how our config story evolves, etc.). The real issue with a dynamic setup, as I see it currently, is that Report doesn't take fully formed metric values. If it did, this would be much simpler. As it stands now, however, we need to export not just the definitions, but also how attributes map into those definitions, and then also ensure that the Reporting uses those exact attributes.

doc/proposal/monitoring.md
douglas-reid: I don't think we can declare that it is not allowed (and i'm not entirely clear on what the story should be for egress proxies that are involved for services outside deployments, etc.)

We would not neuter the telemetry for all data hitting that backend necessarily. Mixer itself could keep stats on backend calls and/or it could Report them to a special adapter that collected them for pushing on a periodic basis (rather than a request-triggered basis).

The key bit is that a request from a mixer to a backend should not trigger another request from mixer to the backend.

doc/proposal/monitoring.md
geeknoid: I think the varz stuff may be historical and not really what we'd like. An external service built on endpoints wouldn't get that, only stuff declared in serviceconfig is relevant to them. I'd rather have everything be explicitly declared. Ideally of course, it would be declared in one place.

It's also worth thinking about code gen. Instead of scrapping code to generate config, you can start with config and generate necessary code stubs. This would let you write all the descriptors in config and then have nice strongly-typed stubs for use in the code.

My expectation for how Check and Report will work for services is that the calls will accept an "event type" parameter. The proxy would generate Check and Report calls with the 'proxy' event type. Call this Check:proxy and Report:proxy if you will. Services can also call Check and Report, but specify their own event types. For example, maybe Report:create_disk. The event type will be used to drive which config to use for the particular Check or Report call. The service will have a manifest that specifies all the event types it can produce. The operator then is responsible for providing config for all the supported event types.

douglas-reid: Codegen is an option... but it feels backwards to me. Today, using something like prometheus, I declare the metrics I'm interested in up front and manipulate them in code. I don't declare/manage them somewhere else and then use them in code via generated code. If we provide libraries that work in a similar fashion (and handle all of the mapping/translation to Istio behind the scenes), then there isn't really any new process to understand or work out for developers, etc. In my head, the fewer number of exposures to Istio config one experiences, the better.

I'm not sure I fully understand the event type model you are proposing here. Maybe we should chat offline about it?

I was imaging a generated `ReportRequest` from say the manager that looked something like:

```
{ 
  "request_index":1 
  "attributes":{
    "dictionary":{0:"manager_known_configs" 1:"manager_config_updates" 2:"manager_known_proxies"...}
    "int64_attributes":{0:2365 1:9991 2:35}
    ...
   }
   ...
}
``` 

and resulted in pushing to a backend the following values:

```
{
  "definition":{"name": "manager_known_configs" kind:Counter ...}
  "metricValue":2365
  ...
}

{
  "definition":{"name":"manager_config_updates" kind:Counter ...}
  "metricValue":9991
  ...
}

{
  "definition":{"name":"manager_known_proxies" kind:Counter ...}
  "metricValue":35
  ...
}
```

It isn't clear to me where event type slots into that model.

doc/proposal/monitoring.md
```diff
@@ -0,0 +1,144 @@
...
+Service Mesh itself.
+
+## Proposed Design
+
+The proposed design adopts two high-level goals:
+1. **Limit special handling for Istio components**
+   - Telemetry data for the Istio Proxy and Istio Manager components _should_ be
+     reported via the mixer through the same mechanisms as telemetry data for 
+     the non-Istio services within the service mesh.
+   - Telemetry data for the Istio Mixer itself _should_ be sent through the same
```
ZackButcher: Why is this a requirement? In particular I'm worried about the case where the mixer is misconfigured and cannot report metrics/logs - how do we detect and report that case? As you point out below, it's not necessarily the common case that metrics are reported through Chemist internally. Certainly none of the services I've worked on internally reported metrics via Chemist. Why impose this as a requirement for Istio components, and the mixer especially (which is even more a special case than the other Istio components - quis custodiet ipsos custodes)?

Consider the following straw man instead:
As part of installing Istio in your cluster a Istio-private statsd<sup>[1](#footnote1)</sup> + datastore instances are stood up. Critical metrics (received requests, response status by code, latency per method, number of outstanding requests, amount of outstanding metrics/logs to report, etc.) are pushed to this statsd instance. These metrics are sent to statsd via a normal statsd client, not touching any reporting code paths in the mixer. We use those metrics to ensure the deployment is healthy.

The obvious downside is that under the scheme you propose, metrics for Istio components show up next to metrics for my own services, while under this straw man I have to go somewhere else to look at the health of my Istio deployment. This could be remedied by double-reporting mixer metrics: metrics are reported to statsd out of band of any mixer report-specific codepaths, and _also_ reported via a self `Report` so they're exported to whatever metrics backends I have configured.

<hr>
<a name="footnote1">1.</a> It could be Prometheus or whatever other stats backend we want. I propose statsd in this example because Envoy already self reports via statsd, which means we can easily extend this straw man to the proxy reporting critical metrics out of band of the mixer too. This is especially important if the mixer is running behind a proxy itself.

doc/proposal/monitoring.md
```diff
@@ -0,0 +1,144 @@
...
+behavior of the individual Istio components (Proxy, [Manager](http://github.com/istio/manager),
+Mixer).
+
+This document proposes a high-level design for the monitoring of the Istio 
+Service Mesh itself.
+
+## Proposed Design
+
+The proposed design adopts two high-level goals:
+1. **Limit special handling for Istio components**
```
ZackButcher: Add a line here to get the bullet point on its own.

doc/proposal/monitoring.md
```diff
@@ -0,0 +1,144 @@
...
+   - Telemetry data for the Istio Proxy and Istio Manager components _should_ be
+     reported via the mixer through the same mechanisms as telemetry data for 
+     the non-Istio services within the service mesh.
+   - Telemetry data for the Istio Mixer itself _should_ be sent through the same 
+     processing code as data received via RPCs.
+1. **Prevent mixer-generated traffic from generating additional reports**
```
ZackButcher: newline to separate the sub-bullet above from this

doc/proposal/monitoring.md
```diff
@@ -0,0 +1,144 @@
...
+   - In order to prevent a never-ending loop of reporting of metrics, traffic
+     generated by mixer to downstream services (e.g. logging/metrics backends)
+     _must_ be exempted from triggering the push of new telemetry data to 
+     backend systems.
+
+### Design Summary
+
+Each Istio component will be responsible for the following:
+1. Generating its own set of telemetry data (logs/metrics)
```
ZackButcher: Newline between this and above so it's correctly detected as a numbered list. You might need newlines between the numbered points too.

doc/proposal/monitoring.md
douglas-reid: Good question. I don't think that we want to prevent istio components from using other mechanisms for reporting, but just that we want to encourage them to use mixer for their *primary* reporting channel (as you suggest in your downside comments).

The reason that I think we want this (as compared to the internal case) is that there is no universally adopted metrics / logs side channel externally. I don't want to require that, in addition to Istio, operators have to install a prometheus instance or a statsd instance or whatever. Doubly so for when they have a custom metrics backend and want everything to be monitored by the same tools. And I don't think we want duplicated logic in proxy and manager and mixer for various other systems.

If mixer cannot report logs/metrics, then I think there would be bigger issues.

Perhaps we need customer / other feedback here?

geeknoid:
PS: Regarding this experiment, I don't feel as though Github is a great to have this kind of design discussion. I'd rather if we use Google docs to track the discussions. We can create a Github issue that holds a link to the doc which would connect all the dots...

douglas-reid:
@geeknoid as you suggest, i will be moving this to a doc in the team drive this week (to clear out the backlog from our PRs).

douglas-reid:
Ok. I've finally gotten around to moving the content here to a google doc in the team drive. There has been a lot of good thinking along this path since this original draft was put together, so it needs a lot of work, but at the very least, the content will live on there.

Closing out this PR.

PR #339, opened 2017-03-07
Fill-in QuotaResponse for Quota API calls.

geeknoid:
This weaves a method return value through the layer in a way similar to how I weaved method arguments through the layers. This change introduces a TTL at the quota adapter layer, needed for prefetching of rate limiting quotas.

Also note the lack of tests for the changes in api/handlers.go. That whole file has almost non-existent testing for the time being. I plan to rewrite this file nearly completely to support OOO evaluation of incoming mixer API calls. When I do the rewrite, I'll add a full complement of tests.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/339)

### DIFF
--- a/adapter/memQuota/memQuota.go
+++ b/adapter/memQuota/memQuota.go
@@ -52,16 +52,21 @@ type memQuota struct {
 	windows map[string]*rollingWindow
 
 	// two ping-ponging maps of active dedup ids
-	recentDedup map[string]int64
-	oldDedup    map[string]int64
+	recentDedup map[string]dedupState
+	oldDedup    map[string]dedupState
 
 	// used for reaping dedup ids
 	ticker *time.Ticker
 
 	// indirection to support fast deterministic tests
-	getTick func() int64
+	getTime func() time.Time
 
 	logger adapter.Logger
+}
+
+type dedupState struct {
+	amount int64
+	exp    time.Time
 }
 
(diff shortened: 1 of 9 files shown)


adapter/memQuota/memQuota.go
```diff
@@ -150,21 +155,16 @@ func (mq *memQuota) Close() error {
 	return nil
 }
 
-// getCurrentTicks returns the number of ticks since January 1st 1970 UTC
-func getCurrentTicks() int64 {
-	return time.Now().UnixNano() / nanosPerTick
-}
-
-func (mq *memQuota) Alloc(args adapter.QuotaArgs) (int64, error) {
+func (mq *memQuota) Alloc(args adapter.QuotaArgs) (int64, time.Duration, error) {
```
douglas-reid: makes me wonder if we want a type like `QuotaResults` in case we decide we later want to have something more than the amount and valid duration in returned results.

so something like: `func Alloc(args adapter.QuotaArgs) (adapter.QuotaResults, error)`

geeknoid: Yeah, I was debating whether it was worth it. I guess it is, it would also be symmetric with the input and how I handle this elsewhere.

geeknoid:
Added a QuotaResult struct for use by the adapters.

PTAL.

Issue #340, opened 2017-03-07
Invalid attribute updates can leave attribute dictionary in bad state


geeknoid:
atrribute.Tracker.StartRequest consumes a new dictionary unconditionally even in the case of a botched update. This means even though the RPC as a whole will have failed, the new dictionary supplied with the RPC will get used. This leads to inconsistencies in the protocol.

PR #341, opened 2017-03-07
Fixed bug in attribute tracker around invalid attribute updates.

geeknoid:
- The attribute tracker would get confused if a new dictionary was
supplied with a mixer API request, along with a bogus attribute
update message. When this happened, the new dictionary was consumed
and used, even though the attribute update failed. This would leave
the end-to-end protocol in a confused state until a new
dictionary was pushed.

This fixes issue #340

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/341)

### DIFF
--- a/pkg/api/handler.go
+++ b/pkg/api/handler.go
@@ -88,14 +88,14 @@ func (h *handlerState) execute(ctx context.Context, tracker attribute.Tracker, a
 	// get a new context with the attribute bag attached
 	ctx = attribute.NewContext(ctx, ab)
 
-	untypedCfg := h.cfg.Load()
-	if untypedCfg == nil {
-		// config has NOT been loaded yet
-		const msg = "Configuration is not available"
+	cfg := h.cfg.Load().(config.Resolver)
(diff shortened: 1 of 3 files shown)


pkg/api/handler.go
```diff
@@ -88,14 +88,14 @@ func (h *handlerState) execute(ctx context.Context, tracker attribute.Tracker, a
 	// get a new context with the attribute bag attached
 	ctx = attribute.NewContext(ctx, ab)
 
-	untypedCfg := h.cfg.Load()
-	if untypedCfg == nil {
-		// config has NOT been loaded yet
-		const msg = "Configuration is not available"
+	cfg := h.cfg.Load().(config.Resolver)
```
mandarjog: trying to type assert nil to anything results in panic.
https://play.golang.org/p/59t2a3-Tsf
`panic: interface conversion: interface {} is nil, not config.Resolve`

That is why we need to fetch untypedCfg first.

mandarjog:
tests for this change ?

PR #343, opened 2017-03-07
Update ApplicationLogsManager

ZackButcher:
Follow on to PR https://github.com/istio/mixer/pull/294 that updates application logs to look like metric/quota managers. A follow on PR will update the access log manager to use descriptors as this one does, and a final PR will plumb descriptors through the mixer's config.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/343)

### DIFF
--- a/pkg/aspect/applicationLogsManager.go
+++ b/pkg/aspect/applicationLogsManager.go
@@ -40,12 +40,13 @@ type (
 	applicationLogsManager struct{}
 
 	logInfo struct {
-		format    PayloadFormat
-		severity  string
-		timestamp string
-		tmpl      *template.Template
-		tmplExprs map[string]string
-		labels    map[string]string
+		format     PayloadFormat
+		severity   string
+		timestamp  string
+		timeFormat string
+		tmpl       *template.Template
+		tmplExprs  map[string]string
+		labels     map[string]string
 	}
 
 	applicationLogsWrapper struct {
@@ -57,7 +58,7 @@ type (
 
 const (
 	// TEXT describes a log entry whose TextPayload should be set.
-	TEXT = iota
+	Text PayloadFormat = iota
 	// JSON describes a log entry whose StructPayload should be set.
 	JSON
 )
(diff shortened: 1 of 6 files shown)


pkg/aspect/applicationLogsManager.go
```diff
@@ -17,56 +17,49 @@ package aspect
...
-			"requestSize",
-			"responseSize",
-			"responseTime",
-			"originIp",
-			"originHost",
-		},
-	}
+const (
+	// TEXT describes a log entry whose TextPayload should be set.
+	TEXT = iota
```
douglas-reid: should we just have this be `Text`? I don't think we need all caps (not an initialism, like JSON).

Also, should these be PayloadFormats? I think we want that typing reflected here.

pkg/aspect/applicationLogsManager.go
```diff
@@ -17,56 +17,49 @@ package aspect
...
+	// PayloadFormat describes the format of the LogEntry's payload.
+	PayloadFormat int
+
 	applicationLogsManager struct{}
 
+	logInfo struct {
+		format    PayloadFormat
+		severity  string
+		timestamp string
+		tmpl      *template.Template
```
douglas-reid: is the idea that the timestamp format is now specified in the Template?

pkg/aspect/applicationLogsManager.go
```diff
@@ -108,107 +128,100 @@ func (applicationLogsManager) ValidateConfig(c adapter.AspectConfig) (ce *adapte
...
+			result = multierror.Append(result,
+				fmt.Errorf("failed to eval time for log entry '%s', continuing with time.Now(). Eval err: %s", name, err))
+			et = time.Now()
+		}
+		t, _ := et.(time.Time) // we don't check the cast because expression type checking ensures we get a time.
+
+		entry := adapter.LogEntry{
+			LogName:   e.name,
+			Labels:    labels,
+			Timestamp: t.String(),
```
douglas-reid: i asked this above, but this feels like we've lost the ability to specify how the timestamp is formatted in the logs. Is that the case?

pkg/aspect/config/applicationLogs.proto
```diff
@@ -19,42 +19,35 @@ package pkg.aspect.config;
...
+    message ApplicationLog {
+        // Must match the name of some LogEntryDescriptor.
+        string descriptor_name = 1;
+
+        // The expression to evaluate to determine this log's severity at runtime.
+        string severity = 2;
+
+        // The expression to evaluate to determine this log's timestamp.
+        string timestamp = 3;
+        // The golang time layout string to use to parse the timestamp
```
douglas-reid: missing new line. 

also, can we add an example for link to documentation on the format here?

finally, the comment says "parse". before, this was used for generation of timestamps in the logs.

pkg/aspect/applicationLogsManager.go
ZackButcher: Exactly

pkg/aspect/applicationLogsManager.go
ZackButcher: Yup we definitely do, thanks for catching that.

ZackButcher:
PTAL

PR #345, opened 2017-03-07
Fix bug I introduced around config evaluation.

geeknoid:
- Fix panic when trying to process an incoming request before
config has been loaded.

- Add a unit test to show that an error processing attributes leaves
the attribute state unchanged.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/345)

### DIFF
--- a/pkg/attribute/tracker_test.go
+++ b/pkg/attribute/tracker_test.go
@@ -86,3 +86,231 @@ func BenchmarkTracker(b *testing.B) {
 		}
 	}
 }
+
+func TestTracker_StartRequest(t *testing.T) {
+	t9 := time.Date(2001, 1, 1, 1, 1, 1, 9, time.UTC)
+	t10 := time.Date(2001, 1, 1, 1, 1, 1, 10, time.UTC)
+	ts9, _ := ptypes.TimestampProto(t9)
+	ts10, _ := ptypes.TimestampProto(t10)
+	d := time.Duration(42) * time.Second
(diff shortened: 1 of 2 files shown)


pkg/attribute/tracker_test.go
```diff
@@ -86,3 +86,232 @@ func BenchmarkTracker(b *testing.B) {
...
+	_, err := tracker.StartRequest(&attr1)
+	if err != nil {
+		t.Errorf("Expecting success, got %v", err)
+	}
+
+	oldDict := tracker.currentDictionary
+	oldBag := copyBag(tracker.contexts[0])
+
+	tracker.StartRequest(&attr2)
+	_, err = tracker.StartRequest(&attr2)
```
mandarjog: a Q: Does the `checkPreconditions` fail here? it seems repeatedly starting request with the same `attr2` should be idempotent. why does the 2nd one fail?

geeknoid: It's calling StartRequest(attr1) which works, followed by StartRequest(attr2) which fails because of a bad attribute index. The failure is indeed discovered within the checkPrecondition method.

mandarjog: looks lie its calling it 2 times.
```
tracker.StartRequest(&attr2)
_, err = tracker.StartRequest(&attr2)
```

geeknoid:
Oh, I hadn't seen that. Cut & paste snafu, fixed now.

PR #351, opened 2017-03-09
Update the mixer to depend on the latest changes in istio/api.

ZackButcher:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/351)

### DIFF
--- a/BUILD.api
+++ b/BUILD.api
@@ -55,14 +55,11 @@ gogoslick_proto_library(
     name = "mixer/v1/config",
     importmap = {
         "google/protobuf/struct.proto": "github.com/gogo/protobuf/types",
-        "mixer/v1/config/descriptor/*.proto": "istio.io/api/mixer/v1/config/descriptor",
-        "mixer/v1/config/descriptor/label_descriptor.proto": "istio.io/api/mixer/v1/config/descriptor",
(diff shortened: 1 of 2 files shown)


BUILD.api
```diff
@@ -40,15 +40,34 @@ gogoslick_proto_library(
...
+    "mixer/v1/config/descriptor/principal_descriptor.proto",
+    "mixer/v1/config/descriptor/quota_descriptor.proto",
+    "mixer/v1/config/descriptor/value_type.proto",
+]
+
 gogoslick_proto_library(
     name = "mixer/v1/config",
     importmap = {
         "google/protobuf/struct.proto": "github.com/gogo/protobuf/types",
+        "mixer/v1/config/descriptor/*.proto": "istio.io/api/mixer/v1/config/descriptor",
```
douglas-reid: do you want the `*.proto` here? it seems strange to both wildcard it and explicitly list the protos. 

was this meant to be `attribute_descriptor.proto`? Or can we remove it altogether?

ZackButcher: Yup, it was originally `attribute_descriptor.proto`, I updated to `*.proto` in the hope that I wouldn't have to list them all, but forgot to remove it when that didn't work. Cleaned up this section.

ZackButcher:
We need to mark manager/mixer-pr as not required. I can't figure out how, but it's broken due to a race in the proxy's mixer client and is blocking any of our merges.

PR #360, opened 2017-03-13
Introduce out-of-order async evaluation of API requests.

geeknoid:
- API dispatch is now handled using the goroutine pool.
This means we can accept API calls at a much higher rate
on a given stream, and we now process each call in parallel.
This in turn means that mixer can now complete API calls out
of order for a given stream.

- Allow command-line control of how many goroutines are used
for the adapter worker pool.

- Add the --singleThreaded command-line option which makes the
mixer run all API processing and adapter logic on a single
goroutine. This can be handy for debugging.

- Extend the goroutine pool to support a single-threaded
option. The pool behaves as normal, but doesn't actually use
any goroutines in that case.

- Add Env.ScheduleDaemon for use in adapters. Whereas ScheduleWork
is intended for short duration functions, ScheduleDaemon is intended
for long-running background stuff. When the mixer is running in
single-threaded mode, ScheduleDaemon continues to dispatch real
goroutines in the background, whereas ScheduleWork doesn't.

- Bump code coverage to 100% for pkg/api, pkg/attribute and pkg/status.

The one unfortunate part about dispatching API processing in
parallel is that the protocol-level attribute bag must be fully
copied for every API call, rather than being referenced. For now,
this is good enough. Eventually, we'll want to fancy up the
attribute bag functionality to support a versioned data structure
so that we can avoid the repetitive copying.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/360)

### DIFF
--- a/pkg/adapterManager/manager_test.go
+++ b/pkg/adapterManager/manager_test.go
@@ -186,8 +186,10 @@ func TestManager(t *testing.T) {
 		if tt.mgrFound {
 			mgr = newFakeMgrReg(tt.wrapper)
 		}
-		m := newManager(r, mgr, mapper, nil, nil)
-		defer m.Shutdown()
+
+		gp := pool.NewGoroutinePool(1, true)
+		agp := pool.NewGoroutinePool(1, true)
+		m := newManager(r, mgr, mapper, nil, gp, agp)
 
 		out := m.Execute(context.Background(), tt.cfg, attrs, nil)
 		errStr := out.Message()
@@ -219,6 +221,8 @@ func TestManager(t *testing.T) {
 			t.Errorf("[%d] Unexpected mgr.NewAspect call %d", idx, fmgr.called)
 		}
 
+		gp.Close()
+		agp.Close()
 	}
 }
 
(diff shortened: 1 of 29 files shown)


pkg/api/handler_test.go
```diff
@@ -51,15 +52,98 @@ func TestAspectManagerErrorsPropagated(t *testing.T) {
...
 
-	o := h.execute(context.Background(), attribute.NewManager().NewTracker(), &mixerpb.Attributes{}, aspect.CheckMethod, nil)
+	bag, _ := attribute.NewManager().NewTracker().ApplyAttributes(&mixerpb.Attributes{})
+	o := h.execute(context.Background(), bag, aspect.CheckMethod, nil)
 	if o.Status.Code != int32(rpc.INTERNAL) {
 		t.Errorf("execute(..., invalidConfig, ...) returned %v, wanted status with code %v", o.Status, rpc.INTERNAL)
 	}
 }
 
+func TestHandler(t *testing.T) {
```
ZackButcher: Can you change this to a table driven test? It's reasonably close already.

geeknoid:
PTAL.

pkg/adapterManager/manager_test.go
```diff
@@ -383,11 +409,14 @@ func TestExecute_TimeoutWaitingForResults(t *testing.T) {
...
 		found: true,
 	}
-	gp := pool.NewGoroutinePool(128)
-	defer gp.Close()
+
+	gp := pool.NewGoroutinePool(128, true)
+	gp.AddWorkers(32)
+
+	agp := pool.NewGoroutinePool(128, true)
 	gp.AddWorkers(32)
```
ZackButcher: agp.AddWorkers

PR #369, opened 2017-03-14
Start plumbing in Descriptors

ZackButcher:
This PR adds a `DescriptorFinder` interface and implementation, and adds it to the config manager. Subsequent PRs will add it as a parameter to the aspect's `ValidateConfig` methods. It's likely I'll need to restructure how/where the descriptor finder is created and passed around in the config validator and manager as a result of those subsequent PRs, I only touched them in this PR because our linters complained about dead code if I didn't create a descriptor finder somewhere outside of a test.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/369)


pkg/config/descriptors/descriptors.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	dpb "istio.io/api/mixer/v1/config/descriptor"
+	pb "istio.io/mixer/pkg/config/proto"
+)
+
+// Finder describes anything that can provide a view into the config's descriptors by name and type.
+type Finder interface {
+	// GetLog retrieves the log descriptor named `name`
+	GetLog(name string) (*dpb.LogEntryDescriptor, bool)
```
douglas-reid: should these be named `Find*` to match `Finder` interface?  or should we name this interface `Getter` ?

mandarjog: `Finder` and `Find*` is better.

pkg/config/descriptors/descriptors.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// GetMetric retrieves the metric descriptor named `name`
+	GetMetric(name string) (*dpb.MetricDescriptor, bool)
```
mandarjog: When the return type is a `pointer` 2nd `bool` return is redundant. 
you can always test the pointer for `nil`

pkg/config/descriptors/descriptors.go
```diff
@@ -0,0 +1,107 @@
...
+		logs:               logs,
+		metrics:            metrics,
+		monitoredResources: monitoredResources,
+		principals:         principals,
+		quotas:             quotas,
+	}
+}
+
+func (d finder) GetLog(name string) (*dpb.LogEntryDescriptor, bool) {
+	l, found := d.logs[name]
```
mandarjog: it simplifies all this code, just return the pointer.

PR #371, opened 2017-03-14
Add the descriptor finder to ConfigValidator.ValidateConfig

ZackButcher:
Add `descriptors.Finder` as a parameter to `ConfigValidator.ValidateConfig` and update everything. Follow on PRs will actually make use of the finder.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/371)


ZackButcher:
Don't bother reviewing this yet. We want to expose descriptors to the managers, but not the adapters. Today both managers and adapters implement the same set of interfaces so this isn't possible. I'm going to split the interfaces into manager and adapter versions so we can hide this kind of detail from adapters.

ZackButcher:
Ditching this in favor of the PR currently out for review.

PR #375, opened 2017-03-15
Boost coverage for pkg/adapter to 100%

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/375)

### DIFF
--- a/pkg/adapter/BUILD
+++ b/pkg/adapter/BUILD
@@ -28,6 +28,8 @@ go_test(
     name = "small_tests",
     size = "small",
     srcs = [
+        "applicationLogs_test.go",
+        "builder_test.go",
         "configError_test.go",
         "metrics_test.go",
     ],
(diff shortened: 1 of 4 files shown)


pkg/adapter/BUILD
```diff
@@ -28,6 +28,8 @@ go_test(
     name = "small_tests",
     size = "small",
     srcs = [
+        "applicationLogs_test.go",
```
douglas-reid: aside: this reminds me that we probably should have all lowercase filenames (so, `applicationlogs.go` and `applicationlogs_test.go`, here).

pkg/adapter/applicationLogs_test.go
```diff
@@ -0,0 +1,72 @@
...
+		Info,
+		Notice,
+		Warning,
+		Error,
+		Critical,
+		Alert,
+		Emergency,
+	}
+
+	for i, s := range severities {
```
douglas-reid: could we just iterate through the severityMap here (instead of possibly getting out of sync in our slice -- if the intent is to check all values)?

geeknoid:
Done. PTAL.

PR #384, opened 2017-03-15
Address a number of Gogland code inspection issues so I don't need to look  at so many squiggles.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/384)

### DIFF
--- a/cmd/server/inventory.go
+++ b/cmd/server/inventory.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"fmt"
 	"sort"
 
 	"github.com/ghodss/yaml"
@@ -27,7 +26,7 @@ import (
 	"istio.io/mixer/pkg/aspect"
 )
 
-func adapterCmd(errorf errorFn) *cobra.Command {
+func adapterCmd(outf outFn, errorf errorFn) *cobra.Command {
 	adapterCmd := cobra.Command{
 		Use:   "inventory",
 		Short: "Inventory of available adapters and aspects in the mixer",
@@ -37,7 +36,7 @@ func adapterCmd(errorf errorFn) *cobra.Command {
 		Use:   "adapter",
 		Short: "List available adapter builders",
 		Run: func(cmd *cobra.Command, args []string) {
-			err := listBuilders()
+			err := listBuilders(outf)
 			if err != nil {
 				errorf("%v", err)
 			}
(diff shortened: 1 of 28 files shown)


cmd/server/inventory.go
```diff
@@ -96,7 +96,7 @@ func listBuilders() error {
 
 func printConfigValidator(v pkgadapter.ConfigValidator) {
 
-	fmt.Printf("Params: \n")
+	fmt.Print("Params: \n")
```
douglas-reid: should we be using fmt.Print at all here? aren't we supposed to be on glog?

pkg/aspect/quotasManager_test.go
```diff
@@ -193,7 +194,8 @@ func TestQuotaWrapper_Execute(t *testing.T) {
 		{goodMd, 1, nil, false, errEval, make(map[string]o), "expected"},
 		{goodMd, 1, nil, false, labelErrEval, make(map[string]o), "expected"},
 		{goodMd, 1, nil, false, goodEval, map[string]o{"request_count": {1, []string{"source", "target"}}}, ""},
-		{goodMd, 0, fmt.Errorf("alloc-forced-error"), false, goodEval, map[string]o{"request_count": {1, []string{"source", "target"}}}, "alloc-forced-error"},
+		{goodMd, 0, errors.New("alloc-forced-error"), false, goodEval, map[string]o{"request_count": {1, []string{"source", "target"}}},
```
douglas-reid: nit: this looks weird with the new line break.  should we move the `map[string]o{"request_count": {1, []string{"source", "target"}}}` up (as it is used in another test case, too)?

cmd/server/inventory.go
geeknoid: The two front-ends(mixc and mixs) use normal stdout stuff. The core mixer uses glog.

pkg/aspect/quotasManager_test.go
geeknoid: Done.

geeknoid:
PTAL.

cmd/server/inventory.go
douglas-reid: Then should we not use glog in the main.go for server (https://github.com/istio/mixer/blob/master/cmd/server/main.go#L21)?

We probably want consistency there. And, we should some mention of the exception (with reasoning) in our [docs](https://github.com/istio/mixer/blob/master/doc/dev/conventions.md#logging-conventions).

geeknoid:
Consolidated out stdout & stderr are handled in mixc and mixs. It's all consistent now, and if we decide to change the output to something else, it's all centralized. That can be a separate issue if we find it useful.

PTAL.

PR #392, opened 2017-03-16
Improve expr performance

mandarjog:
initial perf: 1600 ns/op
new: 450 ns/op

Use unmanaged functions.
  Previously framework was doing a lot of common work and giving
  elementary types to Funcs, but it was not performant.
  Now writing a func is more involved, but it is more performant.
  Now funcs receive []*Expression

convert all existing funcs to unmanaged.

Issues with managed funcs:
1. involved creating a lot of objects. Could be resolved by using a pool of slices,
   but now removed the need for it.
2. Short circuit functions are easier with unmanaged.  { A && B && C } should return false if A is false and not evaluate B and C.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/392)

### DIFF
--- a/pkg/expr/benchmark_test.go
+++ b/pkg/expr/benchmark_test.go
@@ -15,7 +15,6 @@
 package expr
 
 import (
-	"errors"
 	"strings"
 	"testing"
 
@@ -45,21 +44,20 @@ type exprFunc func(a attribute.Bag) (bool, error)
 // a == 20 || request.header["host"] == 50
 
 func dff(a attribute.Bag) (bool, error) {
-	var aa int64
+	var v interface{}
 	var b bool
-	var s map[string]string
-	if aa, b = a.Int64("a"); !b {
+	if v, b = a.Get("a"); !b {
 		return false, fmt.Errorf("a not found")
 	}
+	aa := v.(int64)
 	if aa == 20 {
 		return true, nil
 	}
 
-	s, b = a.StringMap("request.header")
-	if !b {
+	if v, b = a.Get("request.header"); !b {
 		return false, fmt.Errorf("a not found")
 	}
-	ss := s["host"]
+	ss := v.(map[string]string)["host"]
 	if ss == "abc" {
 		return true, nil
 	}
--- a/pkg/expr/benchmark_test.go
+++ b/pkg/expr/benchmark_test.go
@@ -59,8 +59,7 @@ func dff(a attribute.Bag) (bool, error) {
 	if !b {
 		return false, fmt.Errorf("a not found")
 	}
-	var ss string
-	ss = s["host"]
+	ss := s["host"]
 	if ss == "abc" {
 		return true, nil
 	}
--- a/pkg/expr/benchmark_test.go
+++ b/pkg/expr/benchmark_test.go
@@ -15,7 +15,7 @@
 package expr
 
 import (
-	"fmt"
+	"errors"
 	"strings"
 	"testing"
 
@@ -48,7 +48,7 @@ func dff(a attribute.Bag) (bool, error) {
 	var v interface{}
 	var b bool
 	if v, b = a.Get("a"); !b {
-		return false, fmt.Errorf("a not found")
+		return false, errors.New("a not found")
 	}
 	aa := v.(int64)
 	if aa == 20 {
@@ -56,7 +56,7 @@ func dff(a attribute.Bag) (bool, error) {
 	}
 
 	if v, b = a.Get("request.header"); !b {
-		return false, fmt.Errorf("a not found")
+		return false, errors.New("a not found")
 	}
 	ss := v.(map[string]string)["host"]
 	if ss == "abc" {
--- a/pkg/expr/benchmark_test.go
+++ b/pkg/expr/benchmark_test.go
@@ -15,11 +15,11 @@
 package expr
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
 	"istio.io/mixer/pkg/attribute"
-	"fmt"
 )
 
 // run micro benchmark using expression evaluator.
@@ -41,8 +41,9 @@ ok  	istio.io/mixer/pkg/expr	10.049s
 
 type exprFunc func(a attribute.Bag) (bool, error)
 
-// a == 20 || request.header["host"] == 50
-
+// dff implements `a == 20 || request.header["host"] == "abc"`
+// this is done so we can compare Expr processed expression with
+// raw direct golang performance.
 func dff(a attribute.Bag) (bool, error) {
 	var v interface{}
 	var b bool
@@ -77,7 +78,7 @@ func benchmarkExpression(b *testing.B, stype string) {
 	exprStr := `a == 20 || request.header["host"] == "abc"`
 	exf, _ := Parse(exprStr)
 
-	fmt.Printf("%s\n", exf.String())
+	b.Logf("%s\n", exf.String())
 	fm := FuncMap()
 	tests := []struct {
 		name   string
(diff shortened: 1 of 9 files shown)


pkg/expr/expr.go
```diff
@@ -105,20 +105,30 @@ func (e *Expression) TypeCheck(attrs AttributeDescriptorFinder, fMap map[string]
 	return e.Fn.TypeCheck(attrs, fMap)
 }
 
+// Eval returns value of the contained variable or error
+func (v *Variable) Eval(attrs attribute.Bag) (interface{}, error) {
+	val, ok := attribute.Value(attrs, v.Name)
```
douglas-reid: nit: this could be:
```
if val, ok := attribute.Value(attrs, v.Name); ok {
  return val, nil
}
return nil, fmt.Errorf("...", v.Name)
```

pkg/expr/func.go
```diff
@@ -82,27 +87,44 @@ func newNEQ() Func {
...
-		// s0 is string and s1 is not
-		return false
+	return res, nil
+}
+
+func (f *eqFunc) call(args0 interface{}, args1 interface{}) bool {
+	switch s0 := args0.(type) {
+	default:
+		return reflect.DeepEqual(args0, args1)
+	case bool:
```
douglas-reid: do we ever imagine different behaviors for bool, int64, and float64?  if not, should we collapse the cases to be:
```
switch ... {
case bool, int64, float64:
   return args0 == args1
...
}
```

pkg/expr/func.go
```diff
@@ -177,17 +213,17 @@ func newOR() Func {
...
 
-// Call selects first non empty argument
+// Call selects first non empty argument / non erroneous
 // may return nil
-func (f *orFunc) Call(args []interface{}) interface{} {
-	if args[0] == nil {
-		return args[1]
+func (f *orFunc) Call(attrs attribute.Bag, args []*Expression, fMap map[string]FuncBase) (interface{}, error) {
+	for _, arg := range args {
+		ret, _ := arg.Eval(attrs, fMap)
```
douglas-reid: nit: collapse to:
```
if ret, _ := arg.Eval(attrs, fMap); ret != nil {
  return ret, nil
}
```

pkg/expr/eval_test.go
```diff
@@ -149,13 +178,44 @@ func TestGoodEval(tt *testing.T) {
...
+					"X-FORWARDED-HOST": "aaa",
+				},
+				"headername": "X-FORWARDED-HOST",
+			},
+			true, "",
+		},
 	}
 
 	for idx, tst := range tests {
+		if tst.src == "(x == 20 && y == 10) || x == 3" {
```
douglas-reid: this seems like a strange block to have in the tests.  Is this intentional?

pkg/expr/benchmark_test.go
```diff
@@ -130,21 +130,26 @@ func BenchmarkExpression(b *testing.B) {
...
-			for n := 0; n < bb.N; n++ {
-				_, _ = tst.df(attrs)
-			}
-		})
+		if stype == "AST" {
+			ii, err := tst.ex.Eval(attrs, fm)
+			assertNoError(err, tst.err, ii, tst.result, b)
+			b.Run(tst.name+"AST", func(bb *testing.B) {
+				for n := 0; n < bb.N; n++ {
+					_, _ = tst.ex.Eval(attrs, fm)
```
douglas-reid: why even included the `_, _ :=` here? to make the linters happy?

pkg/expr/benchmark_test.go
```diff
@@ -24,54 +24,61 @@ import (
...
 
-	exprConstant := "true"
-	ex2, _ := Parse(exprConstant)
-	df2 := func(a attribute.Bag) (bool, error) { return true, nil }
+func benchmarkExpression(b *testing.B, stype string) {
+	success := "_SUCCESS_"
+	exprStr := `a == 20 || request.header["host"] == "abc"`
+	exf, _ := Parse(exprStr)
 
+	fmt.Printf("%s\n", exf.String())
```
douglas-reid: do you want `b.LogF` instead?

pkg/expr/benchmark_test.go
```diff
@@ -24,54 +24,61 @@ import (
...
-			return true, nil
-		}
-		return false, nil
+	s, b = a.StringMap("request.header")
+	if !b {
+		return false, fmt.Errorf("a not found")
 	}
-	exf, _ := Parse(exprStr)
+	ss := s["host"]
+	if ss == "abc" {
```
douglas-reid: the choice of "abc" here seems a bit magical.  can we add some comments to help understanding what this function is doing and why these values are important?

mandarjog: added comment at the top.

pkg/expr/benchmark_test.go
mandarjog: done.

pkg/expr/benchmark_test.go
mandarjog: yes.

pkg/expr/eval_test.go
mandarjog: leftover from debugging.

pkg/expr/expr.go
mandarjog: done.

pkg/expr/func.go
mandarjog: done.

pkg/expr/func.go
mandarjog: I like the collapsed thing, when the return value is important; but the call gets lost.
This way the *call* is important.

mandarjog:
comments addressed PTAL.

pkg/expr/benchmark_test.go
```diff
@@ -24,54 +24,61 @@ import (
...
 
-// a == 20 || request.header["host"] == 50
+// dff implements `a == 20 || request.header["host"] == "abc"`
+// this is done so we can compare Expr processed expression with
+// raw direct golang performance.
+func dff(a attribute.Bag) (bool, error) {
+	var v interface{}
+	var b bool
+	if v, b = a.Get("a"); !b {
+		return false, fmt.Errorf("a not found")
```
geeknoid: replace fmt.Errorf with errors.New

pkg/expr/benchmark_test.go
```diff
@@ -24,54 +24,61 @@ import (
...
+	}
+	aa := v.(int64)
+	if aa == 20 {
+		return true, nil
+	}
+
+	if v, b = a.Get("request.header"); !b {
+		return false, fmt.Errorf("a not found")
```
geeknoid: Use errors.New

pkg/expr/func.go
```diff
@@ -19,10 +19,11 @@ import (
...
 
 	config "istio.io/api/mixer/v1/config/descriptor"
+	"istio.io/mixer/pkg/attribute"
 )
 
-// Func defines the interface that every expression function provider must implement.
-type Func interface {
+// FuncBase defines the interface that every expression function must implement.
+type FuncBase interface {
 	// Name uniquely identified the function.
```
geeknoid: identified -> identifies

pkg/expr/func.go
```diff
@@ -31,16 +32,21 @@ type Func interface {
 
 	// ArgTypes specifies the argument types in order expected by the function.
 	ArgTypes() []config.ValueType
+}
 
-	// NullArgs specifies if the function accepts null args
-	AcceptsNulls() bool
+// Func implements a function call.
+// It needs to know details about Expressions and attribute bag.
+// Func is performant.
```
geeknoid: What does "Func is performant" mean? Seems like an odd statement in an API comment.

PR #394, opened 2017-03-17
Major revamp of pkg/attribute.

geeknoid:
- Removed all strong typing. Attribute bags now just hold a uniform map
of name/value pairs with all values being of type interface{}. It turns
out the strong typing wasn't even used anywhere except in tests.
Attribute typing is instead provided through AttributeDescriptors at a
higher level.

- Combined rootBag into mutableBag since the split was no longer useful.

- Turned the symbol MutableBag from an interface to a struct. This is so
the implementation of cross-bag functions can take advantage of implementation
details of the bag rather than just dealing with interfaces.

- Added support for merging bags, which will be used with the fork-join attribute
mutation model we'll be adding to the mixer.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/394)


pkg/attribute/bag.go
```diff
@@ -16,160 +16,33 @@ package attribute
...
-	// BoolKeys returns a snapshot of all keys corresponding to bool attributes this bag knows about.
-	BoolKeys() []string
-
-	// Time returns the named attribute if it exists.
-	Time(name string) (time.Time, bool)
-
-	// TimeKeys returns a snapshot of all keys corresponding to time attributes this bag knows about.
-	TimeKeys() []string
+	// Get returns an attribute value.
+	Get(name string) (value interface{}, found bool)
```
douglas-reid: non-blocking: would `Value(attribute string)` be clearer ?

pkg/attribute/mutableBag.go
```diff
@@ -15,330 +15,347 @@
...
+				log.WriteString("    <empty>\n")
+			}
+		}
 	}
 
-	mb.Unlock()
-}
+	if log != nil {
+		if log.Len() > 0 {
+			glog.Infof("Updatiing attribute bag %d:\n%s", mb.id, log.String())
```
douglas-reid: nit: s/Updatiing/Updating/

pkg/attribute/mutableBag.go
```diff
@@ -15,330 +15,347 @@
...
+			m[dictionary[k2]] = v2
+		}
+
+		if log != nil {
+			if len(m) > 0 {
+				for k2, v2 := range m {
+					log.WriteString(fmt.Sprintf("    %s:%s\n", k2, v2))
+				}
+			} else {
+				log.WriteString("    <empty>\n")
```
douglas-reid: aside: all the "\n"s in this file make me wonder if we should wrap the log.WriteString in some sort of helper method that always appends a new line...

pkg/attribute/tracker_test.go
```diff
@@ -142,113 +135,79 @@ func TestTracker_ApplyAttributes(t *testing.T) {
...
-		v1, _ := b1.Bytes(k)
-		v2, _ := b2.Bytes(k)
 
-		if len(v1) != len(v2) {
-			return false
-		}
-
-		for i := 0; i < len(v1); i++ {
-			if v1[i] != v2[i] {
+		switch t1 := v1.(type) {
```
douglas-reid: is reflect.DeepEquals not sufficient here?

pkg/attribute/mutableBag.go
geeknoid: Done.

pkg/attribute/tracker_test.go
geeknoid: Unfortunately not, this only deals with the attributes while bags can contain other fields too.

pkg/attribute/bag.go
geeknoid: I like the Get/Set symmetry.

mandarjog:
Can we have a simple way to do `func CreateBag(m map[string]interface{}) attribute.Bag` 
for unit tests the protcol details are irrelevant.

geeknoid:
Sure. I'll add that to my next PR.

PR #402, opened 2017-03-17
Weave through support for response attributes.

geeknoid:
This does all the plumbing, but doesn't actually implement the logic
needed to produce the mixerpb.Attributes message to send to the
mixer. I'll do that next.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/402)

### DIFF
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -120,8 +120,13 @@ func newManager(r builderFinder, m map[aspect.Kind]aspect.Manager, exp expr.Eval
 }
 
 // Execute iterates over cfgs and performs the actions described by the combined config using the attribute bag on each config.
-func (m *Manager) Execute(ctx context.Context, cfgs []*config.Combined, attrs attribute.Bag, ma aspect.APIMethodArgs) aspect.Output {
+func (m *Manager) Execute(ctx context.Context, cfgs []*config.Combined,
+	requestBag *attribute.MutableBag, responseBag *attribute.MutableBag, ma aspect.APIMethodArgs) aspect.Output {
 	numCfgs := len(cfgs)
+
+	// TODO: consider implementing a fast path when there is only a single config.
+	//       we don't need to schedule goroutines, we could use the incoming attribute
+	//       bags without needing children & merging, etc.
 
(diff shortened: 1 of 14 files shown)


pkg/adapterManager/manager.go
```diff
@@ -149,6 +159,21 @@ func (m *Manager) Execute(ctx context.Context, cfgs []*config.Combined, attrs at
 		}
 	}
 
+	bags := make([]*attribute.MutableBag, numCfgs)
+	for i, r := range results {
+		bags[i] = r.responseBag
+	}
+
+	err := responseBag.Merge(bags)
```
ZackButcher: You can combine these to `if err := responseBag.Merge(bags); err != nil { ... }`

pkg/adapterManager/manager.go
```diff
@@ -149,6 +159,21 @@ func (m *Manager) Execute(ctx context.Context, cfgs []*config.Combined, attrs at
 		}
 	}
 
+	bags := make([]*attribute.MutableBag, numCfgs)
```
ZackButcher: Sucks that we have to alloc a temp structure for the bags. Is it worth updating `Merge` to take a single bag at a time and we handle the looping ourselves here? We don't need to alloc anything other than the channel, but we lose the chance to verify there are no conflicting keys before we start merging. However, aspects should declare their response attributes in their config, so I'd argue that guaranteeing no conflicts could be pushed to config validation.

pkg/adapterManager/manager.go
geeknoid: Done.

pkg/adapterManager/manager.go
geeknoid: I agree that in the long run the conflict validation won't be useful, we can probably catch all this stuff up front at config load time rather than runtime. Until we get there though, I'd like to avoid surprises.

As for the issue of wasting an allocation, this array could easily be clumped into a per-request object pool as mentioned in the TODO statement a few lines up. I expect that once we start measuring end-to-end perf, we'll systematically identify and eliminate all the allocations in the main runtime paths. Anyway, I'll add a TODO here too.

Issue #393, opened 2017-03-16
Internal Mixer protocol errors should not be leaked to app client


kyessenov:
Mixer client is passing mixer errors as-is to the application client.
I think Mixer/or the client should define an external message in addition to the internal error details.

douglas-reid:
Adding some detail: Mixer sends back a status message, with a code and a message. At the moment, the message includes full details of the error (see: https://github.com/istio/mixer/blob/master/pkg/api/grpcServer.go#L107).

We should think about returning a better, developer-facing error message, with the current level of details passed via the details field (using google.rpc.DebugInfo or google.rpc.BadRequest messages).

It will still be up to the mixerclient to decide how best to pass that info on to the application.

PR #404, opened 2017-03-17
Add initial support for use of details field in rpc.Status messages.

douglas-reid:
This is the first step in addressing #393.

It addresses the `attribute_update`-related messages that are 
returned from Mixer. These were the messages that had been 
leaking through mixerclient out to application clients.

Mixerclient will still need to figure out how best to translate this
messaging to something more appropriate for applications to see
(they should not care about issues with attributes, etc.).

The PR adds support for creating BadRequest messages, which is 
one of the standard rpc error types. Special handling is added for
go-multierrors (to generate a list of field violation messages).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/404)

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -16,7 +16,10 @@
 package status
 
 import (
+	"github.com/gogo/protobuf/proto"
+	"github.com/gogo/protobuf/types"
 	rpc "github.com/googleapis/googleapis/google/rpc"
+	me "github.com/hashicorp/go-multierror"
 )
 
 // OK represents a status with a code of rpc.OK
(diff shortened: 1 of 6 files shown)


pkg/status/status.go
```diff
@@ -71,3 +74,33 @@ func WithDeadlineExceeded(message string) rpc.Status {
 func IsOK(status rpc.Status) bool {
 	return status.Code == int32(rpc.OK)
 }
+
+// InvalidWithDetails builds a google.rpc.Status proto with the provided
```
geeknoid: Please add tests for new methods.

douglas-reid: Done.  Totally missed the status_tests.

PR #405, opened 2017-03-18
Move repositories to a bzl file

qiwzhang:
This bzl file can be shared by other repo if it wants to build mixer.
Istio/proxy is building mixer to use its api package for its integration test.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/405)


kyessenov:
We need a more systematic approach to the API proto generation. I think it is not a good idea to introduce a direct dependency from the proxy on the mixer code. I am also not a fan of the "fake" istio.io/api repository that mixer init script generates. I feel like the BUILD file for API repo should belong to API repo, or some other shared repository ("toolbox", "common").

Issue #403, opened 2017-03-17
Commit b83d5b2 breaks traffic flow for echo app


andraxylia:
Mixer image from docker labeled 2017-03-17-18.02.40 allows the traffic to pass; latest images gives 503 code

None of the newer mixer images open port 42422, although there are no errors in the logs.

douglas-reid:
@geeknoid can you please take a look at what is happening here?

geeknoid:
Zack, I think you have more context than me on this...

geeknoid:
Oh, it's attribute-related. I'll take a look.

geeknoid:
Turns out this broke when I pushed in the OOO return functionality. It's a state capture problem around spawning goroutines. I'll have a PR out tonight with appropriate fixes.

andraxylia:
There is another problem with metrics for 42422 not being available. Can this be related?

douglas-reid:
I don't think it is related. @ZackButcher something is wrong with the configuration of logs aspects in testdata/serviceconfig.yml.

PR #408, opened 2017-03-18
Fix bug preventing multiple aspects from being executed for a given request

geeknoid:
This fixes issue #403.

I just spent 1/2 hour trying to figure out how to test this particular edge case and gave up. As I expect to be rewriting large parts of the adapter manager in order to introduce full language support, I decided it wasn't worth it spending the time to test this case for the time being.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/408)

### DIFF
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -129,9 +129,10 @@ func (m *Manager) Execute(ctx context.Context, cfgs []*config.Combined, attrs at
 
 	// schedule all the work that needs to happen
 	for _, cfg := range cfgs {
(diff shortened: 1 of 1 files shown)


geeknoid:
I also thought "What!? None of those stupid linters told me about that?". Turns out, it's only a problem if the function runs async, it's totally fine if the function is used as a synchronous callback. And my ScheduleWork method obfuscates the fact the function is being dispatched asynchronously.

Maybe some linter would discover the issue if the code contained a direct inline "go" statement to indicate asynchronous semantics.

PR #411, opened 2017-03-18
Increase test coverage for pkg/tracing to 100%

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/411)

### DIFF
--- a/pkg/tracing/tracing.go
+++ b/pkg/tracing/tracing.go
@@ -67,16 +67,27 @@ type Tracer struct {
 	ot.Tracer
 
 	enabled bool
+
+	// this indirection exists strictly to enable error injection
+	inject func(sm ot.SpanContext, format interface{}, carrier interface{}) error
 }
 
 // NewTracer wraps the provided tracer and enables tracing.
 func NewTracer(tracer ot.Tracer) Tracer {
-	return Tracer{tracer, true}
+	return Tracer{
+		Tracer:  tracer,
+		enabled: true,
+		inject:  tracer.Inject,
+	}
 }
 
 // DisabledTracer disables tracing, letting methods in tracing.go short-circuit.
 func DisabledTracer() Tracer {
-	return Tracer{ot.NoopTracer{}, false}
+	return Tracer{
+		Tracer:  ot.NoopTracer{},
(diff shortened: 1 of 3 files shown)


pkg/tracing/tracing_test.go
```diff
@@ -113,6 +114,17 @@ func TestPropagateSpan(t *testing.T) {
 	}
 }
 
+func TestPropagateSpanError(t *testing.T) {
+	tracer := NewTracer(mocktracer.New())
+	tracer.inject = func(sm ot.SpanContext, format interface{}, carrier interface{}) error {
+		return errors.New("error")
```
ZackButcher: `t.Errorf`

geeknoid: This is a callback function that needs to return an error, so I think the code is right as written.

ZackButcher: Oh whoops, yup you're correct. Feels odd that nowhere in this test do we have a failure condition - how do we know short of a panic that things are working as intended?

pkg/tracing/tracing.go
```diff
@@ -67,16 +67,27 @@ type Tracer struct {
 	ot.Tracer
 
 	enabled bool
+
+	// this indirection exists strictly to enable error injection
+	inject func(sm ot.SpanContext, format interface{}, carrier interface{}) error
```
ZackButcher: I don't love adding stuff like this just to we can tickle code paths for test coverage.

pkg/tracing/tracing_test.go
geeknoid: The comment at the end of the function says it all...

This is indeed a simple test to ensure that the logging statement in the failure handling code doesn't lead to a crash (which unexercised failure handling path are prone to do)

pkg/tracing/tracing.go
geeknoid: I want us to be at 100% test coverage in the near future. I'm happy to hear of alternative approaches to do that in this case.

I consider this  a pretty cheap and expedient way to get the job done without having to create a whole bunch of mock infrastructure.

PR #412, opened 2017-03-18
Increase robustness of config and report processing

douglas-reid:
This PR addresses several issues uncovered while investing #403. These issues were making the mixer brittle with respect to changing inputs from the proxy (and test client). This PR attempts to make the mixer be more flexible with respect to inputs and to, somewhat, simplify the config.

Items addressed in this PR:

- Restored support for time.Durations in config files. This enables the setting of default values for labels, etc., that should be durations (ex: request_latency).
- Adds missing 'not' (and includes error detail) in the prometheus error strings for metric recording
- Makes access-logs much more tolerant of missing attributes.

RE: Access log tolerance

The currently supported formats (Common Log Format, Combined Log Format) have well-known placeholder values for missing information ("-"). The accessLogsManager already includes logic to handle that issue directly. This means that we don't need to add `| <const>` to each attribute expression in the config file. Rather, we just need to not fail the `Report` call when some values are missing.

In particular, this is useful for dealing with the `timestamp` part of the schema. We do not want to encode a const value as the alternate when `request.time` is missing from the `Report`. Rather, we would like to fallback to `time.Now()`.

This means that a simple request (ex: `$ ./bazel-bin/cmd/client/mixc report "test"`) will result in an access log like: 
```
{"logName":"access_log","labels":{"timestamp":"2017-03-18T13:29:36.003153445-07:00"},"textPayload":"- - - [18/Mar/2017:13:29:36 -0700] \"- - -\" - -"}
```
rather than returning an error to the caller.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/412)


pkg/expr/expr.go
```diff
@@ -150,9 +151,14 @@ func newConstant(v string, vType config.ValueType) (*Constant, error) {
 			return nil, err
 		}
 	default: // string
-		if typedVal, err = strconv.Unquote(v); err != nil {
+		var unquoted string
+		if unquoted, err = strconv.Unquote(v); err != nil {
 			return nil, err
 		}
+		if typedVal, err = time.ParseDuration(unquoted); err == nil {
+			break
```
mandarjog: in the default section `vType` is `config.STRING`, but since you are parsing it as a ParseDuration, add
`vType = config.Duration`

pkg/expr/expr.go
```diff
@@ -150,9 +151,14 @@ func newConstant(v string, vType config.ValueType) (*Constant, error) {
 			return nil, err
 		}
 	default: // string
```
mandarjog: perhaps this comment will help.

// newConstant converts literals recognized by parser (int, float, char, string) to config.ValueType

Several config.ValueTypes are "strings", so they must be parsed separately as those value types, and then the correct vType should be set.

mandarjog: Update expr to support config.Duration should be a separate PR, and should really add *all the missing config types that need to be hand parsed*
like duration, email, etc.

pkg/aspect/accessLogsManager.go
```diff
@@ -117,14 +118,29 @@ func (e *accessLogsWrapper) Close() error {
...
-		return Output{Status: status.WithError(fmt.Errorf("failed to eval labels for log %s with err: %s", e.name, err))}
+
+	labels := make(map[string]interface{}, len(e.labels))
+	for name, exp := range e.labels {
+		v, err := mapper.Eval(exp, attrs)
+		if err == nil {
+			labels[name] = v
+			continue
+		}
+		if name == "timestamp" {
```
ZackButcher: Is there a way we can accomplish the same thing without relying on an explicit name for the label? e.g. `if expr.TypeOf(exp) == time {... }`? I suppose we don't have the correct interfaces to do that just right now (after I add descriptors I plan to expose types for expressions on the Evaluator interface). In theory the user could call this whatever they want.

pkg/aspect/accessLogsManager.go
```diff
@@ -117,14 +118,29 @@ func (e *accessLogsWrapper) Close() error {
...
+			labels[name] = time.Now()
+		}
 	}
 
-	templateVals, err := evalAll(e.templateExprs, attrs, mapper)
-	if err != nil {
-		return Output{Status: status.WithError(fmt.Errorf("failed to eval template expressions for log %s with err: %s", e.name, err))}
+	templateVals := make(map[string]interface{}, len(e.templateExprs))
+	for n, expr := range e.templateExprs {
```
ZackButcher: Can we extract this to a local func so it can be shared between template and label expressions?

pkg/aspect/accessLogsManager.go
douglas-reid: I think you are right that we don't have a good way to do it just yet. I can come back to this when we do. It also is OK-ish, as the "timestamp" bit is hardcoded in the templates at the moment.

ZackButcher: I actually added in typechecking this morning, so what I described is possible now, just not during execute time. AFAIK we don't ever plan on introducing descriptors into the exec path, so it likely will never be possible to do what I described. We'll have to noodle on this a bit to come up with a general solution, but no reason for that to block this PR.

pkg/expr/expr.go
douglas-reid: I agree that this feels like a separate PR. However, we don't need those missing config types yet... and this PR is about restoring functionality that was working before, so I feel like it is OK to include this little bit in this PR.  Sound ok?

pkg/aspect/accessLogsManager.go
douglas-reid: done.

douglas-reid:
PTAL.

pkg/expr/expr.go
mandarjog: With the config.Duration fix, this is good for now.
Can we file a separate Issue for the remaining items?

douglas-reid: done in #450 .

PR #413, opened 2017-03-19
Use a separate repositoes.bzl to share.

qiwzhang:
Istio/proxy will use that file to load and build mixer.

Tried to move some api related dependencies to istio/api repo.  It is not easy to mix both cc_proto with go_proto.  It can be done, but will require time to figure out how cc and go can co-exist.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/413)


qiwzhang:
Close it. Too much trouble to make "go test" work.

PR #415, opened 2017-03-19
Bunch of cleanup work in mixc & mixs.

geeknoid:
- Produce errors when junk extraneous arguments are supplied.

- Use the right error code when exiting with an error.

- Eliminate useless "message" argument that was required for the 'mixc report' command.

- Correctly distinguish between command-line errors vs. operational errors. We don't want
to print the usage line when it's an operational error.

- Introduce the mixcol command which is used to output the collateral files for the mixer CLI commands: markdown doc files, man page files, and bash completion files.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/415)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -131,6 +131,24 @@ new_go_repository(
 )
 
 new_go_repository(
+    name = "com_github_cpuguy83_go_md2man",
+    commit = "648eed146d3f3beacb64063cd0daae908015eebd", # Mar 19, 2017 (no releases)
+    importpath = "github.com/cpuguy83/go-md2man",
+)
+
+new_go_repository(
+    name = "com_github_russross_blackfriday",
+    commit = "35eb537633d9950afc8ae7bdf0edb6134584e9fc", # Mar 19, 2017 (no releases)
+    importpath = "github.com/russross/blackfriday",
+)
+
+new_go_repository(
+    name = "com_github_shurcooL_sanitized_anchor_name",
+    commit = "10ef21a441db47d8b13ebcc5fd2310f636973c77", # Mar 19, 2017 (no releases)
+    importpath = "github.com/shurcooL/sanitized_anchor_name",
+)
+
+new_go_repository(
     name = "com_github_hashicorp_go_multierror",
     commit = "ed905158d87462226a13fe39ddf685ea65f1c11f", # Dec 16, 2016 (no releases)
     importpath = "github.com/hashicorp/go-multierror",
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -15,122 +15,23 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 
-	"github.com/spf13/cobra"
+	"istio.io/mixer/cmd/client/cmd"
 )
 
-type rootArgs struct {
-	// attributes is the list of name/value pairs of auto-sensed attributes that will be sent with requests
-	attributes string
-
-	// stringAttributes is the list of name/value pairs of string attributes that will be sent with requests.
-	stringAttributes string
-
-	// int64Attributes is the list of name/value pairs of int64 attributes that will be sent with requests.
-	int64Attributes string
-
-	// float64Attributes is the list of name/value pairs of float64 attributes that will be sent with requests.
-	doubleAttributes string
-
-	// boolAttributes is the list of name/value pairs of bool attributes that will be sent with requests.
-	boolAttributes string
-
-	// timestampAttributes is the list of name/value pairs of timestamp attributes that will be sent with requests.
-	timestampAttributes string
-
-	// durationAttributes is the list of name/value pairs of duration attributes that will be sent with requests.
-	durationAttributes string
-
-	// bytesAttributes is the list of name/value pairs of bytes attributes that will be sent with requests.
-	bytesAttributes string
-
-	// stringMapAttributes is the list of string maps that will be sent with requests
-	stringMapAttributes string
-
-	// mixerAddress is the full address (including port) of a mixer instance to call.
-	mixerAddress string
-
-	// enableTracing controls whether client-side traces are generated for calls to the mixer.
-	enableTracing bool
-
-	// # times to repeat the operation
-	repeat int
-}
-
-// A function used for normal output.
-type outFn func(format string, a ...interface{})
-
-// A function used for error output.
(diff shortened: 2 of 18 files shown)


cmd/collateral/main.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2016 Istio Authors
```
mandarjog: 2017

WORKSPACE
```diff
@@ -131,6 +131,24 @@ new_go_repository(
 )
 
 new_go_repository(
```
ZackButcher: Do we use these new repos anywhere? I couldn't spot them in the BUILD files/imports.

geeknoid:
Yes, they're all dependencies that show up as a result of using the
cobra/doc package.

All these extra dependencies is one big reason why I created a separate
mixcol command rather than embedding the collateral output directly in the
main commands. Embedding would have meant that the mixer server would pull
in all this useless logic. Now at least, it's being pulled in only by the
mixcol command and not mixs and mixc.

cmd/client/main.go
```diff
@@ -119,18 +121,20 @@ func withArgs(args []string, outf outFn, errorf errorFn) {
...
-	withArgs(os.Args[1:],
+	rootCmd := GetRootCmd(os.Args[1:],
 		func(format string, a ...interface{}) {
 			fmt.Printf(format, a...)
 		},
 		func(format string, a ...interface{}) {
-			fmt.Fprintf(os.Stderr, format, a...)
-			os.Exit(1)
+			fmt.Fprintf(os.Stderr, format+"\n", a...)
+			os.Exit(-1)
```
douglas-reid: should we just use `log.FatalF` here (and elsewhere)?

cmd/server/BUILD
```diff
@@ -9,7 +9,6 @@ go_library(
         "main.go",
         "server.go",
     ],
-    visibility = ["//visibility:private"],
```
douglas-reid: same question as above. we don't want things depending on this package, correct?

cmd/client/report.go
```diff
@@ -26,31 +26,24 @@ import (
 
 func reportCmd(rootArgs *rootArgs, outf outFn, errorf errorFn) *cobra.Command {
 	return &cobra.Command{
-		Use:   "report <message>...",
```
douglas-reid: +1 !

cmd/collateral/BUILD
```diff
@@ -0,0 +1,32 @@
...
+        "@com_github_opentracing_basictracer//:go_default_library",
+        "@com_github_opentracing_opentracing_go//:go_default_library",
+        "@com_github_opentracing_opentracing_go//ext:go_default_library",
+        "@com_github_spf13_cobra//:go_default_library",
+        "@com_github_spf13_cobra//doc:go_default_library",
+        "@org_golang_google_grpc//:go_default_library",
+    ],
+)
+
+go_binary(
```
douglas-reid: what is `mixcol` ? Should we add some doc somewhere describing the role (client and server are more obvious)? The PR description doesn't mention it, so it seems like we should describe it somewhere.

cmd/client/BUILD
```diff
@@ -11,7 +11,6 @@ go_library(
         "report.go",
         "util.go",
     ],
-    visibility = ["//visibility:private"],
```
geeknoid: Yes, the new mixcol command needs access.

cmd/client/main.go
geeknoid: I want this to look like a command-line tool when you get the normal output, as opposed to seeing log-like output with timestamps and other metadata. That's the main motivation for using normal print calls into of log calls throughout the commands.

cmd/collateral/BUILD
geeknoid: Sorry, the PR description didn't include mixcol, I added that commit as an extra addition. I added the following comment to the source:

// This is a simple command that is used to output the auto-generated collateral
// files for the various mixer CLI commands. More specifically, this outputs
// markdown files and man pages that describe the CLI commands, along with
// bash completion files.

I sent you email over the weekend about how we can run a command within a build that was built by the current build itself. The idea was that when we build mixer, it automatically builds mixcol and then runs mixcol to produce the collateral files as part of the end-to-end build.

geeknoid:
PTAL.

cmd/client/BUILD
douglas-reid: can we limit the visibility to something like `//cmd:__subpackages__`or similar?

cmd/client/main.go
douglas-reid: ok. That's reasonable. I'm confused by the switch from 1 to -1 (where 1 seems to be a fairly standard catch-all for general errors and is the exit code used by log.Fatal).

cmd/client/BUILD
geeknoid: Done.

cmd/client/main.go
geeknoid: I switched to -1 since I saw this in a few cobra examples I read over the weekend. I just did a quick search to see if there is a general prefered return value to indicate failure and nothing emerged from that. 0 means success and != means failure. So -1 or 1 seem the same in the end.

geeknoid:
PTAL.

WORKSPACE
```diff
@@ -131,6 +131,24 @@ new_go_repository(
...
+new_go_repository(
+    name = "com_github_russross_blackfriday",
+    commit = "35eb537633d9950afc8ae7bdf0edb6134584e9fc", # Mar 19, 2017 (no releases)
+    importpath = "github.com/russross/blackfriday",
+)
+
+new_go_repository(
+    name = "com_github_shurcooL_sanitized_anchor_name",
+    commit = "10ef21a441db47d8b13ebcc5fd2310f636973c77", # Mar 19, 2017 (no releases)
+    importpath = "github.com/shurcool/sanitized_anchor_name",
```
douglas-reid: this dep is a bit strange. do we need the cap `L` in the `importpath`?  and should we remove it from the `name` ?

geeknoid:
Make Jenkins happy by using L in the import path to be consistent with the name field for shurcooL.

Also had to refactor how the commands are organized to keep go test happy. It didn't like the fact I was importing programs. So I had to split the importable bits into a distinct library package.

PTAL.

PR #428, opened 2017-03-21
Add quota-related logging

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/428)

### DIFF
--- a/pkg/aspect/quotasManager.go
+++ b/pkg/aspect/quotasManager.go
@@ -21,6 +21,7 @@ import (
 
 	ptypes "github.com/gogo/protobuf/types"
 
+	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


pkg/aspect/quotasManager.go
```diff
@@ -147,18 +154,26 @@ func (w *quotasWrapper) Execute(attrs attribute.Bag, mapper expr.Evaluator, ma A
 
 	var qr adapter.QuotaResult
 
+	glog.Info("Invoking adapter %s for quota %s with amount %d", w.adapter, qa.Definition.Name, qa.QuotaAmount)
```
mandarjog: This should be v(2), it will produce a lot of messages in info mode.

PR #429, opened 2017-03-21
Split the adapter.ConfigValidator interface in two

ZackButcher:
Aspect Managers now need different configuration information at validation time than aspects. Previously both described their ability to validate config via the `adapter.ConfigValidator` interface. This PR creates a new interface, `config.AspectValidator`, which aspect managers use to validate config. We also shuffle some names around (`adapter.Config` rather than `adapter.AspectConfig`, etc).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/429)

### DIFF
--- a/adapter/denyChecker/denyChecker.go
+++ b/adapter/denyChecker/denyChecker.go
@@ -41,7 +41,7 @@ func newBuilder() builder {
 	return builder{adapter.NewDefaultBuilder(name, desc, conf)}
 }
 
-func (builder) NewDenialsAspect(env adapter.Env, c adapter.AspectConfig) (adapter.DenialsAspect, error) {
+func (builder) NewDenialsAspect(env adapter.Env, c adapter.Config) (adapter.DenialsAspect, error) {
 	return newDenyChecker(c.(*config.Params))
 }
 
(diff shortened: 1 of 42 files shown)


ZackButcher:
Adapters still do their own validation (of the global config), so there needs to be some interface to express the ability to do that validation.

geeknoid:
My point is that doesn't adapter.Builder provide the appropriate interface already? It's just a ConfigValidator with a Name(), Description(), and Close() added.

I bet if you did a search and replace of ConfigValidator with adapter.Builder, things would work just fine.

ZackButcher:
Ah, gotcha, I forgot we embed the validator interface in the `Builder` and not in the `Aspect` itself.

I don't have a strong opinion, but am leaning slightly against removing that interface: today the only code that depends on the `adapter.ConfigValidator` interface is `//pkg/config`. The config package doesn't need the full functionality described by `Builder`, it only needs the config validation piece. It seems reasonable to depend on only the smallest bit of functionality we can.

If you feel strongly I can inline the interface into Builder in this PR.

ZackButcher:
In fact, everywhere else that `Builder` is used doesn't need the `ConfigValidator` interface either. The more I think about it the more I think that the interface shouldn't really even be embedded in `Builder` at all, that's just the most convenient place for us to put it to require that adapter authors write that bit of code that we need.

geeknoid:
I'm fine keeping ConfigValidator if you think it adds value.

PR #431, opened 2017-03-21
Add descriptors to the `config.AspectValidator` interface, plumb them through config validation

ZackButcher:
We add a `descriptor.Finder` parameter to the `ValidateConfig` method of the `config.AspectValidator` interface, and wire a `descriptor.Finder` through config validation. Follow on PRs will use the finder to validate configuration in the aspect managers.

This PR also renames the `descriptors` package to `descriptor` to conform to our style guide (https://github.com/istio/mixer/issues/374).

Note this PR is blocked on https://github.com/istio/mixer/pull/429

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/431)

### DIFF
--- a/pkg/aspect/listsManager.go
+++ b/pkg/aspect/listsManager.go
@@ -21,6 +21,7 @@ import (
 	aconfig "istio.io/mixer/pkg/aspect/config"
 	"istio.io/mixer/pkg/attribute"
 	"istio.io/mixer/pkg/config"
+	"istio.io/mixer/pkg/config/descriptor"
 	cpb "istio.io/mixer/pkg/config/proto"
 	"istio.io/mixer/pkg/expr"
 	"istio.io/mixer/pkg/status"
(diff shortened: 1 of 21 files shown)


pkg/aspect/listsManager.go
```diff
@@ -67,7 +68,7 @@ func (listsManager) DefaultConfig() config.AspectParams {
 	}
 }
 
-func (listsManager) ValidateConfig(c config.AspectParams) (ce *adapter.ConfigErrors) {
+func (listsManager) ValidateConfig(c config.AspectParams, _ descriptor.Finder) (ce *adapter.ConfigErrors) {
```
geeknoid: I don't think you need the _ in here...

ZackButcher: If any parameter is named all parameters must be named; since `c` is used in the body we have to have some name for the `Finder` param too.

PR #433, opened 2017-03-21
Fix another bug in the Quota method.

geeknoid:
Given the current config, the quota method can be invoked from the mixc command
with:

   bazel-bin/cmd/client/mixc quota --amount 1 -name RequestCount

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/433)


qiwzhang:
Do I still need to provide --globalConfigFile and --serviceConfigFile flag in addition to these quota flags?

PR #435, opened 2017-03-22
Add redis quota adapter.

chowchow316:
This is the first PR for adding redis quota adapter into mixer, including basic logic how mixer talks to redis. 
Will add unit test and register in inventory later.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/435)


adapter/redisQuota/config/config.proto
```diff
@@ -0,0 +1,35 @@
...
+
+import "google/protobuf/duration.proto";
+
+option go_package = "config";
+
+message Params {
+    // Minimum number of seconds that deduplication is possible for a given operation.
+    google.protobuf.Duration min_deduplication_duration = 1;
+
+    // Redis network address
```
douglas-reid: nit: you might consider dropping `redis_` from all of the fields, as this is a config proto only exposed to the redisQuota package -- and it is clear that Url, SocketType, and PoolSize apply to redis itself.

adapter/redisQuota/driver_impl.go
```diff
@@ -0,0 +1,79 @@
...
+		cp.pool.Put(impl.client)
+	} else {
+		// radix does not appear to track if we attempt to put a connection back with pipelined
+		// responses that have not been flushed. If we are in this state, just kill the connection
+		// and don't put it back in the pool.
+		impl.client.Close()
+	}
+}
+
+func NewconnPoolImpl(redisUrl string, redisSocketType string, redisPoolSize int64) (connPool, error) {
```
douglas-reid: I would just name this `NewConnPool`.

adapter/redisQuota/redisQuota.go
```diff
@@ -0,0 +1,326 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package redisQuota provides a quota implementation with redis as backend.
+// The prerequisite is to have a redis server running.
+
+package redisQuota
```
douglas-reid: The package name should be all lowercase.

We haven't been as vigilant as we should have been early in mixer about this, but we should start fixing this now.

https://blog.golang.org/package-names
> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

adapter/redisQuota/driver.go
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package redisQuota
+
+// Errors that may be raised during config parsing.
+type redisError string
```
douglas-reid: this doesn't seem to be used anywhere? Do we need it now?

adapter/redisQuota/redisQuota.go
```diff
@@ -0,0 +1,326 @@
...
+	return amount, err
+}
+
+func (rq *redisQuota) Close() error {
+	return nil
+}
+
+type quotaFunc func(d *adapter.QuotaDefinition, key string, currentTime time.Time, currentTick int64) (int64, time.Time, time.Duration)
+
+// TODO: extract the common code below between memQuota and redisQuota into a util-like package, need to figure out where to put the package?
```
douglas-reid: interested in other opinions, but I'd rather we do this upfront than leave it as a TODO and possibly have to address drift down the road. In particular, the keys bit seems movable almost immediately.

adapter/redisQuota/driver_impl.go
```diff
@@ -0,0 +1,79 @@
...
+	// set env variable REDIS_SOCKET_TYPE=tcp, REDIS_URL=localhost:6379
```
douglas-reid: did you mean to include this comment?

adapter/redisQuota/redisQuota.go
geeknoid: We had discussions about creating this shared package a while ago. I told Qian I would give her an answer regarding how to factor this out, but I never did. Sorry 'bout that.

So again, where do we factor this code out? It's not really generic in nature, it's just convenient between these two adapters. Should we just create a memQuota/util package and have redisQuota import from there?

geeknoid: Note that we have the same sharing issue around tests. Many of the tests for memQuota would apply equally well to redisQuota. We just need to refactor things to enable this.

In fact, I opened issue #436 this morning to track the creation of generic aspect tests for all the different aspect types.

adapter/redisQuota/driver.go
chowchow316: Removed for now.

adapter/redisQuota/driver_impl.go
chowchow316: No, just removed.

adapter/redisQuota/redisQuota.go
chowchow316: Changed it to all lowercase. Thanks.

adapter/redisQuota/redisQuota.go
chowchow316: My preference for now is to import from memQuota/util. Will do it in subsequent PR.

adapter/redisQuota/driver_impl.go
chowchow316: Done.

adapter/redisQuota/config/config.proto
chowchow316: Done.

adapter/redisquota/driver.go
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package redisquota
+
+// ConnPool interface for a redis connection pool.
+type ConnPool interface {
```
geeknoid: In general, all symbols declared in the redisquota adapter should be internal (start with a lowercase), except those explicitly required (like NewAspect & friends).

So please change to connPool

adapter/redisquota/driver.go
```diff
@@ -0,0 +1,38 @@
...
+	// Get a connection from the pool. Call Put() on the connection when done.
+	Get() (Connection, error)
+
+	// Put a connection back into the pool.
+	Put(c Connection)
+}
+
+// Connection interface for a redis connection.
+type Connection interface {
```
geeknoid: connection

adapter/redisquota/driver_impl.go
```diff
@@ -0,0 +1,85 @@
...
+
+package redisquota
+
+import (
+	"github.com/golang/glog"
+	"github.com/mediocregopher/radix.v2/pool"
+	"github.com/mediocregopher/radix.v2/redis"
+)
+
+// ConnPoolImpl stores the info for redis connection pool.
```
geeknoid: connPoolImpl

adapter/redisquota/driver_impl.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Put is to put a connection c back to the pool.
+func (cp *ConnPoolImpl) Put(c Connection) {
+	impl := c.(*connectionImpl)
+
+	if impl.pending == 0 {
+		cp.pool.Put(impl.client)
+	} else {
+		// radix does not appear to track if we attempt to put a connection back with pipelined
```
geeknoid: radix -> redis

adapter/redisquota/driver.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2017 Google Inc.
```
geeknoid: It's not clear to me that there is any value in using interfaces here instead of directly manipulating the underlying types. This would reduce the overall amount of code, would eliminate some casting, and would likely run faster too).

In other words, instead of having ConnPool and connPoolImpl, you can just have the single struct type connPool and use it in this module.

adapter/redisquota/driver_impl.go
```diff
@@ -0,0 +1,85 @@
...
+		// responses that have not been flushed. If we are in this state, just kill the connection
+		// and don't put it back in the pool.
+		err := impl.client.Close()
+		if err != nil {
```
geeknoid: Combine this line with the previous one:

if err := impl.client.Close(); err != nil {

adapter/redisquota/driver_impl.go
```diff
@@ -0,0 +1,85 @@
...
+			glog.Errorf("Unable to close connection with redis: %v", err)
```
geeknoid: Adapters shouldn't use glog directly. Instead, adapters should be using env.Logger(). The env pointer is passed in to the adapter.

adapter/redisquota/driver_impl.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+}
+
+// NewConnPool creates a new connection to redis in the pool.
+func NewConnPool(redisURL string, redisSocketType string, redisPoolSize int64) (ConnPool, error) {
+	pool, err := pool.New(redisSocketType, redisURL, int(redisPoolSize))
+	return &ConnPoolImpl{pool}, err
```
geeknoid: Should be returning nil, err in the case of an error. So you'll need a conditional here.

adapter/redisquota/driver_impl.go
```diff
@@ -0,0 +1,85 @@
...
+func (ci *connectionImpl) PipeResponse() (response, error) {
+	// assert.Assert(this.pending > 0)
+	ci.pending--
+
+	resp := ci.client.PipeResp()
+	return &responseImpl{resp}, resp.Err
+}
+
+func (ri *responseImpl) Int() int64 {
+	i, _ := ri.response.Int64()
```
geeknoid: Is this ignoring an error return? What would a failure mean here? Should we log it?

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
...
+type keyWorkspace struct {
+	keys []string
+}
+
+// pool of reusable keyWorkspace structs
+var keyWorkspacePool = sync.Pool{New: func() interface{} { return &keyWorkspace{} }}
+
+var (
+	name = "redisQuota"
+	desc = "redis based quotas."
```
geeknoid: Redis-based

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
...
+	conf = &config.Params{
+		MinDeduplicationDuration: &ptypes.Duration{Seconds: 1},
+		Url:        "localhost:6379",
+		SocketType: "tcp",
+		PoolSize:   10,
+	}
+)
+
+const (
+	// See the rollingWindow comment for a description of what this is for.
```
geeknoid: No rollingWindow in this package.

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
...
+// newAspect returns a new aspect.
+func newAspect(env adapter.Env, c *config.Params, definitions map[string]*adapter.QuotaDefinition) (adapter.QuotasAspect, error) {
+	dedupWindow, _ := ptypes.DurationFromProto(c.MinDeduplicationDuration)
+
+	return newAspectWithDedup(env, time.NewTicker(dedupWindow), c, definitions)
+}
+
+// newAspectWithDedup returns a new aspect.
+func newAspectWithDedup(env adapter.Env, ticker *time.Ticker, c *config.Params, definitions map[string]*adapter.QuotaDefinition) (adapter.QuotasAspect, error) {
+	connPool, _ := NewConnPool(c.SocketType, c.Url, c.PoolSize)
```
geeknoid: Discarding error?

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
...
+	rq := &redisQuota{
```
geeknoid: Can just return, no need for the rq temp variable.

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
...
+
+}
+
+func (rq *redisQuota) ReleaseBestEffort(args adapter.QuotaArgs) (int64, error) {
+	amount, _, err := rq.commonWrapper(args,
+		func(d *adapter.QuotaDefinition, key string, currentTime time.Time, currentTick int64) (int64, time.Time, time.Duration) {
+			result := args.QuotaAmount
+			conn, err := rq.redisPool.Get()
+			if err != nil {
+				return 0, time.Time{}, 0
```
geeknoid: We shouldn't be dropping error returns, they must properly be propagated up. So quotaFunc should be expanded to expect an 'error' value and propagate it upwards accordingly.

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
...
+			}
+
+			return result, time.Time{}, 0
+		})
+
+	return amount, err
+}
+
+func (rq *redisQuota) Close() error {
+	return nil
```
geeknoid: This should undo anything done by NewQuotasAspect. Isn't there anything that should be shutdown, like sockets/connections to redis?

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,315 @@
+// Copyright 2017 Google Inc.
```
geeknoid: I don't see in here how expiring vs. non-expiring quotas are handled. How does that work?

adapter/redisquota/redisquota.go
chowchow316: Opened #457 to track this.

adapter/redisquota/driver.go
```diff
@@ -0,0 +1,85 @@
...
+func (c *connection) pipeAppend(cmd string, args ...interface{}) {
+	c.client.PipeAppend(cmd, args...)
+	c.pending++
+}
+
+func (c *connection) pipeResponse() (*response, error) {
+	c.pending--
+
+	resp := c.client.PipeResp()
+	return &response{resp}, resp.Err
```
geeknoid: return nil here in case of error.

adapter/redisquota/driver.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2017 Google Inc.
```
geeknoid: The code in here needs to be made thread-safe. 

The various methods of an adapter instance can be called multiple times concurrently so the driver code can also be as a result.

adapter/redisquota/redisquota.go
```diff
@@ -0,0 +1,319 @@
...
+	ptypes "github.com/gogo/protobuf/types"
+
+	"istio.io/mixer/adapter/redisquota/config"
+	"istio.io/mixer/pkg/adapter"
+	"istio.io/mixer/pkg/pool"
+)
+
+type builder struct{ adapter.DefaultBuilder }
+
+type redisQuota struct {
```
geeknoid: Need some tests for this code. You can try and reuse some of the memQuota tests. If you can figure out how to make the memQuota tests more generic so they can be reused, that'd be great (but not necessary in this PR, if it's too much hassle)

chowchow316: Yes, I will definitely add more unit tests (maybe in next PR). Actually I'm thinking about implementing a mock redis for test, does it make sense to you?

adapter/redisquota/config/config.proto
```diff
@@ -0,0 +1,35 @@
...
+import "google/protobuf/duration.proto";
+
+option go_package = "config";
+
+message Params {
+    // Minimum number of seconds that deduplication is possible for a given operation.
+    google.protobuf.Duration min_deduplication_duration = 1;
+
+    // Redis network address
+    string url = 2;
```
chowchow316: @douglas-reid Are you OK with more descriptive variable names?

adapter/redisquota/redisquota.go
geeknoid: I think creating a mock redis would be useful. I don't know redis, so how hard would this be?

We want to use redis for our config value store for the mixer as a whole. So having a mock redis we could use during testing would be generally useful. @mandarjog 

Probably should check to see if someone's already done a mock redis implementation in Go.

adapter/redisquota/config/config.proto
```diff
@@ -0,0 +1,35 @@
...
+
+    // Socket type for talking to redis
+    string socket_type = 3;
+
+    // Maximum number of idle connections to redis
+    int64 pool_size = 4;
```
chowchow316: Done.

adapter/redisquota/driver.go
chowchow316: Added TODO here.

adapter/redisquota/driver.go
chowchow316: Done.

adapter/redisquota/redisquota.go
chowchow316: Just opened #463 to track this.

chowchow316:
PTAL.
I'd like to check in this PR first, and then use different PRs to resolve issues brought up here.

adapter/redisquota/config/config.proto
```diff
@@ -0,0 +1,35 @@
...
+
+message Params {
+    // Minimum number of seconds that deduplication is possible for a given operation.
+    google.protobuf.Duration min_deduplication_duration = 1;
+
+    // Redis network address
+    string redis_server_url = 2;
+
+    // Socket type for talking to redis
+    string socket_type = 3;
```
geeknoid: Should this be an enum? Or should the comment say something about what this is expected to be?

chowchow316:
It seems that all recent PRs fails jenkins job manager/mixer-pr: https://testing.istio.io/job/manager/job/mixer-pr/.
Is this a required check for merging?

geeknoid:
The only checks that are currently required are cla/google and
mixer/presubmit, the rest aren't currently reliable. From Github, you can
see the required tests that by 'Required' label.

PR #438, opened 2017-03-22
Implement expression type checking and plumb it through config validation

ZackButcher:
Implement a public type checking interface in the expr package. Update the config.AspectValidator interface to use it, and plumb everything through config validation so that expressions can be type checked in the managers.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/438)


pkg/aspect/quotasManager.go
```diff
@@ -117,7 +117,7 @@ func (m *quotasManager) NewAspect(c *cpb.Combined, a adapter.Builder, env adapte
 
 func (*quotasManager) Kind() Kind                         { return QuotasKind }
 func (*quotasManager) DefaultConfig() config.AspectParams { return &aconfig.QuotasParams{} }
-func (*quotasManager) ValidateConfig(config.AspectParams, descriptor.Finder) (ce *adapter.ConfigErrors) {
+func (*quotasManager) ValidateConfig(config.AspectParams, expr.Validator, descriptor.Finder) (ce *adapter.ConfigErrors) {
```
geeknoid: So is the descriptor finder also going to be supplied to the Execute method?

ZackButcher: I don't think it needs to be: descriptors shouldn't be required at request-time, only during config validation and builder creation.

geeknoid: Sorry, I didn't mean Execute, I meant NewAspect.

ZackButcher: Yes, I think you're correct, we'll have to give the `Finder` to `NewAspect`. I'm about to wade into that as I start implementing validation in the manager, so I'll probably add it to the `Builder` interface in my next PR.

PR #440, opened 2017-03-22
Add descriptor.Finder to the aspect.Manager.NewAspect interface method

ZackButcher:
Open question is how to get a descriptor finder into the adapterManager: it's the one place where the config and request paths intersect (otherwise all config loading and validation happens out of band).

I suspect long term the solution to plumbing the `descriptor.Finder` through is fully separating creation of adapters and managers when new config is provided from anything in the request path (issue https://github.com/istio/mixer/issues/272). I'm not sure how we want to handle it in the short term.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/440)

### DIFF
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -277,7 +277,9 @@ func (m *Manager) cacheGet(cfg *configpb.Combined, mgr aspect.Manager, builder a
 
 	// create an aspect
 	env := newEnv(builder.Name(), m.adapterGP)
-	asp, err = mgr.NewAspect(cfg, builder, env)
+
+	// TODO: figure out how we want to plumb the descriptor finder through the API layer and into this manager.
+	asp, err = mgr.NewAspect(cfg, builder, env, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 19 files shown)


ZackButcher:
And I don't really want to gate us getting descriptors into config on resolving https://github.com/istio/mixer/issues/272; I suspect the solution there will require fairly substantial changes.

ZackButcher:
Chatted with @geeknoid, in the short term we're going to update the `config.ChangeListener` interface to supply both the resolver and the descriptor Finder. This lets us plumb descriptors through api.Handler into adapterManager.Manager and from there into the `NewAspect` methods of the managers.

ZackButcher:
PTAL

pkg/adapterManager/manager.go
```diff
@@ -277,7 +278,9 @@ func (m *Manager) cacheGet(cfg *configpb.Combined, mgr aspect.Manager, builder a
 
 	// create an aspect
 	env := newEnv(builder.Name(), m.adapterGP)
-	asp, err = mgr.NewAspect(cfg, builder, env)
+
+	// TODO: figure out how we want to plumb the descriptor finder through the API layer and into this manager.
```
geeknoid: Deprecated comment.

ZackButcher: deleted

Issue #424, opened 2017-03-21
Support mixer versioning


geeknoid:
We need to make it easy for folks to figure out what they're running. So:

- We should generate a version.go file as part of the build which includes the SHA of the latest commit as a const. Perhaps a date string would be handy too.

- When the mixer starts, it should pump its version info (taken from version.go) into its logs.

- mixs should have a 'version' command that prints out the version info.

- We could also consider adding a Version API call to the mixer which returns the version info, and then we could add a version command to mixc to report the version of a running instance.

mandarjog:
This is similar to what manager does. We would like to use a common approach.
https://github.com/istio/manager/tree/master/cmd/version

geeknoid:
Anything else to do here?

douglas-reid:
No. I think this is closable now.

PR #448, opened 2017-03-23
Add version command to mixer server cmd

douglas-reid:
This PR establishes a basic versioning for bazel builds to help
identify mixer builds and track compatibility across istio
components. It introduces a script to extract the date and git
commit SHA for a build. The script is used by bazel to write
status files which are then linked into the new version.go file
via the `linkstamp` rule of the BUILD file.

With these changes, a simple `bazel build //cmd/server:mixs` should
produce a binary that can be executed as follows to retrieve a
build id:

```shell
$ ./bazel-bin/cmd/server/mixs version
build: 2017-03-23-359f709
```
Addresses #424.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/448)


cmd/server/version.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	BuildID = "nil"
+)
+
+func versionCmd() *cobra.Command {
```
mandarjog: I think we want BuildID to be available more generally, and not just in the command line.
It could be *used* by the version command. 

For example when we add a `/healtz`    or `/info` endpoint it needs to report the same info.
I suggest we move the `BuildID` to   `pkg/aspect. BuildID` or similar.

douglas-reid:
@mandarjog PTAL.  I've made the following changes:

- moved version stuff to //pkg/version
- updated scripts to include info on version (from tags) and clean/modified build status
- moved the version cmd from //cmd/server to //cmd/server/cmd

Now, with a basic bazel build, we can run the `mixs version` and get output like: `version: unknown (build: 2017-03-23-dd6d3ce, status: Clean)`

cmd/server/cmd/version.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"istio.io/mixer/pkg/version"
+)
+
+func versionCmd(outf outFn) *cobra.Command {
+	versionCmd := cobra.Command{
+		Use:   "version",
+		Short: "Prints out build version information for Mixer",
```
geeknoid: We generally use mixer as a common noun, so that should be "for the mixer".

cmd/server/cmd/BUILD
```diff
@@ -8,6 +8,7 @@ go_library(
         "inventory.go",
         "root.go",
         "server.go",
+        "version.go",
```
geeknoid: Can we add the exact same version command to mixc so we can tell whether it's compatible with mixs?

cmd/server/cmd/version.go
```diff
@@ -0,0 +1,32 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			outf("%s\n", version.Info)
+		},
+	}
+	return &versionCmd
```
geeknoid: Can just use:

return &cobra.Command{...

cmd/server/cmd/version.go
douglas-reid: done.

cmd/server/cmd/version.go
douglas-reid: done.

cmd/server/cmd/BUILD
douglas-reid: I've now moved the version command into a shared package (along with OutFn and ErrorFn).

douglas-reid:
PTAL.

Changes:
- new //cmd/shared package
- version command added to mixc

cmd/shared/shared.go
```diff
@@ -0,0 +1,40 @@
...
+type OutFn func(format string, a ...interface{})
+
+// ErrorFn is a function used for error output.
+type ErrorFn func(format string, a ...interface{})
+
+// VersionCmd is a command used to print version information.
+func VersionCmd(outf OutFn) *cobra.Command {
+	return &cobra.Command{
+		Use:   "version",
+		Short: "Prints out build version information for the mixer",
```
geeknoid: This is no longer really accurate given the fact this is now shared code...

cmd/shared/shared.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"istio.io/mixer/pkg/version"
+)
+
+// OutFn is a function used for normal output.
+type OutFn func(format string, a ...interface{})
+
+// ErrorFn is a function used for error output.
```
geeknoid: Worth mentioning that calls to this function do not return.

cmd/shared/shared.go
```diff
@@ -0,0 +1,40 @@
...
+package shared
+
+import (
+	"github.com/spf13/cobra"
+
+	"istio.io/mixer/pkg/version"
+)
+
+// OutFn is a function used for normal output.
+type OutFn func(format string, a ...interface{})
```
geeknoid: Given this has been elevated to some shared code rather than being an implementation detail of a command, I wonder if this stuff would be better if we gave the commands an io.Writer instead of a function pointer? It would compose better and would still provide the same flexibility to capture all the output in tests...

douglas-reid: Perhaps, but then I think we want to have something that matches the fatal signature. My latest commit attacks this in a different manner.

cmd/shared/shared.go
douglas-reid: Addressed in a different manner in latest commit.

cmd/shared/shared.go
douglas-reid: Updated.

douglas-reid:
PTAL.

Changes:
- Refactors OutFn and ErrorFn to shared.FormatFn.
- Added shared.Printf and shared.Fatalf impls

geeknoid:
LGTM

PR #452, opened 2017-03-23
Rejigger the layering between pkg/api and pkg/adapterManager.

geeknoid:
- The adapter manager now explicitly exposes Check/Report/Quota methods

- Eliminate the ugly APIMethodArgs/APIMethodResp pattern. Method-specific
parameters can now be passed directly to the adapter manager's new methods.

- Move the definition of APIMethods to pkg/adapterManager and make private since
that's the only place where it's used.

- Deprectaed the Output type. We now just manipulate rpc.Status values throughout.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/452)


geeknoid:
The tests don't yet all pass, but it's just because I didn't finish updating them. I'm pretty confident that the code is generally correct. Anyway, I wanted your eyes on this ASAP since it touches so many things. Please let me know if there's anything to refactor further. Meanwhile, I'll work on finishing the adapter manager test updates.

pkg/adapterManager/manager.go
```diff
@@ -121,9 +100,68 @@ func newManager(r builderFinder, m map[aspect.Kind]aspect.Manager, exp expr.Eval
...
+			cw := wrapper.(aspect.CheckWrapper)
+			return cw.Execute(requestBag, evaluator)
+		})
+}
+
+// Report dispatches to the set of aspects associated with the Report API method
+func (m *Manager) Report(ctx context.Context, requestBag *attribute.MutableBag, responseBag *attribute.MutableBag) rpc.Status {
+	return m.dispatch(ctx, requestBag, responseBag, reportMethod,
+		func(wrapper aspect.Wrapper, evaluator expr.Evaluator) rpc.Status {
+			rw := wrapper.(aspect.CheckWrapper)
```
douglas-reid: shouldn't this be `wrapper.(aspect.ReportWrapper)` ?

pkg/adapterManager/manager.go
```diff
@@ -121,9 +100,68 @@ func newManager(r builderFinder, m map[aspect.Kind]aspect.Manager, exp expr.Eval
...
+	if cfg == nil {
+		// config has not been loaded yet
+		const msg = "Configuration is not yet available"
+		glog.Error(msg)
+		return status.WithInternal(msg)
+	}
+
+	cfgs, err := cfg.Resolve(requestBag, m.methodMap[method])
+	if err != nil {
+		msg := fmt.Sprintf("unable to resolve config: %v", err)
```
douglas-reid: I think we want a simpler message here, something like: `Bad configuration for Mixer` and then use the details in the status to house what could be the multi-error from Resolve.  See: `status.InvalidWithDetails()`. We probably should expand that to support `status.InternalWithDetails()` and a `status.NewDebugInfo()` message.

Right now, these messages will leak through mixerclient to app.

pkg/aspect/manager.go
```diff
@@ -28,50 +29,96 @@ import (
...
+
 	// Wrapper encapsulates a single aspect and allows it to be invoked.
 	Wrapper interface {
 		io.Closer
+	}
 
-		// Execute dispatches to the adapter.
-		Execute(attrs attribute.Bag, mapper expr.Evaluator, ma APIMethodArgs) Output
+	// CheckWrapper encapsulates a single CheckManager aspect and allows it to be invoked.
+	CheckWrapper interface {
```
douglas-reid: I'm not wild about these names (never have been really).  Any reason to not call them CheckExecuter, ReportExecuter, and QuotaExecuter?

pkg/aspect/manager.go
```diff
@@ -28,50 +29,96 @@ import (
...
+	// QuotaManager take care of aspects used to implement the Quota API method
+	QuotaManager interface {
+		Manager
+
+		// NewQuotaWrapper creates a new aspect wrapper given configuration.
+		NewQuotaWrapper(cfg *cpb.Combined, builder adapter.Builder, env adapter.Env, df descriptor.Finder) (QuotaWrapper, error)
+	}
+
 	// Wrapper encapsulates a single aspect and allows it to be invoked.
 	Wrapper interface {
```
douglas-reid: Is this interface needed still?

pkg/aspect/quotasManager.go
```diff
@@ -117,21 +117,19 @@ func (*quotasManager) ValidateConfig(config.AspectParams, expr.Validator, descri
...
 	if !ok {
 		msg := fmt.Sprintf("Unknown quota '%s' requested", qma.Quota)
 		glog.Error(msg)
-		return Output{Status: status.WithInvalidArgument(msg)}
+		return status.WithInvalidArgument(msg), nil
 	}
 
 	labels, err := evalAll(info.labels, attrs, mapper)
 	if err != nil {
 		msg := fmt.Sprintf("Unable to evaluate labels for quota '%s' with err: %s", qma.Quota, err)
```
douglas-reid: as above: any place where the error could be a multi-error, I think we should try to use DebugInfo, etc. and have more user-friendly messages.

pkg/adapterManager/manager.go
```diff
@@ -332,24 +407,40 @@ func (m *Manager) BuilderValidatorFinder() config.AdapterValidatorFinder {
 	}
 }
 
-// MethodMap returns map of APIMethod --> AspectSet.
-func (m *Manager) MethodMap() map[aspect.APIMethod]config.AspectSet {
-	return m.methodMap
+// GetAspects returns a fully constructed manager map.
+func GetAspects(inventory aspect.ManagerInventory) map[aspect.Kind]aspect.Manager {
```
douglas-reid: nit: i'd prefer just `Aspects()`

pkg/adapterManager/manager.go
geeknoid: I've fixed this. Will shortly be adding a new test (in this PR) to catch this.

pkg/adapterManager/manager.go
geeknoid: I'm deferring this for now (this is just copy/pasted code after all). This is tracked by issue #393

pkg/adapterManager/manager.go
geeknoid: Done

pkg/aspect/manager.go
geeknoid: Yep, the base type is used in many places.

pkg/aspect/manager.go
geeknoid: Done, almost. I went with *Executor to avoid my editor claiming a million symbols have typos in them.

All uses of 'wrapper' in this context have been replaced.

geeknoid:
PTAL.

geeknoid:
All review comments addressed, unit tests added.

PTAL.

Note: I realize this is pretty beffy, sorry about that. I've already experienced several nasty merges, so I've been punished already for my transgression...

pkg/aspect/manager.go
```diff
@@ -28,50 +29,96 @@ import (
...
+		// Execute dispatches to the aspect manager.
+		Execute(attrs attribute.Bag, mapper expr.Evaluator) rpc.Status
+	}
+
+	// QuotaExecutor encapsulates a single QuotaManager aspect and allows it to be invoked.
+	QuotaExecutor interface {
+		Executor
+
+		// Execute dispatches to the aspect manager.
+		Execute(attrs attribute.Bag, mapper expr.Evaluator, qma *QuotaMethodArgs) (rpc.Status, *QuotaMethodResp)
```
ZackButcher: Small nit: I think it's more idiomatic to flip around these return params: even though it's not directly `error`, `rpc.Status` is how we communicate the success or failure of the call, and QuotaMethodResp is the primary return value of the call. Looking through the quota manager impl, it's strikes me as a little odd that in error cases we're doing `return foo, nil` rather than the reverse.

geeknoid: Done.

PR #462, opened 2017-03-24
Implement String() for adapter.ConfigError(s)

ZackButcher:
So that they've pretty-printed when used with %v directives.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/462)

### DIFF
--- a/pkg/adapter/configError.go
+++ b/pkg/adapter/configError.go
@@ -78,11 +78,19 @@ func (e ConfigError) Error() string {
 	return fmt.Sprintf("%s: %s", e.Field, e.Underlying)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/adapter/configError.go
```diff
@@ -78,13 +78,21 @@ func (e ConfigError) Error() string {
 	return fmt.Sprintf("%s: %s", e.Field, e.Underlying)
 }
 
+func (e ConfigError) String() string {
```
geeknoid: Please add a unit test to keep coverage at 100%.

ZackButcher: Done

PR #466, opened 2017-03-25
Switch a number of uses of strings to Kind enums instead.

geeknoid:
The Kind enum is moved from pkg/aspect to pkg/config, making it
possible to use Kind enums in the config package. This eliminates
many uses of generic strings in favor of the more mnemonic and
efficient Kind enum.

This also enables invalid kinds to be reported sooner in the config validation
process.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/466)


pkg/adapterManager/registry.go
```diff
@@ -60,66 +64,65 @@ func BuilderMap(builders []adapter.RegisterFn) BuildersByName {
...
+	r.insert(config.MetricsKind, b)
 }
 
-func (r *registry) insert(k aspect.Kind, b adapter.Builder) {
-	kind := k.String()
+func (r *registry) insert(k config.Kind, b adapter.Builder) {
 	ok := true
 	if glog.V(1) {
 		defer func() {
 			if ok {
```
ZackButcher: Should we log something when `!ok`?

pkg/config/kind.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package aspect
+package config
 
 // Kind of aspect
 type Kind int
```
ZackButcher: You use `Kind` as a `uint` everywhere, should we alias `uint` instead of `int` here?

pkg/config/runtime.go
```diff
@@ -24,41 +24,41 @@ import (
...
+	runtime struct {
 		Validated
 		// used to evaluate selectors
 		eval expr.PredicateEvaluator
 	}
 
-	// AspectSet is a set of aspects by name.
-	AspectSet map[string]bool
+	// KindSet is a set of aspects by kind.
+	KindSet int
```
ZackButcher: Can you define `SetKind(k kind)` and `AddKind(k kind)` methods on `KindSet` for `kind &= 1 << uint(k)` and `kind |= 1 << uint(k)` respectively? I think that'll be far more readable than the bit shifts littered around and ensures the code using `KindSet` doesn't have to blindly assume it's an `int`, or that `Kind` is an alias for `uint`.

geeknoid:
Great suggestions. All done.

PTAL.

pkg/adapterManager/manager.go
```diff
@@ -76,33 +78,46 @@ type builderFinder interface {
...
 		managers:      m,
 		mapper:        exp,
-		methodMap:     am,
 		executorCache: make(map[cacheKey]aspect.Executor),
 		gp:            gp,
 		adapterGP:     adapterGP,
 	}
+
+	for _, m := range inventory.Check {
+		mg.checkKindSet |= 1 << uint(m.Kind())
```
ZackButcher: Update this to use the new methods here and elsewhere.

PR #471, opened 2017-03-29
Simplify pkg/api.

geeknoid:
- Eliminate a useless abstraction layer. Now that there's a clean
AspectDispatcher interface, we don't really need the extra Handlers
concept in pkg/api.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/471)


douglas-reid:
@geeknoid trying to catch up to the AspectDispatcher... Won't we need a layer of abstraction if we want pre and post execution adapter phases (something like a k8s info adapter before check adapters, etc.) ?  Or would that logic now have to live in a Dispatcher impl?

geeknoid:
Good question.

We'll need to inject some new pre-aspect and (maybe) post-aspect callout in this code. But the handlers layer that was there would not be the right abstraction for that. We'll just need something new.

With this update, pkg/api takes things from incoming requests to calls to the aspectDispatcher. You can imagine having REST functionality in the same package that takes incoming REST calls and invokes the dispatcher too.

PR #482, opened 2017-03-30
Implement validation using descriptors for the quotas aspect

ZackButcher:
Note that the `ValidateConfig` test is essentially identical to the metricManager's `ValidateConfig` test; I expect we can refactor out nearly 100% of the config validation logic into a common place and test it once there. I'll take a stab at that when I move to the next aspect on the list.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/482)

### DIFF
--- a/pkg/aspect/quotasManager.go
+++ b/pkg/aspect/quotasManager.go
@@ -56,45 +56,16 @@ func newQuotasManager() QuotaManager {
 func (m *quotasManager) NewQuotaExecutor(c *cpb.Combined, a adapter.Builder, env adapter.Env, df descriptor.Finder) (QuotaExecutor, error) {
 	params := c.Aspect.Params.(*aconfig.QuotasParams)
 
-	// TODO: get this from config
-	if len(params.Quotas) == 0 {
-		params = &aconfig.QuotasParams{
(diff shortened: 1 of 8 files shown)


pkg/aspect/quotasManager.go
```diff
@@ -194,7 +170,10 @@ func quotaDefinitionFromProto(desc *dpb.QuotaDescriptor) (*adapter.QuotaDefiniti
 		labels[label.Name] = l
 	}
 
-	dur, _ := ptypes.DurationFromProto(desc.Expiration)
+	dur, err := ptypes.DurationFromProto(desc.Expiration)
```
geeknoid: We should update the descriptors to use the gogo features such that the Go version of the descriptor would just contain a time.Duration or time.Time directly.

That's not blocking this PR, but we should do the gogo treatment on all protos where it makes sense.

ZackButcher: We'll also need to update //pkg/config/proto and use and generate a gogo version rather than a regular old go version. Today we parse our config into a proto struct generated with protoc due to golang's broken handling of `Struct` protos; AFAIK no reason we couldn't use gogo's struct instead. @mandarjog

PR #485, opened 2017-03-30
Implement the attribute.Tracker.ApplyBag method.

geeknoid:
This implements the second half of the attribute protocol in the
mixer. The mixer can now both receive and send attributes using
the full dictionary/delta/context model.

This also removes the attribute.Manager interface type. This was
unnecessary, I added it when still getting a hang of Golang...

The mixc command is updated to take advantage of the new functionality
and to print any attributes that the mixer returns in a response.

Given this foundation, I'll setup a test environment to evaluate
API mixer perf.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/485)

### DIFF
--- a/cmd/client/cmd/util.go
+++ b/cmd/client/cmd/util.go
@@ -209,7 +209,10 @@ func parseAttributes(rootArgs *rootArgs) (*mixerpb.Attributes, error) {
 	}
 
 	var attrs mixerpb.Attributes
-	attribute.NewManager().NewTracker().ApplyBag(b, 0, &attrs)
+	if err := attribute.NewManager().NewTracker().ApplyBag(b, 0, &attrs); err != nil {
+		return nil, err
+	}
+
 	return &attrs, nil
 }
 
--- a/cmd/client/cmd/util.go
+++ b/cmd/client/cmd/util.go
@@ -15,10 +15,13 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 	"time"
 
 	rpc "github.com/googleapis/googleapis/google/rpc"
@@ -26,6 +29,8 @@ import (
 	"google.golang.org/grpc"
 
 	mixerpb "istio.io/api/mixer/v1"
+	"istio.io/mixer/cmd/shared"
+	"istio.io/mixer/pkg/attribute"
 	"istio.io/mixer/pkg/tracing"
 )
 
@@ -111,45 +116,41 @@ func parseStringMap(s string) (interface{}, error) {
 		v := pair[colon+1:]
 		m[k] = v
 	}
+
 	return m, nil
 }
 
-// add to dictionary
-func add2Dict(dictionary map[int32]string, s string) int32 {
-	// linear search to see if this string is already in the dictionary
-	for k, v := range dictionary {
-		if v == s {
-			return k
-		}
+func parseAny(s string) (interface{}, error) {
+	// auto-sense the type of attributes based on being able to parse the value
+	if val, err := parseInt64(s); err == nil {
+		return val, nil
+	} else if val, err := parseFloat64(s); err == nil {
+		return val, nil
+	} else if val, err := parseBool(s); err == nil {
+		return val, nil
+	} else if val, err := parseTime(s); err == nil {
+		return val, nil
+	} else if val, err := parseDuration(s); err == nil {
+		return val, nil
+	} else if val, err := parseBytes(s); err == nil {
+		return val, nil
+	} else if val, err := parseStringMap(s); err == nil {
+		return val, nil
 	}
-
-	index := int32(len(dictionary))
-	dictionary[index] = s
-	return index
-}
-
-func makeStringMap(dictionary map[int32]string, m map[string]string) mixerpb.StringMap {
-	sm := mixerpb.StringMap{Map: make(map[int32]string)}
-
-	for k, v := range m {
-		sm.Map[add2Dict(dictionary, k)] = v
-	}
-
-	return sm
+	return s, nil
 }
 
 type convertFn func(string) (interface{}, error)
 
-func process(dictionary map[int32]string, s string, f convertFn) (map[int32]interface{}, error) {
-	m := make(map[int32]interface{})
+func process(b *attribute.MutableBag, s string, f convertFn) error {
 	if len(s) > 0 {
 		for _, seg := range strings.Split(s, ",") {
 			eq := strings.Index(seg, "=")
 			if eq < 0 {
-				return nil, fmt.Errorf("attribute value %v does not include an = sign", seg)
+				return fmt.Errorf("attribute value %v does not include an = sign", seg)
 			}
 			if eq == 0 {
-				return nil, fmt.Errorf("attribute value %v does not contain a valid name", seg)
+				return fmt.Errorf("attribute value %v does not contain a valid name", seg)
 			}
 			name := seg[0:eq]
 			value := seg[eq+1:]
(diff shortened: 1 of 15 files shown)


pkg/attribute/tracker.go
```diff
@@ -15,29 +15,33 @@
...
 import (
+	"fmt"
 	"sync"
+	"time"
 
 	mixerpb "istio.io/api/mixer/v1"
 )
 
-// Tracker is responsible for tracking a set of live attributes over time.
+// Tracker is responsible for implementing the mixer's so-called attribute protocol.
```
douglas-reid: we can probably drop the "so-called" here.

pkg/attribute/tracker.go
```diff
@@ -99,13 +106,184 @@ func (at *tracker) ApplyRequestAttributes(attrs *mixerpb.Attributes) (*MutableBa
...
+			return false
+		}
+
+		for k, v := range t1 {
+			if v != t2[k] {
+				return false
+			}
+		}
+	default:
+		panic(fmt.Sprintf("Unsupported attribute value type: %T", v1))
```
douglas-reid: is panic really the right thing here?

I feel like we should be avoid using panics wherever possible in this code. Returning an error, which can be passed up and out of ApplyBag() seems feasible.

geeknoid:
PTAL.

cmd/client/cmd/util.go
```diff
@@ -282,3 +228,34 @@ func decodeStatus(status rpc.Status) string {
...
+	if len(names) == 0 {
+		return
+	}
+
+	sort.Strings(names)
+
+	buf := bytes.Buffer{}
+	tw := tabwriter.NewWriter(&buf, 0, 0, 2, ' ', 0)
+
+	fmt.Fprint(tw, "  Attribute\tType\tValue\n")
```
douglas-reid: curious: any reason to not start the indents with a `\t` any get the padding?

pkg/api/grpcServer_test.go
```diff
@@ -459,7 +466,7 @@ func TestRudeClose(t *testing.T) {
 }
 
 func TestBrokenStream(t *testing.T) {
-	ts, err := prepTestState(30000)
+	ts, err := prepTestState(30009)
```
douglas-reid: I feel like we are doing the wrong thing here. I don't think we should be specifying ports at all in testing, if we can avoid it. IIRC, we should be able to listen on port 0 (or unspecified) and then get the actual port from the listener addr assigned, if needed.

pkg/api/grpcServer_test.go
```diff
@@ -488,6 +495,42 @@ func TestBrokenStream(t *testing.T) {
...
+		err = errors.New("nothing good")
+		wg.Done()
+		return err
+	}
+
+	ts.reportBadAttr = true
+
+	stream, err := ts.client.Report(context.Background())
+	if err != nil {
+		t.Errorf("Report failed %v", err)
```
douglas-reid: do we want t.FatalF() here?

pkg/api/grpcServer_test.go
```diff
@@ -488,6 +495,42 @@ func TestBrokenStream(t *testing.T) {
...
+		return
+	}
+
+	request := mixerpb.ReportRequest{}
+	if err := stream.Send(&request); err != nil {
+		t.Errorf("Failed to send request: %v", err)
+	}
+
+	wg.Wait()
+	// we never hear back...
```
douglas-reid: this seems ominous.

pkg/api/grpcServer_test.go
ZackButcher: I thought bazel sandboxed ports in tests? AFAIK you could run two http servers in different tests on the "same" port. I suppose we could write a quick test to find out.

pkg/api/grpcServer_test.go
ZackButcher: Yeah, we have `t.Errorf(...); return` all over the place in our tests; we should swap to using `Fatal`. This requires swapping to `t.Run` in table driven tests, though.

pkg/api/grpcServer.go
```diff
@@ -78,8 +78,10 @@ func (s *grpcServer) dispatcher(stream grpc.Stream, methodName string,
 		request proto.Message, response proto.Message)) error {
 
 	// tracks attribute state for this stream
-	tracker := s.attrMgr.NewTracker()
-	defer tracker.Done()
+	reqTracker := s.attrMgr.NewTracker()
```
ZackButcher: Do we ever need/use a singular tracker outside of tests? Should `NewTracker` be updated to ` func NewTracker() (req Tracker, resp Tracker)`?

pkg/api/grpcServer.go
```diff
@@ -133,8 +135,11 @@ func (s *grpcServer) dispatcher(stream grpc.Stream, methodName string,
 			worker(ctx2, requestBag, responseBag, request, response)
 
 			sendLock.Lock()
-			tracker.GetResponseAttributes(responseBag, responseAttrs)
-			err := s.sendMsg(stream, response)
+			if err = respTracker.ApplyBag(responseBag, 0, responseAttrs); err != nil {
```
ZackButcher: Shouldn't this be `respTracker.ApplyBag(responseBag, req.RequestIndex, responseAttrs)` instead of a hardcoded `0` for the context id?

cmd/client/cmd/util.go
geeknoid: 'Cause I feel as though padding is different than indenting...

pkg/api/grpcServer_test.go
geeknoid: This is not bazel, this is the go test infrastructure. Multiple individual tests in a single test program can be run in parallel and so they _do_ get port conflicts (that's why the code was the way it was).

I didn't know that a 0 port would give a valid port automatically. So I've updated the code accordingly.

pkg/api/grpcServer_test.go
geeknoid: I updated to Fatalf in this file. Frankly, I don't see the difference in either though. The test would fail in the effective same way.

pkg/api/grpcServer_test.go
geeknoid: Updated the comment to be more explicit

pkg/api/grpcServer.go
geeknoid: Considering we do use single trackers in tests, I like having them as singletons. They are completely independent from one another...

pkg/api/grpcServer.go
geeknoid: That's not how the context stuff is intended to work.

The context is intended to let the sender have different buckets of state held by the other side. A single message results in only one context's attributes being affected. If the mixer can readily identify different kinds of traffic and select between different contexts on the fly, then it could potentially minimize the amount of delta traffic produced. 

The idea was that the proxy could potentially use this to represent the traffic coming from different sources.  

Each context used costs memory on the other side to track a distinct set of attributes. So there must be a constraint and contexts must be reused as a result.

The mixer doesn't have what it needs right now to make an intelligent context choice, and the proxy currently avail itself of this feature either.

geeknoid:
PTAL.

pkg/api/grpcServer.go
ZackButcher: Gotcha, I was misunderstanding the purpose.

pkg/api/grpcServer_test.go
```diff
@@ -135,16 +142,16 @@ func (ts *testState) Quota(ctx context.Context, requestBag *attribute.MutableBag
 }
 
 func TestCheck(t *testing.T) {
-	ts, err := prepTestState(29999)
+	ts, err := prepTestState()
 	if err != nil {
-		t.Errorf("unable to prep test state %v", err)
+		t.Fatalf("Unable to prep test state: %v", err)
 		return
```
douglas-reid: so, if we use t.FatalF, we shouldn't need the `return`. It is supposed to stop the test execution immediately, which is why I originally suggested it.

PR #487, opened 2017-03-30
Update configs to match istio/istio configs

ZackButcher:
Update configs to match istio/istio configs, removing attributes we don't actually populate today

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/487)

### DIFF
--- a/testdata/globalconfig.yml
+++ b/testdata/globalconfig.yml
@@ -1,4 +1,4 @@
-subject: namespace:ns
+subject: "namespace:ns"
 revision: "2022"
 adapters:
   - name: default
@@ -20,10 +20,6 @@ attributes:
     value_type: 1 # STRING
   - name: target.name
     value_type: 1 # STRING
-  - name: api.name
-    value_type: 1 # STRING
-  - name: api.method
-    value_type: 1 # STRING
   - name: response.code
     value_type: 2 # INT64
   - name: response.latency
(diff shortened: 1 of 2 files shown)


testdata/serviceconfig.yml
```diff
@@ -17,16 +17,12 @@ rules:
         labels:
           source: source.name | "unknown"
           target: target.name | "unknown"
-          service: api.name | "unknown"
```
douglas-reid: do we need to remove these?  leaving them as "unknown" seems somewhat reasonable. it would also likely save us from having to update the example dashboard (with promql).

testdata/globalconfig.yml
```diff
@@ -1,4 +1,4 @@
-subject: namespace:ns
+subject: "namespace:ns"
 revision: "2022"
```
douglas-reid: aside: we should figure out something sensible for this revision bit.

testdata/serviceconfig.yml
ZackButcher: We removed the `api.*` attributes from the canonical set, they're still up in the air. I'm fine keeping unknown, but I did go ahead and pull those fields out of the istio/istio repo's configs.

douglas-reid: If we remove them here, we should add the updates to the existing dashboard config in this PR as well that pulls them out too.

ZackButcher:
Not strictly required and I'm about to go change config again, so this is going away.

PR #489, opened 2017-03-30
Implement config validation for the lists aspect

ZackButcher:
No configs actually use this aspect so we don't need to update any of the yaml files.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/489)

### DIFF
--- a/pkg/aspect/listsManager.go
+++ b/pkg/aspect/listsManager.go
@@ -19,6 +19,7 @@ import (
 
 	rpc "github.com/googleapis/googleapis/google/rpc"
 
+	apipb "istio.io/api/mixer/v1/config/descriptor"
 	"istio.io/mixer/pkg/adapter"
 	aconfig "istio.io/mixer/pkg/aspect/config"
 	"istio.io/mixer/pkg/attribute"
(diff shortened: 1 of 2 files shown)


pkg/aspect/listsManager.go
```diff
@@ -66,14 +67,20 @@ func (listsManager) Kind() config.Kind {
...
-	lc := c.(*aconfig.ListsParams)
-	if lc.CheckExpression == "" {
-		ce = ce.Appendf("CheckExpression", "Missing")
+func (listsManager) ValidateConfig(c config.AspectParams, v expr.Validator, df descriptor.Finder) (ce *adapter.ConfigErrors) {
+	cfg := c.(*aconfig.ListsParams)
+	if cfg.CheckExpression == "" {
+		ce = ce.Appendf("CheckExpression", "no expression provided")
+		return
+	}
+	if t, err := v.TypeCheck(cfg.CheckExpression, df); err != nil {
```
mandarjog: Looks like the API we really want is

`v.TypeAssert(expr, df, expectedType)` That will save this double call every time.

ZackButcher: The other option would be to extract this logic into a function in `//pkg/aspect/common.go`. Do we prefer that, or updating the validator interface? Either way I'll make the update in my next validation PR and update all the managers.

ZackButcher: Chatted offline, we'll update the expression validator interface to add the new method in a follow on PR.

PR #500, opened 2017-03-31
Implement config validation for access logs, and implement use of descriptors

ZackButcher:
We implement config validation. At the same time we update the access log manager to use descriptors. This update makes it natural to remove the log_format enum from the access log config: it adds substantial special logic if we 'know' about certain descriptors automagically, so I've removed the special casing around apache common/combined access logs. As a result, once again, access logs and application logs are essentially the same.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/500)

### DIFF
--- a/pkg/aspect/accessLogsManager.go
+++ b/pkg/aspect/accessLogsManager.go
@@ -78,7 +78,7 @@ func (accessLogsManager) Kind() config.Kind { return config.AccessLogsKind }
 func (accessLogsManager) DefaultConfig() config.AspectParams {
 	return &aconfig.AccessLogsParams{
 		LogName: "access_log",
-		Log: &aconfig.AccessLogsParams_AccessLog{
+		Log: aconfig.AccessLogsParams_AccessLog{
 			DescriptorName: "common",
 		},
 	}
@@ -88,10 +88,6 @@ func (accessLogsManager) ValidateConfig(c config.AspectParams, v expr.Validator,
 	cfg := c.(*aconfig.AccessLogsParams)
 	if cfg.LogName == "" {
 		ce = ce.Appendf("LogName", "no log name provided")
-	}
-	if cfg.Log == nil {
-		// We can't do any more validation without a Log
-		return ce.Appendf("AccessLogs.Log", "an AccessLog entry must be provided")
 	}
 
 	desc := df.GetLog(cfg.Log.DescriptorName)
--- a/pkg/aspect/accessLogsManager.go
+++ b/pkg/aspect/accessLogsManager.go
@@ -32,6 +32,10 @@ import (
 	"istio.io/mixer/pkg/status"
 )
 
+// TODO: there are several hard-coded but end-user-configurable things in this file we need to work to eliminiate:
+// - default config assumes a log descriptor named "common" is available
+// - if we get a label or template param named exactly "timestamp" we populate the value with time.Now()
+
 type (
 	accessLogsManager struct{}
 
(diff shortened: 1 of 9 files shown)


testdata/serviceconfig.yml
```diff
@@ -32,7 +32,7 @@ rules:
     params:
       logName: "access_log"
       log:
-        logFormat: 1 # COMMON
+        descriptorName: "common"
```
mandarjog: we have used `descriptor_name` and `descriptorName` let's pick one.

mandarjog: looking at the keys again, and how it plays out with UX, `descriptor_name` is really `logType`

ZackButcher:
This PR also implements validation on log template param expressions and wires it up in both access and application logs.

testdata/globalconfig.yml
```diff
@@ -60,7 +76,14 @@ metrics:
...
-- name: RequestCount
-  max_amount: 5
-  expiration:
-    seconds: 1
+  - name: RequestCount
+    max_amount: 5
+    expiration:
+      seconds: 1
+logs:
+  - name: common
```
mandarjog: Let's make the name `accesslog.common` and `accesslog.combined`
That we __can__ retire the accessLog `kind`.  Having descriptive names serves similar purpose.

testdata/serviceconfig.yml
ZackButcher: `descriptor_name` is consistent with all the other aspects (including application logs). I think it'd be pretty odd to make this one aspect different.

ZackButcher:
PTAL

pkg/aspect/accessLogsManager_test.go
```diff
@@ -54,36 +74,20 @@ func TestAccessLoggerManager_NewAspect(t *testing.T) {
...
-	//	name:   "custom_access_log",
-	//	aspect: tl,
-	//	labels: map[string]string{"test": "test"},
-	//}
-
 	combinedStruct := &aconfig.AccessLogsParams{
 		LogName: "combined_access_log",
 		Log: &aconfig.AccessLogsParams_AccessLog{
-			LogFormat: aconfig.COMBINED,
+			DescriptorName: "common",
```
mandarjog: Since common log and combined log are known log formats, can we add a test that ensures that our packaged default descriptors actually  produce the correct log entries?

We should also agree on a mechanism to provide default descriptors. I think mixer config manager with packaged default set of descriptors as yml files. (like we have in testdata/ is a reasonable start)

ZackButcher: I was chatting with Doug about this (and more generally how we've been going round and round on this) and we came to the same conclusion: we'll have to figure out some way to package an always-available istio-global-config with things like these descriptors, the common metric descriptors, and so on. I'm actually not opposed to it being static protos that we compile into the binary (speaking of going round-and-round).

ZackButcher:
PTAL.

testdata/serviceconfig.yml
```diff
@@ -30,19 +30,10 @@ rules:
           response_code: response.code | 200
   - kind: access-logs
     params:
-      logName: "access_log"
+      logName: access_log
       log:
-        logFormat: 1 # COMMON
-        templateExpressions:
```
douglas-reid: can we keep the labels?  i still think we want the structured logs, if possible -- even if the config is ugly as sin.

pkg/aspect/accessLogsManager.go
```diff
@@ -61,22 +56,9 @@ func newAccessLogsManager() ReportManager {
...
-	}
-
-	// TODO: when users can provide us with descriptors, this error can be removed due to validation
-	tmpl, err := template.New("accessLogsTemplate").Parse(templateStr)
-	if err != nil {
-		return nil, fmt.Errorf("log %s failed to parse template '%s' with err: %s", cfg.LogName, templateStr, err)
-	}
+	// validation ensures both that the descriptor exists and that its template is parsable by the template library.
+	desc := df.GetLog(cfg.Log.DescriptorName)
+	tmpl, _ := template.New("accessLogsTemplate").Parse(desc.LogTemplate)
```
ZackButcher: The comment above `desc` two lines up covers both.

testdata/serviceconfig.yml
ZackButcher: I'll add them back

ZackButcher:
PTAL

pkg/aspect/accessLogsManager.go
douglas-reid: doh! reading comprehension fail.

testdata/serviceconfig.yml
douglas-reid: I see the descriptors now... do we need the labels section here too?

ZackButcher: It's there, the github diff doesn't render it cause it's unchanged from before. If you expand the file down you should see the labels.

ZackButcher: Oh and make sure you're looking at the latest pushed commit; if you click on the file name here it goes to the commit that you commented on, not the latest thing I pushed (tricked me at first too).

Issue #475, opened 2017-03-29
Define an adapter model to support attribute injection on the mixer.


geeknoid:
We just discussed this. General idea is:

- We'll introduce a new aspect kind for attribute injection.

- We'll extend the manager inventory to include a list of attibute injectors (the inventory currently has CheckManagers, ReportManagers, and QuotaManagers, now it'll gain InjectionManagers)

- We'll change the adapter manager to synchronously invoke injection aspects prior to invoking check/report/quota aspects.

douglas-reid:
Closing this, as we have basic functionality in the mixer codebase yet.

PR #502, opened 2017-03-31
Introduce preprocess step into the mixer dispatch.

douglas-reid:
This PR adds a Preprocess() method to the AspectDispatcher. This
method (and the associated logic used to invoke it before other
mixer dispatch for Check, Report, and Quota) will be used to
execute Aspects that must preceed all further processing. Example
uses include: environment-specific attribute generation and
identity-related tasks.

Attributes generated in preprocess step will be merged back in
with the original request bags and passed to the existing dispatch
methods.

Currently, there are no existing Aspects that are associated
with the preprocess step (as a result, it acts as simple pass-thru). 
However, a new aspect kind was added for testing purposes. It is
expected that that aspect will be used in follow-on PRs.

Note: it is expected that the preprocess aspects will behave and
be configured in the exact same manner as other aspects. However,
in subsequent PRs, we may want to restrict the ability to
configure some of these aspects to only be globally-configurable
to prevent possibly conflicts between selectors and the attributes
they depend on. More thinking is needed here. For now, this potential
problem is left alone.

In addition to unit testing, this change was tested via mixs/mixc as follows
(from two different prompts):

```shell
$ ./bazel-bin/cmd/server/mixs server --globalConfigFile=testdata/globalconfig.yml --serviceConfigFile=testdata/serviceconfig.yml
Istio Mixer: version: stable-b875f9f-4-g5230128 (build: 2017-03-31-5230128, status: Modified)
Starting gRPC server on port 9091
{"logName":"access_log","labels":{"timestamp":"2017-03-31T15:33:39.157182481-07:00"},"textPayload":"- - - [31/Mar/2017:15:33:39 -0700] \"- - -\" - -"}
```

```shell
$ ./bazel-bin/cmd/client/mixc report 
Report RPC returned OK
```

Related Issue: #475

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/502)


pkg/api/grpcServer.go
```diff
@@ -71,11 +94,8 @@ func NewGRPCServer(aspectDispatcher adapterManager.AspectDispatcher, tracer trac
 }
 
 // dispatcher does all the nitty-gritty details of handling the mixer's low-level API
-// protocol and dispatching to the right API handler.
-func (s *grpcServer) dispatcher(stream grpc.Stream, methodName string,
-	getState func() (request proto.Message, response proto.Message, requestAttrs *mixerpb.Attributes, responseAttrs *mixerpb.Attributes, result *rpc.Status),
-	worker func(ctx context.Context, requestBag *attribute.MutableBag, responseBag *attribute.MutableBag,
-		request proto.Message, response proto.Message)) error {
+// protocol and dispatching to the right API dispatchWrapperFn.
+func (s *grpcServer) dispatcher(stream grpc.Stream, methodName string, getState stateGetterFn, worker dispatchFn) error {
```
mandarjog: 1. nit. rename this to `dispatch` 
2. I think it would be clearer if we make a call to `s.preprocess()` here.
I do like the python style func decoration with s.preprocess() !

ZackButcher:
I haven't reviewed the code yet, but there's a few potential problems we need to think carefully about:

- If preprocess adapters are not global, config cannot be validated outside of the request context
  - This is because at config validation time we need to know all attributes that are available so that we can ensure we can satisfy all expressions in config for a given selector; if preprocess adapters are under a selection, we must know if that section applies when determining if its valid for another section to depend on attributes produced by those preprocess adapters. This can only be determined in the request context.*
- How do we describe the attributes produced by those adapters? Today the only place it's possible for us to ingest attribute descriptors is via the global config (obviously this is changing with @mandarjog's config API work)
- How do we handle the case that preprocess adapter B depends on an attribute produced by preprocess adapter A? In general the only way we can guarantee successful execution is to compute a DAG of attribute-dependencies, topologically sort it, and execute adapters in top-sorted order
  - The simplest thing to do here is disallow preprocess adapters depending on anything but proxy attribute. That kills much of their power, especially in the context of platform specific preprocess adapters, where no other preprocess adapter can have access to e.g. k8s specific attributes.

\* Barring @geeknoid's plan to remove attributes from config resolution.


@geeknoid did I miss any of the other potentially tricky cases we discussed the other day?

pkg/adapterManager/manager.go
```diff
@@ -149,10 +159,27 @@ func (m *Manager) Quota(ctx context.Context, requestBag *attribute.MutableBag, r
...
+			}
+			return rpcStatus
+		})
+}
+
 type invokeExecutorFunc func(executor aspect.Executor, evaluator expr.Evaluator) rpc.Status
 
 // dispatch resolves config and invokes the specific set of aspects necessary to service the current request
-func (m *Manager) dispatch(ctx context.Context, requestBag *attribute.MutableBag, responseBag *attribute.MutableBag,
+func (m *Manager) dispatch(ctx context.Context, requestBag, responseBag *attribute.MutableBag,
```
mandarjog: Dispatch ends up calling `config.Resolve()`
When the call chain is `Preprocess --> dispatch` it is not safe to do so `config.Resolve()`


I think we need to change the structure. 
`Preprocess` should only use adapters (because they do not have dependencies on attributes)

`Manager.dispatch` can implement request lifecycle so it will look like this
```
func dispatch () {
	cfg, _ := m.cfg.Load().(config.Resolver)

        m.Preprocess(cfg)    //  <-- do not call resolve on config
                                         // we can only safely access all descriptors and other globals.
                                         // Add new type of global called Preprocessor 

	cfgs, err := cfg.Resolve(requestBag, kindSet)
}

func Preprocess (cfg) {
      adapters := cfg.GetPreProcessAdapters()

      foreach adapter in adapters:
         adapter.preprocess()
}
```

douglas-reid:
@ZackButcher a few responses: 

- first pass, there is no ordering allowed between preprocessor adapters - just as we have no support for ordering amongst adapters for other aspects. In the short term, this should not be a limitation (we don't even support any preprocessing adapters now).
- describing attributes produced by a kind of preprocessing adapter is a problem that is a bit orthogonal to supporting the notion of preprocessing. we can start to figure that out when it comes time to implement the attribute-generation aspect kind.
- i think config validation should fail for non-global use of any attribute-generation preprocess adapters. i'm not 100% convinced that all preprocess adapters will generate new attributes.

pkg/adapterManager/manager.go
douglas-reid: I'm not convinced that we want to have special adapters that have no matching aspects and completely different configuration story. Following the story through config to managers to builders to aspects is confusing enough as is.

It seems like there may be other approaches to solving these concerns that don't involve new, distinct code paths. For instance, we could enforce, at validation time, that certain aspect kinds (`attribute-generation`, for example) cannot live anywhere but in a selector of * (must be global).

douglas-reid:
I'm confused. Preprocessing, in this PR, *is* being treated as middleware. It wraps all dispatch calls to the various API handlers -- functioning like an interceptor or a http.HandleFunc.

pkg/adapterManager/manager.go
mandarjog: I think there are 2 issue with that. 

1. `config.Resolve` is called 2 times on the request path
2.  Resolve won't know to do a shallow evaluation the 1st time.
     We could add that as a parameter, but that won't really work either.

We already support creating and configuring adapters. We do not need to insist that every adapter must have a matching aspect. 

Are we prioritizing code and framework reuse over semantics?

That said, if the concerns can be addressed with code reuse then it is great.

douglas-reid: Yeah, I thought about having Resolve short-circuit for kinds in the preprocess kind set and/or making a ResolveGlobal.

I'm concerned about having a manager by a different name and stuck in a strange part of the codebase. But I'm also concerned about trying to explain why some adapters don't have corresponding aspects in the config file (and why you can't have cross adapter config for them). I'm worried it will be a strange UX.

I'll noodle on this for awhile this weekend. Perhaps there is an approach that will be OK for alpha, but will need tweaking for beyond.

Let's discuss again Monday. SG?

mandarjog:
Yes, SG.

pkg/api/grpcServer.go
douglas-reid: Addressed (1).  I've left (2) alone for now. I think there is work needed to refactor how config gets resolved and passed around as part of this dispatch() call. If OK with you, I'd like to hold off on that bit of clean-up until tackling the larger task in a more wholesale manner.

douglas-reid:
PTAL. I've made the following changes:
- added a Resolver.ResolveUnconditional() method that will only evaluate rules with an empty selector ("").
- shuffled around some of the adapterManager dispatch code to load and resolve before passing to dispatch (this should set the stage for a follow-on refactor to completely separate adapter dispatch from config resolution)
- added a few tests

douglas-reid:
merging to unblock @mandarjog 's pending PR, as discussed.

Issue #457, opened 2017-03-24
Propagate errors from redis quota adapter


chowchow316:
For now, not all of the errors raised in redis quota adapter are appropriately propagated up. We should not drop any error happened when connecting/communicating with redis. And we should propagate the errors up, possibly via expanding `quotaFunc.`

chowchow316:
#507 Propagated redis errors via QuotaUtil.Logger, also need to track returning errors from the commonWrapper callback function.

PR #507, opened 2017-04-03
Add expiration to keys.

chowchow316:
This PR deals with expiring keys with redis (works with real redis server), but since the mock redis does not support timeout function, unit test now has low code coverage. Will explore gomock and try to improve code coverage.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/507)

### DIFF
--- a/adapter/redisquota/redisquota.go
+++ b/adapter/redisquota/redisquota.go
@@ -111,23 +111,34 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
 	amount, exp, err := rq.common.CommonWrapper(args, func(d *adapter.QuotaDefinition, key string, currentTime time.Time, currentTick int64) (int64, time.Time,
 		time.Duration) {
 		result := args.QuotaAmount
-		conn, _ := rq.redisPool.get()
-		// TODO: propagate Connection Pool error here
-		// if err != nil {
-		// 	rq.logger.Infof("Could not get connection to redis")
-		// 	return 0, time.Time{}, 0
-		//}
+		seconds := int32((d.Expiration + time.Second - 1) / time.Second)
+
+		conn, err := rq.redisPool.get()
+		if err != nil {
+			rq.common.Logger.Warningf("Could not get connection to redis %v", err)
+			return 0, time.Time{}, 0
+		}
 		defer rq.redisPool.put(conn)
 
 		// increase the value of this key by the amount of result
 		conn.pipeAppend("INCRBY", key, result)
-		resp, _ := conn.pipeResponse()
-		// TODO: propagate Connection Pool error here
-		// if err != nil {
-		//	rq.logger.Infof("Could not get response from redis")
-		//	return 0, time.Time{}, 0
-		//}
-		ret := resp.int()
+		resp, err := conn.pipeResponse()
+		if err != nil {
+			rq.common.Logger.Warningf("Could not get response from redis %v", err)
+			return 0, time.Time{}, 0
+		}
+		ret, _ := resp.int()
+		if d.Expiration != 0 {
+			conn.pipeAppend("EXPIRE", key, seconds)
+			resp, _ = conn.pipeResponse()
+			rv, err := resp.int()
+			if err != nil {
+				rq.common.Logger.Warningf("Got error when setting expire for key %v", err)
+			}
+			if rv != 1 {
+				rq.common.Logger.Warningf("Could not set expire for key")
+			}
+		}
 
 		if ret > d.MaxAmount {
 			if !bestEffort {
@@ -135,9 +146,10 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
 			}
 			// grab as much as we can
 			result = d.MaxAmount - (ret - result)
-			conn.pipeAppend("SET", key, d.MaxAmount)
+			conn.pipeAppend("DECRBY", key, (ret - d.MaxAmount))
 			resp, _ = conn.pipeResponse()
-			if resp.int() != d.MaxAmount {
+			res, _ := resp.int()
+			if res != d.MaxAmount {
 				rq.common.Logger.Warningf("Could not set value to key.")
 			}
 		}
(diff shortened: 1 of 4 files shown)


adapter/redisquota/redisquota.go
```diff
@@ -111,33 +111,45 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
...
-		// TODO: propagate Connection Pool error here
-		// if err != nil {
-		// 	rq.logger.Infof("Could not get connection to redis")
-		// 	return 0, time.Time{}, 0
-		//}
+		seconds := int32((d.Expiration + time.Second - 1) / time.Second)
+
+		conn, err := rq.redisPool.get()
+		if err != nil {
+			rq.common.Logger.Warningf("Could not get connection to redis %v", err)
```
geeknoid: Here and elsewhere, probably should be an error instead of a warning.

adapter/redisquota/redisquota.go
```diff
@@ -111,33 +111,45 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
...
+		}
+		ret, _ := resp.int()
+		if d.Expiration != 0 {
+			conn.pipeAppend("EXPIRE", key, seconds)
+			resp, _ = conn.pipeResponse()
+			rv, err := resp.int()
+			if err != nil {
+				rq.common.Logger.Warningf("Got error when setting expire for key %v", err)
+			}
+			if rv != 1 {
```
geeknoid: make that "else if"

adapter/redisquota/redisquota.go
```diff
@@ -111,33 +111,45 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
...
-		// TODO: propagate Connection Pool error here
-		// if err != nil {
-		//	rq.logger.Infof("Could not get response from redis")
-		//	return 0, time.Time{}, 0
-		//}
-		ret := resp.int()
+		resp, err := conn.pipeResponse()
+		if err != nil {
+			rq.common.Logger.Warningf("Could not get response from redis %v", err)
+			return 0, time.Time{}, 0
```
geeknoid: Really need to return errors everywhere instead of swallowing the failure code.

adapter/redisquota/redisquota.go
```diff
@@ -111,33 +111,45 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
...
 			if !bestEffort {
 				return 0, time.Time{}, 0
 			}
 			// grab as much as we can
 			result = d.MaxAmount - (ret - result)
-			conn.pipeAppend("SET", key, d.MaxAmount)
+			conn.pipeAppend("DECRBY", key, (ret - d.MaxAmount))
 			resp, _ = conn.pipeResponse()
-			if resp.int() != d.MaxAmount {
+			res, _ := resp.int()
```
geeknoid: Handle error return here and in the line above.

adapter/redisquota/redisquota.go
```diff
@@ -156,28 +168,64 @@ func (rq *redisQuota) ReleaseBestEffort(args adapter.QuotaArgs) (int64, error) {
...
+				resp, err = conn.pipeResponse()
+				if err != nil {
+					rq.common.Logger.Warningf("Could not check key in redis %v", err)
+					return 0, time.Time{}, 0
+				}
+				rv, err := resp.int()
+				if err != nil {
+					rq.common.Logger.Warningf("Got error when setting expire for key %v", err)
+				}
+				if rv != 1 {
```
geeknoid: else if

adapter/redisquota/redisquota.go
```diff
@@ -156,28 +168,64 @@ func (rq *redisQuota) ReleaseBestEffort(args adapter.QuotaArgs) (int64, error) {
...
+					rq.common.Logger.Warningf("Could not set expire for key")
+				}
+			}
 
 			if ret <= 0 {
 				// delete the key since it contains no useful state
 				conn.pipeAppend("DEL", key)
 				// consume the output of previous command
 				resp, _ = conn.pipeResponse()
```
geeknoid: handle error return.

chowchow316: Done.

adapter/redisquota/redisquota.go
chowchow316: Done.

adapter/redisquota/redisquota.go
```diff
@@ -111,34 +114,66 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
...
-		//	return 0, time.Time{}, 0
-		//}
-		ret := resp.int()
+		resp, err := conn.pipeResponse()
+		if err != nil {
+			rq.redisError = rq.common.Logger.Errorf("Could not get response from redis %v", err)
+			return 0, time.Time{}, 0
+		}
+		ret, err := resp.int()
+		if err != nil {
```
geeknoid: Here and elsewhere, please combine the 'if' with the previous line as per Go conventions.

adapter/redisquota/redisquota.go
```diff
@@ -111,34 +114,66 @@ func (rq *redisQuota) alloc(args adapter.QuotaArgs, bestEffort bool) (adapter.Qu
...
+			rq.redisError = rq.common.Logger.Errorf("Unable to get integer value: %v", err)
+			return 0, time.Time{}, 0
+		}
+
+		if d.Expiration != 0 {
+			conn.pipeAppend("EXPIRE", key, seconds)
+			resp, err = conn.pipeResponse()
```
geeknoid: Would it be useful to add a utility method that combines the pipeResponse with the resp.int in one step?

adapter/redisquota/redisquota.go
chowchow316: If the previous line only returns `err,` these two lines can be combined. Otherwise, if I combine these two lines and use `ret` later, it will complain about `ret` not defined.

adapter/redisquota/redisquota.go
chowchow316: Yes, will add it.

chowchow316:
Yes, #457 is actually used to tracking errors, just added some update there.

PR #511, opened 2017-04-05
Register redis adapter into inventory.

chowchow316:
This PR attempted to register redis quota adapter into adapter inventory, and I'm working on test thread safety now.  The steps for testing quota method:
1. change `memQuota` to `redisQuota` in globalconfig.yaml.
2. run mixs with config: _./mixs server \
  --globalConfigFile testdata/globalconfig.yml \
  --serviceConfigFile testdata/serviceconfig.yml  --logtostderr_
3. run mixc: _./mixc quota --amount 10 --name RequestCount_
@geeknoid Should the name for mixc be the same as listed in globalconfig quotas section? Is there anything I'm missing here?

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/511)

### DIFF
--- a/adapter/redisquota/driver.go
+++ b/adapter/redisquota/driver.go
@@ -15,8 +15,6 @@
 package redisquota
 
 import (
-	"sync"
-
 	"github.com/mediocregopher/radix.v2/pool"
 	"github.com/mediocregopher/radix.v2/redis"
 )
@@ -25,7 +23,6 @@ import (
 type connPool struct {
 	// TODO: add number of connections here
 	pool *pool.Pool
-	sync.Mutex
 }
 
 type connection struct {
@@ -39,9 +36,7 @@ type response struct {
 
 // Get is to get a connection from the connection pool.
 func (cp *connPool) get() (*connection, error) {
-	cp.Lock()
 	client, err := cp.pool.Get()
-	cp.Unlock()
 
 	return &connection{client, 0}, err
 }
@@ -57,7 +52,7 @@ func newConnPool(redisURL string, redisSocketType string, redisPoolSize int64) (
 	if err != nil {
 		return nil, err
 	}
-	return &connPool{pool, sync.Mutex{}}, err
+	return &connPool{pool}, err
 }
 
 func (cp *connPool) empty() {
--- a/adapter/redisquota/driver.go
+++ b/adapter/redisquota/driver.go
@@ -15,15 +15,17 @@
 package redisquota
 
 import (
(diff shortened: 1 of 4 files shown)


adapter/redisquota/driver.go
```diff
@@ -37,7 +39,9 @@ type response struct {
 
 // Get is to get a connection from the connection pool.
 func (cp *connPool) get() (*connection, error) {
+	cp.Lock()
```
geeknoid: I think the put function also needs locking, right?

geeknoid: Actually, looking at the implementation for the Redis pool functions, they already seem to be thread-safe. So it doesn't seem as though we need any logic for mutual exclusion of calls to the Redis pool functions.

chowchow316: Yes, at first I also think these functions are already thread-safe. But I want to test within Mixer by myself, could you help to clarify the questions I posed before (in the PR description)? Thank you!

geeknoid: Regarding your question, the command-lines you show there seems like it should be sufficient and should work.

chowchow316: Thanks, Martin. And I found one interesting thing here... even I test memQuota (from the head of master), and run the command-lines, I got the following error:
`Quota RPC returned INVALID_ARGUMENT (builder: name:"default" kind:"quotas" impl:"memQuota" params:&Params{MinDeduplicationDuration:1s,}  aspect: kind:"quotas" adapter:"default" params:&QuotasParams{Quotas:[],} :Unknown quota 'RequestCount' requested), amount 0, expiration 0s`
Do you have any clue of this error?

geeknoid: @ZackButcher made some changes here in the last couple of days, so there might be something wrong with the config or the aspect manager. This needs debugging...

ZackButcher: `Unknown quota 'RequestCount' requested` points to a missing quota descriptor in the global config - can you post your global config? [This is the key section.](https://github.com/istio/mixer/blob/master/testdata/globalconfig.yml#L78-L82)

`builder: name:"default" kind:"quotas" impl:"memQuota"` I think you're still getting memQuota and not redisQuota too. That should be at the top of the globalconfig in the [`adapters` section](https://github.com/istio/mixer/blob/master/testdata/globalconfig.yml#L3).

chowchow316: #518 solved this issue.

PR #517, opened 2017-04-07
Introduce attributes aspect kind to the mixer.

douglas-reid:
Related issue: #475 

This PR adds a new aspect kind (and associated manager and config)
to the istio mixer. The new aspect kind is meant to be used exclusively
in the preprocess phase of mixer request handling. For inputs, attributes
are mapped via expressions into named values that are passed into
individual adapters. The adapters produce a map of named values. These
named values are mapped into attributes through a configured mapping
in the aspect config.

To configure the new aspect kind, the operator will specify the input value
expression map and the attribute bindings (which bind adapter output value
names to attribute descriptor names). If the `attribute_bindings` includes names
of attribute descriptors which are unknown to the mixer, the configuration will
fail validation. Any key-value pairs produced by implementing adapters that 
are not included in the `attribute_bindings` will be dropped on the floor.

Example aspect configs:
```
...
aspects:
  - kind: attributes
    adapter: mixer_info
    params:
      attribute_bindings:
        mixer_version: version
        mixer_build_id: buildID
        mixer_build_status: buildStatus
...
  - kind: attributes
    adapter: k8s_pod_info
    params:
      input_expressions:
        srcIP: source.ip | "unknown"
        tgtIP: target.ip | "unknown"
      attribute_bindings:
        source_name: srcName
        target_name: tgtName
...
```
NOTE: the new aspect kind is named 'attributes' to match the existing
style of the aspect kind names. That is, we don't call the access
logging aspect kind 'access-logs-generator', nor do we call the metrics
reporting aspect kind 'metrics-reporter'. To match, we call the
attribute generation aspect kind simply 'attributes'. However, this
could be slightly confusing in the implementing code (too many things
called Attributes), so the manager and aspect interface are in terms
of AttributesGenerator.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/517)


douglas-reid:
Please hold review. As discussed, I will be refactoring the config story a bit.

douglas-reid:
OK. Configuration story has been adjusted. Please take a look when you have a chance.

sebastienvas:
Jenkins rebuild mixer/manager-regression

pkg/aspect/config/attrgen.proto
```diff
@@ -0,0 +1,38 @@
...
+    // Map of input params name to attribute expressions. At runtime, each
+    // expression will be evaluated to determine the input value provided to the
+    // aspect.
+    map<string, string> input_expressions = 1;
+
+    // Map of generated value names to mixer attribute names. Individual aspect
+    // implementations will produce a map of output values (name -> value).
+    // These output values will be converted into attributes for consumption by
+    // the rest of the mixer via this map.
+    map<string, string> value_attribute_map = 2;
```
ZackButcher: I think the output should be described via attribute descriptors: we need to know the types of the output parameters for config validation. Then the output can be `map[string]interface{}` but correspond to those descriptors. No need to map from output to descriptor names if we just require they output the descriptor names they claim.

pkg/config/kind.go
```diff
@@ -35,24 +35,24 @@ const (
...
-	AccessLogsKindName          = "access-logs"
-	ApplicationLogsKindName     = "application-logs"
-	AttributeGenerationKindName = "attribute-generation"
-	DenialsKindName             = "denials"
-	ListsKindName               = "lists"
-	MetricsKindName             = "metrics"
-	QuotasKindName              = "quotas"
+	AccessLogsKindName      = "access-logs"
+	ApplicationLogsKindName = "application-logs"
+	AttributesKindName      = "attributes"
```
ZackButcher: What about "preprocessors" or something in that vein? I dislike "attributes" because, as you point out, I think it's likely to cause confusion. Plus, it's not strictly required that an adapter running in the preprocess phase generates attributes.

pkg/aspect/attrgenmgr.go
```diff
@@ -0,0 +1,106 @@
...
+func (attrGenMgr) Kind() config.Kind {
+	return config.AttributesKind
+}
+
+func (attrGenMgr) DefaultConfig() (c config.AspectParams) {
+	// NOTE: The default config leads to the generation of no new attributes.
+	return &apb.AttributeGeneratorsParams{}
+}
+
+func (attrGenMgr) ValidateConfig(c config.AspectParams, v expr.Validator, df descriptor.Finder) (cerrs *adapter.ConfigErrors) {
```
ZackButcher: We need to make sure `input_expressions` has valid expressions here too (with `v.TypeCheck`).

pkg/aspect/config/attrgen.proto
ZackButcher: And I think we should require those attributes as part of _aspect_ config, otherwise we'd have to require each user adapter yields up a list of attribute descriptors some other way.

douglas-reid: This was how the original PR was setup. After discussions with @mandarjog, we decided it was better to have the descriptors be part of the global config and have their names referenced in the aspect config.

This choice makes it so that adapters don't have to know anything about how they are mapped into descriptors and so that operators can look at global config to find the full definition of attributes. I believe that having the descriptors be global matches the rest of the config story and makes validation easier too.

pkg/config/kind.go
douglas-reid: they could be other preprocess aspect kinds. this is just _a_ kind of processor that generates attributes. when we identify other proprocess needs, they would be a separate preprocess aspect, with a different kind name.

pkg/aspect/attrgenmgr.go
douglas-reid: ok. i can add that. worth mentioning that we aren't doing type checking on the way out. not sure how big of a deal that could end up being.

pkg/aspect/config/attrgen.proto
ZackButcher: That works

pkg/aspect/attrgenmgr.go
ZackButcher: We don't do any typechecking at runtime, but we do at validation time: given the set of attributes we can type the expression, then we know at runtime the values are always what we expect (barring something like receiving an invalid attribute).

douglas-reid: Understood. Just pointing out we now have one weakness in our system. An adapter value could be erroneously wired into an attribute that has a different valuetype in the descriptor.

I've updated `ValidateConfig` to check the expression types.

douglas-reid: Would you mind clarifying what you mean by "describe the attributes aspect output"?

We will be adding descriptors in config for produced attributes. The `value_attribute_map` will describe how to populate those descriptors (referencing attributes by their descriptor names). I believe this is similar to the other aspects.

We could change from `value_attribute_map` to something like:

```
message AttributesGeneratorParams {
...
message AttributeBinding {
  string attribute_descriptor_name = 1;
  string adapter_value_name = 2;
}

repeated AttributeBinding attributes = 2;
}
```
if you think that would be clearer.

pkg/aspect/attrgenmgr_test.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	notFoundAttr := &apb.AttributeGeneratorsParams{
+		ValueAttributeMap: map[string]string{
+			"src_pod_ip": "not_found",
+		},
+	}
+
+	m := newAttrGenMgr()
+	if err := m.ValidateConfig(validExpr, expr.NewCEXLEvaluator(), dfind); err != nil {
```
douglas-reid: Done.

pkg/aspect/attrgenmgr.go
ZackButcher: Basically what we discussed out of band of the github review: a `Manifest` of some type that lists, with `AttributeDescriptor`s, the attributes that this adapter produces. I'm fine keeping `map<string, string>` for now since we'll change it later.

douglas-reid:
PTAL. I've updated the aspect config a bit to use a repeated `AttributeBinding`s to map between attributes and adapter output values. This should add more clarity on how attributes are generated and hopefully reduce mapping errors. It would be trivial to inline descriptors later, if that proved desirable.

pkg/aspect/config/attrgen.proto
```diff
@@ -0,0 +1,52 @@
...
+
+import "gogoproto/gogo.proto";
+
+option go_package="config";
+option (gogoproto.goproto_getters_all) = false;
+option (gogoproto.equal_all) = false;
+option (gogoproto.gostring_all) = false;
+
+// Configures an AttributesGenerator aspect.
+message AttributesGeneratorParams {
```
mandarjog: Can you put a yml example of this along with the (no yet written) k8s adapter ?

mandarjog:
For the k8s adapter usecase, which is what we *really* care about, let's just add those attribute descriptors to the vocabulary via globalconfig file. That way *we* do get the proper validation for the k8s environment.

geeknoid:
Regarding the config story, I was expecting something quite a bit simpler...

On the input side, I'm surprised to see these expression maps. I was expecting these generator adapters to just be given all the incoming attributes. I realize this would break our "adapters don't see attributes" rule, but given these adapters are geared towards attribute manipulation, it would seem like a reasonable approach.

On the output side, do we really need all the flexibility that this attribute binding stuff is providing? I was expecting these attribute generators to just provide an attribute manifest detailing what they produce and that's all there is. The operator doesn't get to decide which attributes the proxy outputs, and also wouldn't get to decide what attributes these generators produce.

If we are in fact convinced that there must be attribute binding happening, then would it be possible to express it in a way identical to the expression map? Specifically:

```
 aspects:
  - kind: attributes
    adapter: mixer_info
    params:
      attribute_bindings:
        version: mixer_version
        buildID: mixer_build_id
        buildStatus: mixer_build_status
...
 - kind: attributes
    adapter: k8s_pod_info
    params:
      input_expressions:
        srcIP: source.ip | "unknown"
        tgtIP: target.ip | "unknown"
      attribute_bindings:
        srcName: source_name
        tgtName: target_name
```

Finally, a question. Does this produce errors if the operator tries to overwrite an existing attribute within the attribute binding? If so, does it fail at runtime or config time?

pkg/aspect/config/attrgen.proto
geeknoid: This schema doesn't look right to me.

The adapter knows exactly the set of input values it needs and knows exactly the set of output values it produces. Can't the map and attribute bindings be replaced by actual messages in the adapter's config proto?

douglas-reid:
RE: Config

I don't believe we want these adapters to be drastically different than the other adapters. It seems beneficial to have the config story be mostly uniform. This is why there is an input expressions list. Individual adapters should not need the full set of attributes. They shouldn't care what attributes may or may not come with a given proxy, etc.

I had the attribute_bindings as a map originally, but there was a request to make it clearer (and to more closely match the config for the rest of the system). The current scheme with `descriptor_name` makes it more obvious, I think, what is happening.

pkg/aspect/config/attrgen.proto
douglas-reid: One adapter may have a completely different set of inputs from a second adapter that _also_ implements the `attributes` aspect. We can't use an actual message for inputs because they will be different. The same is true for outputs.

At the _adapter_ level, an adapter takes a map[string]interface{} and produces a map[string]interface{}. The manager, using _aspect_ config translates in both directions.

The example in the PR description provides an example. A mixer version adapter and a k8s pod info adapter take different inputs and produce different outputs.

ZackButcher:
I agree with Doug: inputs into these adapters should be expressions, like all the other adapters. I also agree with @geeknoid that they should have a `Manifest` of output attributes. The problem is that right now `Manifest` doesn't exist. I LG'd with the understanding that we'll shift to using a `Manifest` to describe APA output when we define that type.

geeknoid:
Could the adapter use a proto message and have the aspect manager use reflection? 

The usability of the current config model is not great IMO.

douglas-reid:
I do think the config model usability could stand some improvement. But I'm not sure reflection is the cure. I thought about reflection (and maybe generating some sort of struct on the fly from config to be populated by the adapter for output), but it just seems icky and overly-complex -- and doesn't really save you any config. You'd still have to tell the aspect manager how to map into/out of the protos.

Right now, adapters take a set of name-value pairs for input (similar to labels in the other aspects) and produce a set of name-value pairs. The aspect manager knows how to update the attribute bag given the bindings and it is up to operators to decide how/what to provide. This has the advantage of maintaining a clean separation between adapters and the attribute model.

We could limit the expressiveness of the config for these adapters (no expressions in inputs, etc.), but I'm not sure we need or want that simplification at the moment.

geeknoid:
I exactly understand how things work, and I was staring at the example
config for a minute before figuring out what was going on. Maybe it was too
early in the morning, but I took it as a red flag about usability. If I
can't figure out what the config is doing/intending, then an operator
stands no chance.

I do think the input side of things is OK, I'm worried about the output
side, the attribute binding stuff.

Regarding reflection, perhaps I'm missing some subtlety. I am suggesting
that the adapter declares a message in its config proto that has one field
per output value. The adapter would return this filled in struct to the
aspect manager from the Execute method. The aspect manager would reflect
over this struct and get all the individual names & values that it can then
inject into the attribute bag. Seems reasonably simple. Am I missing some
complication?

douglas-reid:
Having the adapter fill in a self-specified struct doesn't really buy us much.  I think you are still going to want to have the operator control what attributes are allowed in the system in general (and at a finer grain than including an adapter or not).

Imagine two adapters that produce values with the exact same name. We would be preventing their combined usage if we didn't allow config to do mapping from adapter value to mixer attribute. So, I envision operators still having to provide configuration similar to what is proposed here.

This scheme empowers the operator to choose which of the values produced by an adapter are propagated out further and avoid collisions. And it saves the manager code from having to introspect over adapter config (which we've typically restricted access to at the //adapter/impl package level).

I think we can probably optimize the expression of config long-term. I started with full attribute descriptors in the config. After some discussion, I removed them and replaced it with a map similar to the one you suggest. After some more discussion, we ended up with the current scheme. So, it seems there are many ideas for what is both simple and clear.

However, I don't expect this to appear in more than once place (it has to be global) and I imagine the config won't change much over time (or even across clusters) in the early days. Maybe having it be slightly awkward (in a similar vein to the *logs aspect configs) is OK for alpha?

ZackButcher:
I don't think using reflection to fill in some adapter-defined output message is needed or gains us much. With this PR the story is "put the attributes you produce into the big global list of all attributes, and in your adapter config map from whatever output name you want to an attribute name". When we implement Manifests it changes to "put the attributes you produce in your Manifest, put your manifest in the big global list of manifests".

Looking back over the PR, I do think we need to add validation that multiple adapters aren't trying to claim the same output attribute name.

douglas-reid:
There isn't really a good way (as far as I can tell), in the scope of this PR, to ensure that an operator has configured the different aspect instances to not use the same name. The manager doesn't have access to the full set of configs for the system and, as a result, can't tell that a binding exists elsewhere. That logic belongs at a higher level of config validation.

In the short term, this isn't a huge issue -- as the config will live side-by-side and we will have a very small number of attribute generation aspects configured.

ZackButcher:
Yeah, that validation is substantially easier when we introduce Manifests.

douglas-reid:
Thanks for all the good feedback!

PTAL. I've changed `attribute_bindings` to a map from attribute descriptor name to value name, as discussed. I've also added more detailed documentation to the proto file.

I've also filed #534 to track improvements to config etc to support better input/output story.

PR #518, opened 2017-04-07
Fix testdata serviceconfig's quota setup

ZackButcher:
I forgot to add quota aspect params when I updated the quota aspect to use descriptors, so while a quota was defined in the globalconfig at runtime the quota adapter thought it had no quotas.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/518)

### DIFF
--- a/testdata/serviceconfig.yml
+++ b/testdata/serviceconfig.yml
@@ -7,6 +7,8 @@ rules:
   aspects:
   - kind: quotas
     params:
+      quotas:
+      - descriptor_name: RequestCount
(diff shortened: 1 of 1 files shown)


ZackButcher:
Ug, would not expect mixer-pr-master to break due to this. One really annoying side effect of that as the e2e test is that we don't get to see the logs from the components in the jenkins output, since k8s hides it all away.

sebastienvas:
jenkins rebuild istio/mixer-pr-master

PR #549, opened 2017-04-14
Move quota max_amount and expiration from QuotaDescriptor to QuotaParams

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/549)

### DIFF
--- a/pkg/aspect/config/quotas.proto
+++ b/pkg/aspect/config/quotas.proto
@@ -17,6 +17,7 @@ syntax = "proto3";
 package pkg.aspect.config;
 
 import "gogoproto/gogo.proto";
+import "google/protobuf/duration.proto";
 
 option go_package="config";
 option (gogoproto.goproto_getters_all) = false;
@@ -32,7 +33,15 @@ message QuotasParams {
         // Map of quota_descriptor label name to attribute expression. At run time each
         // expression will be evaluated to determine the value provided to the aspect. The
         // result of evaluating the expression must match the ValueType of the label in the quota_descriptor.
-        map<string, string> labels = 3;
+        map<string, string> labels = 2;
+
+        // The upper limit for this quota.
+        int64 max_amount = 3;
+
+        // The amount of time allocated quota remains valid before it is
+        // automatically released. This is only meaningful for quotas
+        // annotated as rate limits, otherwise the value must be zero.
+        google.protobuf.Duration expiration = 4 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
     }
 
     // The set of quotas that will be populated and handed to aspects at run time.
--- a/pkg/aspect/quotasManager.go
+++ b/pkg/aspect/quotasManager.go
@@ -17,7 +17,6 @@ package aspect
 import (
 	"fmt"
 
-	ptypes "github.com/gogo/protobuf/types"
 	"github.com/golang/glog"
 	rpc "github.com/googleapis/googleapis/google/rpc"
 
@@ -62,6 +61,9 @@ func (m *quotasManager) NewQuotaExecutor(c *cpb.Combined, a adapter.Builder, env
 		// We don't check the err because ValidateConfig ensures we have all the descriptors we need and that
 		// they can be transformed into their adapter representation.
 		def, _ := quotaDefinitionFromProto(df.GetQuota(quota.DescriptorName))
+		def.MaxAmount = quota.MaxAmount
+		def.Expiration = quota.Expiration
+
 		defs[def.Name] = def
 		metadata[def.Name] = &quotaInfo{
 			definition: def,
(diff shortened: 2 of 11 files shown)


pkg/aspect/config/quotas.proto
```diff
@@ -32,7 +33,15 @@ message QuotasParams {
         // Map of quota_descriptor label name to attribute expression. At run time each
         // expression will be evaluated to determine the value provided to the aspect. The
         // result of evaluating the expression must match the ValueType of the label in the quota_descriptor.
-        map<string, string> labels = 3;
+        map<string, string> labels = 2;
+
+        // The upper limit for this quota.
+        int64 max_amount = 3;
```
ZackButcher: Is this `max_amount` or `default_amount`? With a different quota impl we'll do things like support quota overrides per user, right?

pkg/aspect/quotasManager.go
```diff
@@ -98,6 +100,24 @@ func (*quotasManager) ValidateConfig(c config.AspectParams, v expr.Validator, df
...
+		if quota.MaxAmount < 0 {
+			ce = ce.Appendf("MaxAmount", "must be >= 0")
+		}
+
+		if quota.Expiration < 0 {
+			ce = ce.Appendf("Expiration", "cannot be less than 0")
+		}
+
+		if desc.RateLimit {
+			if quota.Expiration == 0 {
```
ZackButcher: Is there a reason not to combine these conditions to:  
```go
if desc.RateLimit && (quota.Expiration == 0) {
  error
} else if quota.Expiration != 0 {
  error
}
```

pkg/aspect/config/quotas.proto
```diff
@@ -32,7 +33,15 @@ message QuotasParams {
         // Map of quota_descriptor label name to attribute expression. At run time each
         // expression will be evaluated to determine the value provided to the aspect. The
         // result of evaluating the expression must match the ValueType of the label in the quota_descriptor.
-        map<string, string> labels = 3;
+        map<string, string> labels = 2;
```
douglas-reid: is this safe because we don't have any current uses of labels in config?

testdata/serviceconfig.yml
```diff
@@ -7,6 +7,8 @@ rules:
     params:
       quotas:
       - descriptor_name: RequestCount
```
ZackButcher: You know, in the spirit of the other changes we've been making to the config API, moving from messages with an embedded name field to a `map<name, other fields>`, should we do the same with our aspect params too?

```yaml
- selector: true
  aspects:
  -kind: quotas
    params:
      RequestCount:
        max_amount: 5
        expiration: 1s
      FooBar:
        max_amount: 1000000
        expiration: 1s
```

Though maybe this goes too far and makes it non-obvious that this is a reference to the "RequestCount" and "FooBar" quota descriptors.

pkg/aspect/config/quotas.proto
ZackButcher: Proto field compatibility only matters when there are multiple binaries that depend on the proto definitions. Since no other binaries have been compiled that depend on our config APIs, we can change field numbers without breaking things. Even better, since our config is YAML, it doesn't care about field numbers so nothing breaks in our yaml ingestion either.

Once istioctl talks to our config API and has a dependency on our protos we can't shuffle these fields anymore.

testdata/serviceconfig.yml
mandarjog: I think this is too far.

pkg/aspect/config/quotas.proto
mandarjog: I think it should be named `amount` (i.e. ratelimit).    (`max_amount` means the same thing in this context)
We have an orthogonal mechanism to decide which limit finally applies.

This message is what one authority (modeled as scope) would like the quota to be.

geeknoid: I want to leave this as max_amount. Yes, at some point there will be ways to override this value. No matter what this particular field though, it won't convey the fact it can be overridden.

Calling it just "amount" seems confusing to me relative to actual in-use quota amounts.

pkg/aspect/quotasManager.go
geeknoid: I simplified it a bit. But I don't like your form, it requires a tiny bit too much thinking to make sure it's correct. I rather have a few more characters for extra ease of readability.

geeknoid:
PTAL.

pkg/aspect/config/quotas.proto
douglas-reid: I'm still confused on what this field is conveying.

Is this the limit that should be applied for this <scope,subject> (and subject to some effective quota calculation across all <scope,subject>s that apply)? 

Or is this saying that the upper bounds on any limit at this <scope,subject> is X (though there may be lower limits at this <scope,subject>)?

Put another way, is this like `billable_daily_max` or is it just `billable_daily`?

pkg/aspect/quotasManager.go
douglas-reid: You might consider something like:

```
func validExpiration(expire time.Duration, isRateLimit bool) bool {
  if !isRateLimit {
     return expire == 0
  }
  return expire > 0
}

...

if validExpiration(quota.Expiration, descr.RateLimit) {
   ce = ce.Appendf(...)
...
}

geeknoid: Then I'd lose the specific error message...

pkg/aspect/config/quotas.proto
geeknoid: We just don't have the scoping story sorted out.

As it is, the quotasManager will be called with one of these config structs filled-in and will use max_amount as the limit for any quota of the given name. This means that the config will have been resolved to the appropriate <scope, subject>. But the code has no visibility to the hierarchy, so there's no way to enforce a single quota with limits at different levels of the hierarchy.

Similarly, we don't have a story for per-user overrides, beyond using different aspects for different users, which I think is lame.

Solving the hierarchical resolution and overrides in a clean way is a top concern post-Alpha. We need to have something that plays nice in this place that can clearly be understood by our operators and can be handled cleanly in scripting.

mandarjog: This is the number of quota configs quota manager should get.

1. most specific subject within every scope
2. quota combine step which will be given all these quotas to combine.
3. dispatch those quotas 

What visibility are you looking for in inside the quota manager ?

geeknoid:
Right now, the quota manager only receives a single chunk of config. Steps
2 and 3 don't exist as far as I know.

geeknoid:
PTAL.

PR #553, opened 2017-04-15
proto jsonp parsing for descriptors 

mandarjog:
Descriptors are now parsed using the same method as aspects and adapters.
That ensures camelCase and snake_case both work.

It also ensures that we can use symbolic enum names.
symbolic enums can be used in most cases, except when they are inside a proto map.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/553)

### DIFF
--- a/pkg/config/kind.go
+++ b/pkg/config/kind.go
@@ -85,14 +85,17 @@ func (ks KindSet) Set(k Kind) KindSet {
 	return ks | (1 << k)
 }
 
+// make gas happy
+func ignoreErrors(args ...interface{}) {}
+
 func (ks KindSet) String() string {
 	buf := pool.GetBuffer()
 	defer pool.PutBuffer(buf) // fine to pay the defer overhead; this is out of request path
 
-	fmt.Fprint(buf, "[")
+	ignoreErrors(fmt.Fprint(buf, "["))
 	for k, v := range kindToString {
 		if ks.IsSet(k) {
-			fmt.Fprintf(buf, "%s, ", v)
+			ignoreErrors(fmt.Fprintf(buf, "%s, ", v))
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/config/kind.go
```diff
@@ -85,14 +85,17 @@ func (ks KindSet) Set(k Kind) KindSet {
...
 
+// make gas happy
+func ignoreErrors(args ...interface{}) {}
+
 func (ks KindSet) String() string {
 	buf := pool.GetBuffer()
 	defer pool.PutBuffer(buf) // fine to pay the defer overhead; this is out of request path
 
-	fmt.Fprint(buf, "[")
+	ignoreErrors(fmt.Fprint(buf, "["))
```
geeknoid: Are u using different linter versions? 'cause why is this surfacing now?

Also, just using _, _ = fmt.Fprintf is not sufficient?

testdata/globalconfig.yml
```diff
@@ -20,36 +20,45 @@ manifests:
...
     # TODO: we really need to remove these, they're not part of the attribute vocab.
     - name: api.name
-      value_type: 1 # STRING
+      value_type: STRING
     - name: api.method
-      value_type: 1 # STRING
+      value_type: STRING
+
+# Enums as struct fields can be symbolic names.
+# However enums inside maps *cannot* be symbolic names.
```
geeknoid: This is due to a bug/limitation in the parser and is not by design, right?

pkg/config/kind.go
mandarjog:
```
pkg/aspect/metricsManager.go:160:22:warning: unnecessary conversion (unconvert)
pkg/aspect/quotasManager.go:167::error: arg name for printf verb %s of wrong type: int (vet)
pkg/config/kind.go:92::warning: Errors unhandled.,LOW,HIGH (gas)
pkg/config/kind.go:95::warning: Errors unhandled.,LOW,HIGH (gas)
pkg/config/kind.go:105::warning: Errors unhandled.,LOW,HIGH (gas)
```
This is after ensuring `prep_linters` does all the `go gets`.
We need to pin linters to a version.

testdata/globalconfig.yml
mandarjog: Yes, this is due to limitation of `jsonp`

PR #557, opened 2017-04-16
Sanitize error messages

geeknoid:
- Remove redundant use of "with err" in many error messages. These
were leading to generally confusing output.

- Try to use %v to format errors instead of %s.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/557)

### DIFF
--- a/adapter/statsd/statsd.go
+++ b/adapter/statsd/statsd.go
@@ -79,7 +79,7 @@ func (b *builder) ValidateConfig(c adapter.Config) (ce *adapter.ConfigErrors) {
 	}
 	for metricName, s := range params.MetricNameTemplateStrings {
 		if _, err := template.New(metricName).Parse(s); err != nil {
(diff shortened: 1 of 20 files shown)


adapter/statsd/statsd.go
```diff
@@ -79,7 +79,7 @@ func (b *builder) ValidateConfig(c adapter.Config) (ce *adapter.ConfigErrors) {
 	}
 	for metricName, s := range params.MetricNameTemplateStrings {
 		if _, err := template.New(metricName).Parse(s); err != nil {
-			ce = ce.Appendf("MetricNameTemplateStrings", "failed to parse template '%s' for metric '%s' with err: %s", s, metricName, err)
+			ce = ce.Appendf("MetricNameTemplateStrings", "failed to parse template '%s' for metric '%s' : %v", s, metricName, err)
```
douglas-reid: do we need the space after `'%s'` and before the ':' (here and elsewhere) ?

geeknoid: Oops, bad search-and-replace.

Fixed. PTAL.

PR #580, opened 2017-04-19
Add canonicalization of service names to k8s adapter

douglas-reid:
This adds the ability to take an indication of target service and canonicalize it (and other service names) to the kubernetes adapter.

To support this canonicalization, two additional input config params are added to the config.proto:
1. target_service_input_name
1. cluster_domain_name

These new params allow the operator to configure how canonicalization, particularly of target service will work. Example config for using these params (with their default values) can be seen in the updated serviceconfig.yml included in this PR.

This PR also addresses a few minor issues in the kubernetes adapter that were raised, but delayed, from the initial PR (debug-only logging and the need for better stop chan handling in the debug logger).

In addition to unit testing, this PR was tested via cli, as follows:

```shell
./bazel-bin/cmd/server/mixs server --globalConfigFile=testdata/globalconfig.yml --serviceConfigFile=testdata/serviceconfig.yml --alsologtostderr -v=9
```

```shell
$ ./bazel-bin/cmd/client/mixc report -a target.service=svc.cluster.local,request.headers="authority:foo.maze:999" --string_attributes=source.uid="kubernetes://hello-2992298478-qv9f3.ratelimit"
Report RPC returned OK
  Attribute              Type               Value
  source.labels          map[string]string  map[app:hello pod-template-hash:2992298478 version:v1]
  source.name            string             hello-2992298478-qv9f3
  source.namespace       string             ratelimit
  source.service         string             hello.ratelimit.svc.cluster.local
  source.serviceAccount  string             default
  target.service         string             foo.maze.svc.cluster.local

```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/580)

### DIFF
--- a/adapter/kubernetes/kubernetes.go
+++ b/adapter/kubernetes/kubernetes.go
@@ -339,26 +339,28 @@ func canonicalName(service, namespace, clusterDomain string) (string, error) {
 	splits := strings.SplitN(service, ":", 2)
 	s := splits[0]
 	if len(s) == 0 {
-		return "", errors.New("invalid service name: starts with ':'")
+		return "", fmt.Errorf("invalid service name '%s': starts with ':'", service)
 	}
 	// error on ip addresses for now
 	if ip := net.ParseIP(s); ip != nil {
 		return "", errors.New("invalid service name: cannot canonicalize ip addresses at this time")
 	}
-	parts := strings.SplitN(s, ".", 5)
+	parts := strings.SplitN(s, ".", 3)
 	if len(parts) == 1 {
 		return fmt.Sprintf("%s.%s.%s", parts[0], namespace, clusterDomain), nil
 	}
 	if len(parts) == 2 {
 		return fmt.Sprintf("%s.%s", s, clusterDomain), nil
 	}
-	if len(parts) == 3 {
-		dom := strings.SplitAfterN(clusterDomain, ".", 2)
-		return fmt.Sprintf("%s.%s", s, dom[1]), nil
+
+	domParts := strings.Split(clusterDomain, ".")
+	nameParts := strings.Split(parts[2], ".")
+
+	if len(nameParts) >= len(domParts) {
+		return s, nil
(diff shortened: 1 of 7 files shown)


adapter/kubernetes/kubernetes.go
```diff
@@ -260,17 +289,76 @@ func addPodValues(m map[string]interface{}, prefix string, params config.Params,
...
+	splits := strings.SplitN(service, ":", 2)
+	s := splits[0]
+	if len(s) == 0 {
+		return "", errors.New("invalid service name: starts with ':'")
+	}
+	// error on ip addresses for now
+	if ip := net.ParseIP(s); ip != nil {
+		return "", errors.New("invalid service name: cannot canonicalize ip addresses at this time")
+	}
+	parts := strings.SplitN(s, ".", 5)
```
mandarjog: The kube dns domain specified the suffix = `svc.cluster.local`, but it could be longer than that. mixer get this as a command line parameter.

adapter/kubernetes/kubernetes.go
```diff
@@ -260,17 +289,76 @@ func addPodValues(m map[string]interface{}, prefix string, params config.Params,
...
+// "hello.default.svc.cluster.local",
+func canonicalName(service, namespace, clusterDomain string) (string, error) {
+	if len(service) == 0 {
+		return "", errors.New("invalid service name: cannot be empty")
+	}
+	// remove any port suffixes (ex: ":80")
+	splits := strings.SplitN(service, ":", 2)
+	s := splits[0]
+	if len(s) == 0 {
+		return "", errors.New("invalid service name: starts with ':'")
```
ZackButcher: `fmt.Errorf("invalid service name '%s': starts with ':'", s)`

douglas-reid:
PTAL.  Updated:
-  now handles arbitrary length cluster domain names
-  better error message for names starting with `:`

PR #593, opened 2017-04-21
Remove remaining instances of 'the mixer' and 'the istio mixer'

ZackButcher:
These are the only remaining ones outside of generated code according to `grep -irnw . -e "the mixer" -e "the istio mixer"` (from mixer root).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/593)


ZackButcher:
I didn't remove all instances of "proxy" as our rules are as clear about Envoy vs proxy, but I changed a few of those as well.

PR #598, opened 2017-04-21
update mixer build to use entrypoint and cmd

mandarjog:
1. This PR splits command into entry point and CMD, so the podspec can override it with args.
2. Bring the configroot/* files up to date with globalconfig.yml and serviceconfig.yml

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/598)

### DIFF
--- a/testdata/configroot/scopes/global/subjects/global/rules.yml
+++ b/testdata/configroot/scopes/global/subjects/global/rules.yml
@@ -1,36 +1,26 @@
 subject: namespace:ns
 revision: "2022"
 rules:
-        #- selector: service.name == “*”
-        #- selector: service.name == "myservice"
-- selector: target.service == "myservice.*"
+- selector: # must be empty for preprocessing adapters
   aspects:
-  - kind: access-logs
+  # when running local without a kubeconfig file specified in globalconfig,
+  # this aspect should be commented out. It is only needed when the attributes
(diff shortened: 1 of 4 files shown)


testdata/configroot/scopes/global/subjects/global/rules.yml
```diff
@@ -1,36 +1,26 @@
 subject: namespace:ns
 revision: "2022"
 rules:
-        #- selector: service.name == “*”
-        #- selector: service.name == "myservice"
-- selector: target.service == "myservice.*"
+- selector: # must be empty for preprocessing adapters
```
ZackButcher: Not related to this PR, but we need to come up with a better way to do this. I kinda liked the `true` literal personally. I know it's a bit harder to implement with our current config resolution.

ZackButcher:
Oh, and it'd be nice to pull the k8s adapter attributes out of the proxy's manifest and into their own manifest, but not blocking.

douglas-reid:
Overall, this looks good.

Somewhat-related questions: 
- should we also update the latency metrics to match #599 ?
- what is the recommended procedure for updating globalconfig for a running mixer now?  Will that still be configmap'd somehow?
- do we need to update the demo configs, or can we point them at these somehow?

PR #610, opened 2017-04-24
Update metrics/logs configuration in testdata

douglas-reid:
This PR attempts to update the configs that deploy with mixer to aid in use for the bookinfo demo (as well as any follow-on Task docs that will be written). It aims to be flexible with observed behavior (`response.latency ` attribute added to avoid `0ms` durations being continually reported). It also adds some more fallbacks to provide more "interesting" data in collected metrics.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/610)

### DIFF
--- a/pkg/expr/func.go
+++ b/pkg/expr/func.go
@@ -231,7 +231,7 @@ func newIndex() Func {
 	return &indexFunc{
 		baseFunc: &baseFunc{
 			name:     "INDEX",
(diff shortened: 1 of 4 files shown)


douglas-reid:
@ZackButcher let's do that in a follow-on PR. That has a big enough scope to be separate.

PR #622, opened 2017-04-25
Add mutex in unit test.

chowchow316:
Passed local racetest with #615 .

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/622)

### DIFF
--- a/adapter/redisquota/redisquota_test.go
+++ b/adapter/redisquota/redisquota_test.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
@@ -41,6 +42,8 @@ type testServer struct {
 	errOnStart bool
 
 	errOnIntRet bool
+
+	sync.Mutex
 }
 
 func (t *testServer) Start() (*miniredis.Miniredis, *redeo.Server, error) {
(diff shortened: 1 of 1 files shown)


adapter/redisquota/redisquota_test.go
```diff
@@ -478,7 +481,9 @@ func TestBadConfig(t *testing.T) {
 
 func TestErrorResponse(t *testing.T) {
 	mockredis := testServer{errOnStart: false, errOnIntRet: true}
+	mockredis.Mutex.Lock()
```
douglas-reid: I'm surprised this works. What state is this actually protecting? Is this mockredis used anywhere else?

I would have expected some solution around:
- EITHER setting the port in the server address to ":0" (instead of "localhost:9736")
- OR having some global mutex that only allows one test at a time that sets up the `errOnIntRet` server.

chowchow316: There is only one `errOnIntRet` server for now, in order to cover the error handling code (basically to improve the code coverage). But the coverage is still not satisfying... May need better solution.
My understanding is that it is enforcing that there is only one server started at this time with localhost:9736 (correct me if I'm wrong). ":0" seems do not accepted by the redeo server setting. But actually I can remove the port number here, and it will use `0.0.0.0:9736` as default config. 
I'm also wondering if grafana is using any specific port?

chowchow316:
This seems not needed any more, and the mockredis server will probably be re-designed. Close it for now. Please let me know if you have any concern, and we can re-open this. @douglas-reid

PR #624, opened 2017-04-26
Add descriptors for new metrics to global descriptors

douglas-reid:
This PR adds a couple of new descriptors to the global config for mixer (in testdata/configroot).

Notably, these are not added to the rules for any scope/subject. This is by design at this point (to allow for adding them as part of a Task on configuring metrics -- since you there is no way for adding metric descriptors currently supported).

The new metrics are:
- request_size
- response_size

Some of the other descriptors are also updated to support validation.

Tested with minikube (pushed a local docker image in place of mixer, updated via `istioctl mixer` commands).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/624)

### DIFF
--- a/testdata/configroot/scopes/global/descriptors.yml
+++ b/testdata/configroot/scopes/global/descriptors.yml
@@ -52,6 +52,8 @@ manifests:
       value_type: STRING_MAP
     - name: request.scheme
       value_type: STRING
+    - name: request.size
+      value_type: INT64
     - name: response.size
       value_type: INT64
     - name: response.code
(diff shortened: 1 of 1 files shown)


testdata/configroot/scopes/global/descriptors.yml
```diff
@@ -100,12 +102,57 @@ metrics:
...
       response_code: 2 # INT64
+  - name: request_size
+    kind: DISTRIBUTION
+    value: INT64
+    description: request size by source, target, and service
+    buckets:
+      exponential_buckets:
+        num_finite_buckets: 8
+        scale: 1
+        growth_factor: 10
```
ZackButcher: Nit: Use camelCase field names here and elsewhere now that we can parse them correctly. Not required for the label names, though we should probably decide what our convention is w.r.t. label names - I lean towards keeping everything camelCased.

douglas-reid: done.

PR #625, opened 2017-04-26
Update the symbol names for all ConfigErrors to adopt camelCase consi…

geeknoid:
…stently.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/625)

### DIFF
--- a/out.txt
+++ b/out.txt
@@ -1,0 +1,25 @@
+//adapter/denyChecker:small_tests                               (cached) PASSED in 0.1s
+//adapter/genericListChecker:small_tests                        (cached) PASSED in 0.1s
+//adapter/ipListChecker:small_tests                             (cached) PASSED in 0.2s
+//adapter/kubernetes:small_tests                                (cached) PASSED in 1.6s
+//adapter/memQuota:small_tests                                  (cached) PASSED in 0.1s
+//adapter/prometheus:small_tests                                (cached) PASSED in 0.2s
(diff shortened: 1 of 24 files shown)


ZackButcher:
Wait, I thought we were using proto field names (so snake_case)?

geeknoid:
I'm pretty sure we settle on camelCase for our docs, which is what I
changed everything to last week. So these field names should be consistent
with that I think. Did we agree on snake_case and I'm misremembering?

ZackButcher:
The only thing I recall is that we wanted field names to match exactly to make debugging easier. However, I'm also planning on transforming our auto-gen'd docs to use camelCase. The only place snake_case names will exist is in the .proto files. Given that, this is fine I think.

out.txt
```diff
@@ -0,0 +1,25 @@
+//adapter/denyChecker:small_tests                               (cached) PASSED in 0.1s
```
douglas-reid: please remove from PR.

douglas-reid:
I had thought we decided to go with snake_case, too (which explains all those lines in kubernetes.go you had to change). I really don't care much either way. However, I would like to have this definitively decided for moving forward.

Can we add some doc somewhere (maybe in ConfigErrors) that indicates (or even enforces) the expected style?

kyessenov:
camelCase is standard in JSON/YAML and that's what the operator sees. That's how k8s YAMLs look like. Unless there is a reason to go with the snake case, let's utilize camelCase consistently.

PR #632, opened 2017-04-27
Documented but unimplemented routes should return 501 not 404

esnible:
The routes documented by Swagger should return 501 instead of 404 because it is not a client error to invoke them.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/632)

### DIFF
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -72,6 +72,22 @@ func (a *API) register(c *restful.Container) {
 	ws.Produces(restful.MIME_JSON)
 	ws.Path(a.rootPath)
 
+	// List the scopes
+	ws.Route(ws.
+		GET("/scopes").
+		To(a.getScopes).
+		Doc("Gets scopes associated with Mixer").
+		Writes(APIResponse{}))
+
+	// Create a policy
+	ws.Route(ws.
+		POST("/scopes/{scope}/subjects/{subject}").
+		To(a.createPolicy).
+		Doc("Create a policy").
+		// TODO Reads(...).
+		Writes(APIResponse{}))
+
+	// List the rules for a scope and subject
 	ws.Route(ws.
 		GET("/scopes/{scope}/subjects/{subject}/rules").
 		To(a.getRules).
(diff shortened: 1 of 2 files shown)


pkg/config/apiserver.go
```diff
@@ -88,6 +104,58 @@ func (a *API) register(c *restful.Container) {
...
 		Writes(APIResponse{}))
+
+	// Creates or replaces a rule's list of aspects
+	ws.Route(ws.
+		PUT("/scopes/{scope}/subjects/{subject}/rules/{ruleid}/aspects/{aspect}").
+		To(a.putAspect).
+		Doc("Creates or replaces a rule’s list of aspects.").
+		Param(ws.PathParameter("scope", "scope").DataType("string")).
+		Param(ws.PathParameter("subject", "subject").DataType("string")).
+		Param(ws.PathParameter("aspect", "aspect").DataType("string")).
```
ZackButcher: Should this capture `ruleid` as a param too? Also, it looks like you use underscores in other places, here and elsewhere should the `ruleid` param name be `rule_id` instead?

esnible:
jenkins rebuild mixer/manager-regression

pkg/config/apiserver.go
```diff
@@ -88,6 +104,58 @@ func (a *API) register(c *restful.Container) {
...
+		Reads(&pb.Aspect{}).
+		Writes(APIResponse{}))
+
+	ws.Route(ws.
```
ZackButcher: Can you push this route up so it's next to the two other `/scopes/{scope}/subjects/{subject}/rules/` routes please.

pkg/config/apiserver.go
esnible: I missed the ruleid parameter; good catch.  Regarding rule_id vs ruleid, I am following _istio/api/mixer/v1/config_api_swagger.yaml_:124-156 which uses ruleid.

ZackButcher: @mandarjog we should update the swagger spec so naming style is consistent across parameters.

Issue #620, opened 2017-04-25
mixc usage instructions confusing


esnible:
`mixc --help` reports the commands:

```
  check       Invokes the mixer's Check API.
  quota       Invokes the mixer's Quota API.
  report      Invokes the mixer's Report API.
  version     Prints out build version information
```

However "Invokes ... API" is confusing.  A sentence explaining what each command does is needed.

The commands are

```
mixc check [--bestEffort] [--attributes comma-separated-string]
mixc report [--attributes comma-separated-string]
mixc quota [--name string] [--amount int] [--bestEffort] [--attributes comma-separated-string]
```

My examples, which took a lot of trial-and-error to discover, are

```
# Examples
mixc check --attributes target.service=foo.default.svc.cluster.local
mixc report --attributes target.service=foo.default.svc.cluster.local
mixc quota --attributes target.service=foo.default.svc.cluster.local --name=RequestCount
```

Can we include these or better examples?

Those examples produce output, but I don't really understand what they do.  There should be a sentence or two explaining each command's purpose.

PR #643, opened 2017-05-01
Expand help messages for mixc.

geeknoid:
Addresses issue #620

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/643)


geeknoid:
​I hesitate to put an example, since it ends up being largely attributes
and the exact set of attributes depends on the current config...​

PR #648, opened 2017-05-02
Ensure MutableBag is not used after freeing

mandarjog:
1. Add refcounting via "Child()" calls
2. Ensure that mutableBag is never handed over to a different goroutines. Always use mutableBag.Child() to hand over to another goroutine. This ensures that
   a. A parent will never be freed if children have not been freed
3. invokeExecutorFunc's perviously closed over requestBag and responseBag. Now those parameters are explicit. This lets us create Child bags, just before async dispatch.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/648)

### DIFF
--- a/pkg/attribute/mutableBag.go
+++ b/pkg/attribute/mutableBag.go
@@ -37,6 +37,11 @@ type MutableBag struct {
 	parent Bag
 	values map[string]interface{}
 	id     int64 // strictly for use in diagnostic messages
+	// number of children of this bag. will not be reclaimed until children count == 0
+	children int
+	// done has been called on this bag
+	doneScheduled bool
+	sync.Mutex
 }
 
 var id int64
@@ -56,13 +61,18 @@ var mutableBags = sync.Pool{
 //
 // When you are done using the mutable bag, call the Done method to recycle it.
 func GetMutableBag(parent Bag) *MutableBag {
+	pp, _ := parent.(*MutableBag)
+	if pp != nil {
+		pp.Lock()
+		pp.children++
+		pp.Unlock()
+	}
 	mb := mutableBags.Get().(*MutableBag)
-
 	mb.parent = parent
 	if parent == nil {
 		mb.parent = empty
 	}
-
+	mb.doneScheduled = false
 	return mb
 }
 
@@ -96,10 +106,52 @@ func copyValue(v interface{}) interface{} {
 	return v
 }
 
+// Child allocates a derived mutable bag.
+//
+// Mutating a child doesn't affect the parent's state, all mutations are deltas.
+func (mb *MutableBag) Child() *MutableBag {
+	// if code tries to create a child after Done
+	// is scheduled or called
+	if mb.doneScheduled || mb.parent == nil {
+		panic(fmt.Errorf("bag used after scheduled destruction %#v", mb))
+	}
+	return GetMutableBag(mb)
+}
+
+// childDone indicates that a child bag has been reclaimed.
+func (mb *MutableBag) childDone() {
+	mb.Lock()
+	mb.children--
+	// if this bag was already scheduled for done()
+	// then perform the operation.
+	if mb.doneScheduled && mb.children == 0 {
+		mb.done()
+	}
+	mb.Unlock()
+}
+
 // Done indicates the bag can be reclaimed.
 func (mb *MutableBag) Done() {
+	mb.Lock()
+	mb.doneScheduled = true
+	if mb.children == 0 {
+		mb.done()
+	}
+	mb.Unlock()
+}
+
+// done reclaims the bag. always called under a lock to the bag
+func (mb *MutableBag) done() {
 	mb.Reset()
+	mbparent, _ := mb.parent.(*MutableBag)
+	if mbparent != nil {
+		mbparent.childDone()
+	}
 	mb.parent = nil
+	mb.doneScheduled = false
+	if glog.V(3) {
+		glog.Infof("freed bag: %#v", mb)
+	}
 	mutableBags.Put(mb)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/attribute/mutableBag.go
```diff
@@ -56,13 +61,18 @@ var mutableBags = sync.Pool{
 //
 // When you are done using the mutable bag, call the Done method to recycle it.
 func GetMutableBag(parent Bag) *MutableBag {
+	pp, _ := parent.(*MutableBag)
+	if pp != nil {
```
geeknoid: There's a race here. The parent could be 'collected' here before the child is attached.

pkg/attribute/mutableBag.go
```diff
@@ -96,10 +106,52 @@ func copyValue(v interface{}) interface{} {
 	return v
 }
 
+// Child allocates a derived mutable bag.
+//
+// Mutating a child doesn't affect the parent's state, all mutations are deltas.
```
geeknoid: This comment is no longer true.

mandarjog:
Comments addressed. PTAL.
Also    
Replaced bag.Done with attribute.BagDone()
    attribute.BagDone releases the MutableBag
    and sets the underlying pointer to nil

pkg/attribute/mutableBag.go
```diff
@@ -37,6 +37,11 @@ type MutableBag struct {
 	parent Bag
 	values map[string]interface{}
 	id     int64 // strictly for use in diagnostic messages
+	// number of children of this bag. will not be reclaimed until children count == 0
```
geeknoid: 'not be reclaimed' -> 'not returned to the pool'

pkg/attribute/mutableBag.go
```diff
@@ -96,10 +111,58 @@ func copyValue(v interface{}) interface{} {
...
+	return GetMutableBag(mb)
+}
+
+// childDone indicates that a child bag has been reclaimed.
+func (mb *MutableBag) childDone() {
+	mb.Lock()
+	mb.children--
+	// if this bag was already scheduled for done()
+	// then perform the operation.
+	if mb.doneScheduled && mb.children == 0 {
```
geeknoid: I don't like the complexity introduced by this 'doneScheduled' feature. 

All we're trying to do here is prevent misuse of the API and not to make extra affordances for these bad uses. Instead of ever setting doneScheduled to true, let's just panic at that point: "hey dummy, you're marking a parent bag done while there are still outstanding children". This shouldn't happen in a running system.

ZackButcher:
We can put a noCopy struct into the bag so that `go vet` catches copies of bags like it does for mutexes, btw: https://github.com/golang/go/issues/8005#issuecomment-190753527

pkg/api/grpcServer.go
```diff
@@ -177,8 +177,8 @@ func (s *grpcServer) dispatch(stream grpc.Stream, methodName string, getState st
 				glog.Errorf("Unable to send gRPC response message: %v", err)
 			}
 
-			requestBag.Done()
-			responseBag.Done()
+			attribute.BagDone(&requestBag)
```
mandarjog: The value it adds is if the requestBag is used after this point, it will be `nil` and produce a hard failure.

geeknoid:
Unless the variable is aliased. As an example, if you call done within a
function that received the bag as an input parameter, this trick doesn't
work.

The only way to do this pattern in a robust way would be to use a wrapper
to encapsulate the pointer to the bag. The value type would have a pointer
in it which gets set to nil when Done is called. But then you have to
allocate those wrappers and it defeats the value of the pooled bags.

Anyway, I maintain that this has limited usefulness and it complicates the
usage model.

mandarjog:
@geeknoid  will the following not work?
the line `b.name` produces `SIGSEGV`
You can circumvent it by using `badVar`, but at that point it is not inadvertent. 

```func do(b *MutableBag) {
	fmt.Printf("b = %#v", b)
        badVAr := b
	DoneBag(&b)
	fmt.Printf("b = %#v", b.name)
       
       // now use badVar
}
```

geeknoid:
What I mean is:

   Bag b = ...
   myFunc(b);
   b.Get()...

...

func myFunc(Bag *b) {
   attribute.BagDone(&b);
}

b.Get above will not trigger any error.

Instead of this trick, it would be more robust to just have a single
boolean in the bag that says whether the bag is in the pool or not, and
then you check that boolean within the implementation of every bag method.
"If bag in pool then panic". This involves no strange API pattern and is
more robust overall. It remains a bit racy due to TOCTOU issue, but it's a
pretty close approximation, sufficient for the sake of ensuring correct
behavior.

mandarjog:
ok. We can check `isInBag bool`  while holding on to the bag lock, but problem still remains.

                                                                                            contains params from an unrelated request
Thread 1 -->     bag.Done()        ..................................      use bag
Thread 2 -->     ........................ bag=getMutableBag() 

I think all this is fraught with races.
If Thread 1 calls DoneBag(&bag) it will not be able to use the bag later. 

I think it is easier to police "If you allocate it, you free it" during code reviews.

geeknoid:
True. To work around that, you'd need some sort of generation counter.
Getting too complicated again.

pkg/attribute/mutableBag.go
```diff
@@ -96,10 +111,58 @@ func copyValue(v interface{}) interface{} {
...
+	mb.Lock()
+	mb.doneScheduled = true
+	if mb.children == 0 {
+		mb.done()
+	}
+	mb.Unlock()
+}
+
+// done reclaims the bag. always called under a lock to the bag
+func (mb *MutableBag) done() {
```
douglas-reid: super nit: If we are renaming, would we consider something like:

`attributes.Release()` calls `bag.release()` calls `bag.returnToPool()` calls `parent.reclaimChild()`... Or maybe something like `Discard` instead of `Release`?

I feel like `Done` is what makes it awkward, though it does share similarity with waitgroup.

ZackButcher: Those names LGTM too.

pkg/attribute/mutableBag.go
```diff
@@ -96,10 +111,58 @@ func copyValue(v interface{}) interface{} {
...
 	mb.Reset()
```
douglas-reid: so Reset() doesn't clear the parent of the bag nor does it clear the `doneScheduled` field?  Is there ever a context where we want only the contents of the bag reset, but not the other metadata of the bag?

mandarjog:
PTAL.

PR #655, opened 2017-05-03
First pass at attempting to wire up monitored resources

ZackButcher:
This is how I'm planning to pass monitored resources through Mixer, today just to `Report` aspect kinds. This PR doesn't yet wire up things yet, but the basic structure for that plumbing exists. PTAL, if there's not opposition to this approach then I'll carry on wiring things up.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/655)

### DIFF
--- a/pkg/aspect/config/monitoredResource.proto
+++ b/pkg/aspect/config/monitoredResource.proto
@@ -1,0 +1,41 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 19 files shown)


pkg/aspect/config/monitoredResource.proto
```diff
@@ -0,0 +1,41 @@
+// Copyright 2017 Google Inc.
```
geeknoid: Google -> Istio Authors, here and elsewhere.

ZackButcher: Oh whoops, got cause with an old template.

douglas-reid:
With the pending changes to Mixer config and templates/types/etc., should we pursue this now or hold off until that dust has settled and add these in via new templates?

ZackButcher:
In favor of attempting to close out our backlog pre alpha, I'm going to close this PR. Based on comments here and out of band, I think we generally agree on this approach. I'll get back to this post alpha.

PR #656, opened 2017-05-03
Fix up copyright notices

ZackButcher:
```shell
$ find . -type f -name "*.go" -o -type f -name "*.proto" | xargs \
   sed -i \
     -e "s/Copyright 2017 Google Inc/Copyright 2017 the Istio Authors/g" \
     -e "s/Copyright 2017 The Istio Authors/Copyright 2017 the Istio Authors/g"
````

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/656)

### DIFF
--- a/adapter/kubernetes/cache.go
+++ b/adapter/kubernetes/cache.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Google Inc.
+// Copyright 2017 the Istio Authors.
 //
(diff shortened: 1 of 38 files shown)


ldemailly:
We should define "Istio Authors" somewhere ? (https://github.com/istio/istio/blob/master/OWNERS is empty and isn't exactly Authors either)

PR #669, opened 2017-05-08
'user-agent' to 'useragent'

ZackButcher:
https://github.com/istio/istio.github.io/pull/116 changed the name.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/669)

### DIFF
--- a/testdata/configroot/scopes/global/descriptors.yml
+++ b/testdata/configroot/scopes/global/descriptors.yml
@@ -57,7 +57,7 @@ manifests:
         valueType: INT64
       request.time:
         valueType: TIMESTAMP
-      request.user-agent:
+      request.useragent:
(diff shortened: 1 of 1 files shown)


douglas-reid:
so, we will have `request.useragent` and `request.headers["user-agent"]`, is that correct?

ZackButcher:
Yup, unfortunately. I'm of the opinion that we should name our attributes after the standard HTTP headers and have a well-known mapping from HTTP-header space to program-identifier space (e.g. in golang remove the `-`'s and replace with UpperCamelCase) if the full reason for the change is wanting to make sure attribute names are usable as program identifiers (which also has a large problem because we name everything with a `.`, and there are very very few languages that allow a `.` in an identifier name).

PR #694, opened 2017-05-11
Introduce redis backend for config store.

jmuk:
Note: this implements the exact SPI currently we have. With
some redis features, some things (like listening updates)
can be done more efficiently, but that needs to change the SPI.

I'll address that after landing this.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/694)

### DIFF
--- a/pkg/config/redisstore.go
+++ b/pkg/config/redisstore.go
@@ -16,10 +16,10 @@ package config
 
 import (
 	"fmt"
-	"log"
 	"net/url"
 	"strconv"
 
+	"github.com/golang/glog"
 	"github.com/mediocregopher/radix.v2/redis"
 )
 
@@ -139,7 +139,7 @@ func (rs *redisStore) List(key string, recurse bool) (keys []string, index int, 
 		for i, rk := range respKeys {
 			// TODO: check recurse flag for filitering keys.
 			if key, err2 := rk.Str(); err2 != nil {
-				log.Printf("illformed responses %d-th value for cursor %d isn't a string (%v)", i, cursor, rk)
+				glog.Warningf("illformed responses %d-th value for cursor %d isn't a string (%v)", i, cursor, rk)
 				continue
 			} else {
 				keys = append(keys, key)
--- a/pkg/config/redisstore.go
+++ b/pkg/config/redisstore.go
@@ -1,0 +1,142 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package config
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+	"strconv"
+
+	"github.com/mediocregopher/radix.v2/redis"
+)
+
+type redisStore struct {
+	client *redis.Client
+	url    *url.URL
+}
+
+func setupConnection(client *redis.Client, u *url.URL) error {
+	if u.User != nil {
+		if pass, ok := u.User.Password(); ok {
+			resp := client.Cmd("AUTH", pass)
+			if resp.Err != nil {
+				return fmt.Errorf("failed to authenticate with password %s: %v", pass, resp.Err)
+			}
+		}
+	}
+	if len(u.Path) > 1 {
+		dbNum, err := strconv.ParseInt(u.Path[1:], 10, 0)
+		if err != nil {
+			return fmt.Errorf("failed to parse dbNum \"%s\", it should be an integer", u.Path[1:])
+		}
+		if dbNum < 0 {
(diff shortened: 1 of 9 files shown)


pkg/config/redisstore.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if len(u.Path) > 1 {
+		dbNum, err := strconv.ParseInt(u.Path[1:], 10, 0)
+		if err != nil {
+			return fmt.Errorf("failed to parse dbNum \"%s\", it should be an integer", u.Path[1:])
+		}
+		if dbNum < 0 {
+			return fmt.Errorf("dbNum \"%s\" should be a positive integer", u.Path[1:])
+		}
+		client.Cmd("SELECT", dbNum)
```
mandarjog: issue a ping before "selecting" a database?

mandarjog: Move all local validation  (extracting dbNum) before reaching out to the server.

pkg/config/redisstore_test.go
```diff
@@ -0,0 +1,146 @@
...
+			t.Errorf("dbnum is not an integer: %v", err)
+		} else if num != tt.dbNum {
+			t.Errorf("expected to connect %d but connected to %d", tt.dbNum, num)
+		}
+		rs.Close()
+	}
+}
+
+func TestInvalidDBNum(t *testing.T) {
+	s, err := miniredis.Run()
```
mandarjog: You don't need to start miniredis to test invalid db numbers.

pkg/config/store_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/alicebob/miniredis"
+)
+
+func TestNewStore(t *testing.T) {
```
mandarjog: If this is redis_store test, update the filename.

mandarjog:
https://github.com/istio/mixer/issues/695

jmuk:
Review status: 0 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[pkg/config/redisstore.go, line 47 at r1](https://reviewable.io:443/reviews/istio/mixer/694#-KjtHV4lx6HBjK6AvIsx-r1-47:-KjtrulljzExYmnWs0KF:b-896fix) ([raw file](https://github.com/istio/mixer/blob/4ab613a10c0268dd66cbc4088493bb8876ef9369/pkg/config/redisstore.go#L47)):*

Move all local validation  (extracting dbNum) before reaching out to the server.
</blockquote></details>

Done.

---

*[pkg/config/redisstore.go, line 126 at r1](https://reviewable.io:443/reviews/istio/mixer/694#-KjtHV4lx6HBjK6AvIsx-r1-126:-Kjtrz2qtyWSPsM0VbzO:b-896fix) ([raw file](https://github.com/istio/mixer/blob/4ab613a10c0268dd66cbc4088493bb8876ef9369/pkg/config/redisstore.go#L126)):*

best to keep a track of the number of keys we saw last time.
That way you can do
` keys := make([]string, rs.length) `  and minimize reallocating the slice.
</blockquote></details>

Done.

---

*[pkg/config/redisstore_test.go, line 123 at r1](https://reviewable.io:443/reviews/istio/mixer/694#-KjtHV4lx6HBjK6AvIsy-r1-123:-Kjts-LZ9545x1iMVoH6:b-896fix) ([raw file](https://github.com/istio/mixer/blob/4ab613a10c0268dd66cbc4088493bb8876ef9369/pkg/config/redisstore_test.go#L123)):*

You don't need to start miniredis to test invalid db numbers.
</blockquote></details>

Done.

---

*[pkg/config/store_test.go, line 24 at r1](https://reviewable.io:443/reviews/istio/mixer/694#-KjtHV4miHAGzTUfO1g1-r1-24:-Kjts04b08JR-XbI5LeY:b-bc00et) ([raw file](https://github.com/istio/mixer/blob/4ab613a10c0268dd66cbc4088493bb8876ef9369/pkg/config/store_test.go#L24)):*

If this is redis_store test, update the filename.
</blockquote></details>

This was actually moved from fsstore_test.go, testing NewStore function in store.go, so I think this file name is better.
And miniredis was added because otherwise NewStore() will fail.

Removed miniredis code, and modified the expected error message is like connection failure. cf. https://github.com/istio/mixer/blob/master/pkg/config/fsstore_test.go#L116

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/mixer/694)*

jmuk:
mixs server ran with a local redis DB (i.e. the process starts) but I haven't done any further yet. will check.

pkg/config/redisstore.go
```diff
@@ -89,6 +106,7 @@ func (rs *redisStore) Set(key, value string) (index int, err error) {
 }
 
 func (rs *redisStore) List(key string, recurse bool) (keys []string, index int, err error) {
+	keys = make([]string, rs.listLength)
```
mandarjog: should be `make([]string, 0, rs.listLength)` ?
Otherwise keys will have first 5 elements as empty right?

mandarjog: There should be a test in the common store tests to check for this.

jmuk: My bad, fixed. Wrote a new pattern of calling List twice.

jmuk:
I also checked the behavior of mixs binary, and noticed that it won't work well without the default configurations. Possibly nicer to create a data importer of the config for the initial setup.

jmuk:
Hi,

Is there anything I'm missing?

PR #708, opened 2017-05-12
Add basic monitoring server to Mixer.

douglas-reid:
This PR establishes a server within Mixer for providing some basic
monitoring and version information for debugging. This solution is
meant ONLY to be a short term patch while we iterate on a more
thorough and complete design for health and status reporting for Mixer
as well as for self-monitoring of Istio components.

It provides two basic endpoints:
- `/metrics` which exposes data from the default collector for
prometheus. This should include stats on number of goroutines and cpu
usage, etc.
- `version` which provides a quick way to get a snapshot of the version
info for Mixer.

Example snippet from `/metrics`:

```
# HELP go_gc_duration_seconds A summary of the GC invocation durations.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 9.9604e-05
go_gc_duration_seconds{quantile="0.25"} 0.00012713
go_gc_duration_seconds{quantile="0.5"} 0.000288486
go_gc_duration_seconds{quantile="0.75"} 0.00030185
go_gc_duration_seconds{quantile="1"} 0.000349945
go_gc_duration_seconds_sum 0.001295973
go_gc_duration_seconds_count 6
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 2061
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 4.922944e+06
...
```

Example output from `/version`:

```
version: 0.1.1-2-gb35ce11 (build: 2017-05-12-b35ce11, status: Clean)
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/708)

### DIFF
--- a/cmd/server/cmd/server.go
+++ b/cmd/server/cmd/server.go
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"os"
 	"strings"
 	"time"
 
 	bt "github.com/opentracing/basictracer-go"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
@@ -60,6 +62,9 @@ type serverArgs struct {
 	configFetchIntervalSec        uint
 	configIdentityAttribute       string
 	configIdentityAttributeDomain string
+	monitoringPort                uint16
+	metricsPath                   string
+	versionPath                   string
 
 	// @deprecated
 	serviceConfigFile string
@@ -89,6 +94,9 @@ func serverCmd(printf, fatalf shared.FormatFn) *cobra.Command {
 	}
 	serverCmd.PersistentFlags().Uint16VarP(&sa.port, "port", "p", 9091, "TCP port to use for Mixer's gRPC API")
 	serverCmd.PersistentFlags().Uint16VarP(&sa.configAPIPort, "configAPIPort", "", 9094, "HTTP port to use for Mixer's Configuration API")
+	serverCmd.PersistentFlags().Uint16Var(&sa.monitoringPort, "monitoringPort", 1337, "HTTP port to use for the exposing mixer self-monitoring information")
+	serverCmd.PersistentFlags().StringVar(&sa.metricsPath, "metricsPath", "/metrics", "Request path for metrics data for mixer self-monitoring")
+	serverCmd.PersistentFlags().StringVar(&sa.versionPath, "versionPath", "/version", "Request path for version info for mixer self-monitoring")
 	serverCmd.PersistentFlags().UintVarP(&sa.maxMessageSize, "maxMessageSize", "", 1024*1024, "Maximum size of individual gRPC messages")
 	serverCmd.PersistentFlags().UintVarP(&sa.maxConcurrentStreams, "maxConcurrentStreams", "", 32, "Maximum supported number of concurrent gRPC streams")
 	serverCmd.PersistentFlags().IntVarP(&sa.apiWorkerPoolSize, "apiWorkerPoolSize", "", 1024, "Max # of goroutines in the API worker pool")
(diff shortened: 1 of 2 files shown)


cmd/server/cmd/server.go
```diff
@@ -89,6 +94,9 @@ func serverCmd(printf, fatalf shared.FormatFn) *cobra.Command {
 	}
 	serverCmd.PersistentFlags().Uint16VarP(&sa.port, "port", "p", 9091, "TCP port to use for Mixer's gRPC API")
 	serverCmd.PersistentFlags().Uint16VarP(&sa.configAPIPort, "configAPIPort", "", 9094, "HTTP port to use for Mixer's Configuration API")
+	serverCmd.PersistentFlags().Uint16Var(&sa.monitoringPort, "monitoringPort", 1337, "HTTP port to use for the exposing mixer self-monitoring information")
+	serverCmd.PersistentFlags().StringVar(&sa.metricsPath, "metricsPath", "/metrics", "Request path for metrics data for mixer self-monitoring")
+	serverCmd.PersistentFlags().StringVar(&sa.versionPath, "versionPath", "/version", "Request path for version info for mixer self-monitoring")
```
mandarjog: I don't think all these paths should be configurable.
We may want the context root to be configurable   so "/health"  or /status"
and then we can have 
/status/metrics  and /status/version underneath it.

we should just use the `statusz/healthz` endpoint paths that we will eventually support.

cmd/server/cmd/server.go
```diff
@@ -255,6 +263,31 @@ func runServer(sa *serverArgs, printf, fatalf shared.FormatFn) {
...
+	if monitoringListener, err = net.Listen("tcp", fmt.Sprintf(":%d", sa.monitoringPort)); err != nil {
+		fatalf("Unable to listen on socket: %v", err)
+	}
+
+	// NOTE: this is a temporary solution for provide bare-bones debug functionality
+	// for mixer. a full design / implementation of self-monitoring and reporting
+	// is coming. that design will include proper coverage of statusz/healthz type
+	// functionality, in addition to how mixer reports its own metrics.
+	srvMux := http.NewServeMux()
+	srvMux.Handle(sa.metricsPath, promhttp.Handler())
```
mandarjog: Does the default promhttp handler give us all the goodies we want?

cmd/server/cmd/server.go
```diff
@@ -255,6 +263,31 @@ func runServer(sa *serverArgs, printf, fatalf shared.FormatFn) {
...
+	srvMux.HandleFunc(sa.versionPath, func(out http.ResponseWriter, req *http.Request) {
+		if _, verErr := out.Write([]byte(version.Info.String())); verErr != nil {
+			printf("error printing version info: %v", verErr)
+		}
+	})
+	monitoring := &http.Server{Addr: fmt.Sprintf(":%d", sa.monitoringPort), Handler: srvMux}
+	printf("Starting self-monitoring on port %d", sa.monitoringPort)
+	go func() {
+		if monErr := monitoring.Serve(monitoringListener.(*net.TCPListener)); monErr != nil {
```
mandarjog: any reason why this cannot be `glog.Fatal (monitoring.Server ...)` or similar ?

cmd/server/cmd/server.go
```diff
@@ -89,6 +94,9 @@ func serverCmd(printf, fatalf shared.FormatFn) *cobra.Command {
 	}
 	serverCmd.PersistentFlags().Uint16VarP(&sa.port, "port", "p", 9091, "TCP port to use for Mixer's gRPC API")
 	serverCmd.PersistentFlags().Uint16VarP(&sa.configAPIPort, "configAPIPort", "", 9094, "HTTP port to use for Mixer's Configuration API")
+	serverCmd.PersistentFlags().Uint16Var(&sa.monitoringPort, "monitoringPort", 1337, "HTTP port to use for the exposing mixer self-monitoring information")
+	serverCmd.PersistentFlags().StringVar(&sa.metricsPath, "metricsPath", "/metrics", "Request path for metrics data for mixer self-monitoring")
```
mandarjog: can we use consecutive ports by default?

cmd/server/cmd/server.go
douglas-reid: It uses the default gatherer, which includes two basic collectors for free:

> NewProcessCollector returns a collector which exports the current state of process metrics including cpu, memory and file descriptor usage as well as the process start time for the given process id under the given namespace.

> NewGoCollector returns a collector which exports metrics about the current go process.

I think this is good enough to start (especially for a bridge solution to a more full-fledged solution).

cmd/server/cmd/server.go
douglas-reid: no. but, do we want to make a failure to start the monitoring service fatal?

cmd/server/cmd/server.go
douglas-reid: we can use whatever ports we think are reasonable. is your suggestion:
- grpc: 9091
- config: 9092
- monitoring: 9093

?

cmd/server/cmd/server.go
douglas-reid: I don't think we know what paths we will eventually support. These don't have to have a common root. I can hardcode them for now, if that is preferable.

cmd/server/cmd/server.go
costinm: Can we avoid adding a new port instead ? AFAIK grpc allows adding http handlers.

( I assume the config API will move/migrate to manager or run as separate service at some point)

cmd/server/cmd/server.go
```diff
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"os"
 	"strings"
 	"time"
 
 	bt "github.com/opentracing/basictracer-go"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
costinm: can we also add _ "expvar" ? It doesn't hurt, and mostly free. Just needs the default http handler to be exposed.

cmd/server/cmd/server.go
costinm: +1 on hardcoding (also /metrics).

cmd/server/cmd/server.go
costinm: Curious - any reason we're not using the default mux ?

cmd/server/cmd/server.go
mandarjog: hardcoded is good.

cmd/server/cmd/server.go
douglas-reid: grpc-go reusing the same port is a mess. perhaps most troubling: https://github.com/grpc/grpc-go/issues/586.

cmd/server/cmd/server.go
douglas-reid: I've added it for now, though I am skeptical it provides more detail than the newly-exposed /metrics endpoint.

cmd/server/cmd/server.go
douglas-reid: not particularly. updating to use default.

douglas-reid:
PTAL. I believe I have addressed most of the review comments.

Changes:
- use port 9093
- expose /debug/vars via expvar
- use default mux
- hard-coded paths for /metrics and /version

PR #720, opened 2017-05-16
minor fix in validate

jmuk:
I've noticed this make to create empty arrays and then appending
items. This will also result in 'unknown key' log if verbose
level is high enough.

Instead of creating cfgkey properly, this PR just removed it and
made classifyKeys iterate over the keys in the cfg map directly.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/720)


rkpagadala:
LGTM

mandarjog:
@sebastienvas which groups does @jmuk need to be in for presubmit and e2e-smoketest.

jmuk:
I haven't updated the branch (this PR was created before I joined istio team yesterday). I believe those jenkins tasks will run soon.

sebastienvas:
@mandarjog He should be in istio org. The presubmit got queued.

mandarjog:
@jmuk can you ensure that you run bin/fmt.sh  and bin/linters.sh before submitting the PR ?

```
From https://github.com/istio/mixer
 * [new branch]      master     -> parent
Using 95524852332373faf5df6b07f57da917f4e54ad3^ to compare files to.
pkg/config/validator.go:1::warning: file is not gofmted with -s (gofmt)
pkg/config/validator.go:363:11:warning: should omit 2nd value from range; this loop is equivalent to `for key := range ...` (golint)
```

PR #721, opened 2017-05-16
Sets ISTIO_VERSION from istio/istio.RELEASE

sebastienvas:
This will only work in Jenkins where we create the artifacts.
We used to read from istio.RELEASE, but that meant updating the file
after each release.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/721)


sebastienvas:
build mixer/e2e-smoketest

PR #723, opened 2017-05-16
Send concurrent requests, measure time.

costinm:
Small extension to the 'repeat' option.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/723)


cmd/client/cmd/check.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/spf13/cobra"
 
+	"time"
```
douglas-reid: this looks like a linter issue.  `bin/fmt.sh` should help.

costinm: PTAL

cmd/client/cmd/check.go
```diff
@@ -77,13 +90,18 @@ func check(rootArgs *rootArgs, printf, fatalf shared.FormatFn) {
...
+		if rootArgs.repeat < 10 {
+			printf("Check RPC returned %s", decodeStatus(response.Result))
+			dumpAttributes(printf, fatalf, response.AttributeUpdate)
+		}
+	}
 	if err = stream.CloseSend(); err != nil {
 		fatalf("Failed to close gRPC stream: %v", err)
 	}
 
+	println("Done in ", time.Since(t), " avg(ms)=", (time.Since(t).Nanoseconds()/1000)/int64(rootArgs.repeat))
```
douglas-reid: Nanos / 1000 = Micros, no?

I think `time.Since(t).Nanoseconds() / time.Millisecond.Nanoseconds()` might avoid the issue here.

cmd/client/cmd/quota.go
```diff
@@ -100,16 +109,21 @@ func quota(rootArgs *rootArgs, printf, fatalf shared.FormatFn, name string, amou
...
+				response.Amount,
+				response.Expiration)
+			dumpAttributes(printf, fatalf, response.AttributeUpdate)
+		}
 	}
 
 	if err = stream.CloseSend(); err != nil {
 		fatalf("Failed to close gRPC stream: %v", err)
 	}
+	println("Done in ", time.Since(t), " avg=", (time.Since(t).Nanoseconds()/1000)/int64(rootArgs.repeat))
```
douglas-reid: let's add unit to the output here (as above) and use `time.Microsecond` or `time.Millisecond` instead of 1000.

cmd/client/cmd/report.go
```diff
@@ -61,29 +62,39 @@ func report(rootArgs *rootArgs, printf, fatalf shared.FormatFn) {
...
+		if rootArgs.repeat < 10 {
+			printf("Report RPC returned %s", decodeStatus(response.Result))
+			dumpAttributes(printf, fatalf, response.AttributeUpdate)
+		}
 	}
 
 	if err = stream.CloseSend(); err != nil {
 		fatalf("Failed to close gRPC stream: %v", err)
 	}
+	println("Done in ", time.Since(t), " avg=", (time.Since(t).Nanoseconds()/1000)/int64(rootArgs.repeat))
```
douglas-reid: same comment as above.

cmd/client/cmd/quota.go
rkpagadala: same comment as above.

cmd/client/cmd/check.go
rkpagadala: another choice is 
dur := time.Since(t)
println("Done in ", dur, " avg=", dur /time.Duration(rootArgs.repeat))
go just prints time with units

cmd/client/cmd/report.go
rkpagadala: same comment as above.

cmd/client/cmd/check.go
```diff
@@ -17,10 +17,13 @@ package cmd
 import (
 	"context"
 	"io"
+	"time"
 
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/spf13/cobra"
 
+	"sync"
```
douglas-reid: this is likely to be a linter issue.

cmd/client/cmd/root.go
```diff
@@ -92,7 +99,11 @@ func GetRootCmd(args []string, printf, fatalf shared.FormatFn) *cobra.Command {
 	rootCmd.PersistentFlags().StringVarP(&rootArgs.mixerAddress, "mixer", "m", "localhost:9091",
 		"Address and port of a running Mixer instance")
 	rootCmd.PersistentFlags().IntVarP(&rootArgs.repeat, "repeat", "r", 1,
-		"Sends the specified number of requests in quick succession")
+		"Sends the specified number of requests in quick succession. Concurency flag limits the number of outstanding requests.")
+	rootCmd.PersistentFlags().IntVarP(&rootArgs.concurrency, "concurrency", "c", 1,
```
geeknoid: If the flag is defined here, then the behavior should also apply to the Report command. Right now, since it's only implemented in Check, this flag and the next should be applied to the Check command only.

cmd/client/cmd/check.go
```diff
@@ -47,24 +50,55 @@ func check(rootArgs *rootArgs, printf, fatalf shared.FormatFn) {
...
-			printf("Check RPC failed with: %s", decodeError(err))
-		}
+	// TODO: one span for each request - not sure what we can get from the per-stream span.
+
+	t := time.Now()
+
+	var wg sync.WaitGroup
+	wg.Add(rootArgs.clients)
+
+	for s := 0; s < rootArgs.clients; s++ {
```
geeknoid: This is sending 'repeat' x 'clients' number of requests which is not what is implied by the command-line args. It should only be sending 'repeat' requests overall, split between 'clients' senders.

Instead of using channels here, it'd be more efficient to use WaitGroups.

cmd/client/cmd/root.go
costinm: Right - I was planning to add the same load test to report in a separate CL, but maybe it's not needed.

cmd/client/cmd/check.go
costinm: I don't think they do the same - WaitGroups (AFAIK) wait until all goroutines are done.

The channel capacity ensures that only a set number of requests are active. 
However I don't think the new code is doing that anymore :-(, will rewrite this.

douglas-reid:
Marking as closed, after offline discussion with costin@.

PR #729, opened 2017-05-17
Add KUBECONFIG environment variable handling to kubernetes adapter.

douglas-reid:
This PR primarily adds handling to kubernetes adapter to take the
value for kubeconfigPath from mixer config as well as from environmental
config (via the KUBECONFIG env var).

NOTE: this configuration was previously available with via
KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT env vars. This PR
consolidates that into the Istio standard KUBECONFIG env var to
harmonize deployment across componenets.

Tested with out-of-the-box config via:

```bash
$ KUBECONFIG=~/.kube/config bazel-bin/cmd/server/mixs server --configStoreURL=fs://$(pwd)/testdata/configroot --logtostderr -v=5
```

and several client calls, as follows:

```bash
$ bazel-bin/cmd/client/mixc report -v=45 -a target.service=foo.default.svc.cluster.local --string_attributes source.uid=kubernetes://productpage-v1-2087264475-bl9f2.default -m localhost:9091 --trace
Report RPC returned OK
Attribute              Type               Value
source.labels          map[string]string  map[version:v1 app:productpage pod-template-hash:2087264475]
source.name            string             productpage-v1-2087264475-bl9f2
source.namespace       string             default
source.service         string             productpage.default.svc.cluster.local
source.serviceAccount  string             default
```

This PR also fixes the previously observed issue with sync.Once and initialization.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/729)

### DIFF
--- a/adapter/kubernetes/kubernetes.go
+++ b/adapter/kubernetes/kubernetes.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 	"sync"
 	"time"
@@ -40,11 +41,12 @@ import (
 type (
 	builder struct {
 		adapter.DefaultBuilder
+		sync.Mutex
 
 		stopChan             chan struct{}
 		pods                 cacheController
-		once                 sync.Once
 		newCacheControllerFn controllerFactoryFn
+		needsCacheInit       bool
 	}
 	kubegen struct {
 		log    adapter.Logger
@@ -120,7 +122,14 @@ func Register(r adapter.Registrar) {
 
 func newBuilder(cacheFactory controllerFactoryFn) *builder {
 	stopChan := make(chan struct{})
-	return &builder{adapter.NewDefaultBuilder(name, desc, conf), stopChan, nil, sync.Once{}, cacheFactory}
+	return &builder{
+		adapter.NewDefaultBuilder(name, desc, conf),
+		sync.Mutex{},
+		stopChan,
+		nil,
+		cacheFactory,
+		true,
+	}
 }
 
 func (b *builder) Close() error {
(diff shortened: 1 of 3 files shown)


adapter/kubernetes/kubernetes.go
```diff
@@ -184,23 +193,30 @@ func (*builder) ValidateConfig(c adapter.Config) (ce *adapter.ConfigErrors) {
...
+	b.Lock()
+	defer b.Unlock()
+	if b.needsCacheInit {
 		refresh := paramsProto.CacheRefreshDuration
-		controller, err := b.newCacheControllerFn(paramsProto.KubeconfigPath, refresh, env)
+		path, exists := os.LookupEnv("KUBECONFIG")
+		if !exists {
+			path = paramsProto.KubeconfigPath
+		}
+		controller, err := b.newCacheControllerFn(path, refresh, env)
```
mandarjog: Should add debug log, re: which kubconfig `path` was finally used.

adapter/kubernetes/kubernetes_test.go
```diff
@@ -145,6 +147,39 @@ func TestBuilder_BuildAttributesGenerator(t *testing.T) {
...
+		name    string
+		testFn  controllerFactoryFn
+		conf    adapter.Config
+		wantErr bool
+	}{
+		{"success", fakePodCache, conf, false},
+	}
+
+	wantPath := "/want/kubeconfig"
+	if err := os.Setenv("KUBECONFIG", wantPath); err != nil {
```
mandarjog: test for what take precedence  between env var and proto?

adapter/kubernetes/kubernetes.go
douglas-reid: this happens in the first line of `newCacheControllerFn`:

`env.Logger().Infof("getting kubeconfig from: %#v", kubeconfigPath)`

adapter/kubernetes/kubernetes_test.go
douglas-reid: done.

douglas-reid:
Thanks for the review. I've updated the test to match comments. PTAL.

PR #731, opened 2017-05-18
Fix copyright.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/731)


ZackButcher:
@sebastienvas any idea what's going on with Jenkins in this PR? It appears to have lost its mind, and is spamming our inboxes as a result.

sebastienvas:
And then people ask me why do you want to switch from Jenkins? I upgraded plugins yesterday night this is what happens. I cleared the trigger data

PR #734, opened 2017-05-18
Implement Zipkin Tracing

ZackButcher:
@douglas-reid can you think of a better way to handle the WORKSPACE file? It's just horrible.

Plus, all the linters fail. Somewhere along the way it tries to get istio.io/mixer/vendor/github.com/opentracing instead of using github.com/opentracing:
```
cmd/server/cmd/server.go:253: cannot assign "github.com/opentracing/opentracing-go".Tracer to tracer (type "istio.io/mixer/vendor/github.com/opentracing/opentracing-go".Tracer) in multiple assignment:
	"github.com/opentracing/opentracing-go".Tracer does not implement "istio.io/mixer/vendor/github.com/opentracing/opentracing-go".Tracer (wrong type for Extract method)
		have Extract(interface {}, interface {}) ("github.com/opentracing/opentracing-go".SpanContext, error)
		want Extract(interface {}, interface {}) ("istio.io/mixer/vendor/github.com/opentracing/opentracing-go".SpanContext, error)
```

Totally unclear to me why this is happening, but presumably there's some weirdness in the WORKSPACE build files I set up. But I can't spot it.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/734)


ZackButcher:
https://github.com/bazelbuild/bazel/issues/2732 so skipping files starting with `_` is a "feature". Meh.

WORKSPACE
```diff
@@ -454,3 +454,255 @@ git_repository(
     remote = "https://github.com/istio/test-infra.git",
 )
 
+##
+## Zipkin opentracing implementation. Apparently bazel refuses to synthesize BUILD files
```
douglas-reid: i think we can remove some of the emotional commentary here ;). Maybe something like "use of underscore in repo requires separate build issue".

It is probably also worth filing an issue with the zipkin repo to ask why they need an underscore and if they would consider not using one. You might even offer to provide them with the bazel files for their repo (and save us embedding this whole mess to begin with).

Also, do we need the "_thrift" stuff?  Can we just build around it?

WORKSPACE
```diff
@@ -454,3 +454,255 @@ git_repository(
...
+    srcs = glob([
+        "_thrift/gen-go/zipkincore/*.go",
+    ]),
+    visibility = ["//visibility:public"],
+    deps = ["@com_github_apache_thrift//lib/go/thrift:go_default_library"],
+)
+
+"""
+
+ZIPKIN_THRIFT_SCRIBE_BUILD= """
```
douglas-reid: is this needed for anything?

WORKSPACE
```diff
@@ -454,3 +454,255 @@ git_repository(
...
+    srcs = glob(
+        include = ["types/*.go"],
+        exclude = ["types/*_test.go"],
+    ),
+    visibility = ["//visibility:public"],
+)
+
+"""
+
+ZIPKIN_THRIFT_ZIPKINCORE_BUILD = """
```
douglas-reid: i've asked this several times, but can we just skip this?

cmd/server/cmd/server.go
```diff
@@ -111,7 +112,8 @@ func serverCmd(printf, fatalf shared.FormatFn) *cobra.Command {
 	serverCmd.PersistentFlags().StringVarP(&sa.clientCertFiles, "clientCertFiles", "", "", "A set of comma-separated client X509 cert files")
 
 	// TODO: implement an option to specify how traces are reported (hardcoded to report to stdout right now).
-	serverCmd.PersistentFlags().BoolVarP(&sa.enableTracing, "trace", "", false, "Whether to trace rpc executions")
+	serverCmd.PersistentFlags().StringVarP(&sa.traceOutput, "traceOutput", "t", "",
+		"If the literal string 'STDOUT', traces will be produced and written to stdout. Otherwise the address is assumed to be a URL and HTTP zipkin traces are sent to that address.")
```
douglas-reid: what about STDERR ?  Any reason not to support that?

pkg/tracing/basictracing/recorder_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tracing
+package basictracing
```
douglas-reid: so, this is package `pkg/tracing/basictracing` ?  is there a better name?  would even `pkg/tracing/basic` work?

and what is meant by `basic`?

ZackButcher:
Unfortunately the comment on the PR is still accurate: bazel is able to build everything just fine, but even after running our bazel_to_go script you cannot `go build ./cmd/server`; it fails with the error I posted above. It's unclear to me how its getting confused about vendoring. If nothing else it breaks all our linters.

ZackButcher:
Since the unary branch is now dead, I'll address the existing review comments then close this PR and open a new one against master.

ZackButcher:
Killing this in favor of https://github.com/istio/mixer/pull/903, since this was against the unary branch and very much out of date.

PR #736, opened 2017-05-19
Tagging minor version when creating a new tag

sebastienvas:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/736)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,9 @@ ISTIO_VERSION_URL = 'https://raw.githubusercontent.com/istio/istio/master/istio.
 
 def setVersions() {
   def version = sh(returnStdout: true, script: "curl ${ISTIO_VERSION_URL}").trim()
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -14,10 +14,16 @@ def bazel = new Bazel()
 // This should be updated for a release branch.
 ISTIO_VERSION_URL = 'https://raw.githubusercontent.com/istio/istio/master/istio.RELEASE'
 
+def setVersions() {
+  def version = sh(returnStdout: true, script: "curl ${ISTIO_VERSION_URL}").trim()
+  def v = version.tokenize('.')
+  env.ISTIO_VERSION = version
```
mandarjog: What happens if istio.RELEASE does not have the correct format for some reason?

I think we should check if `version` has the expected format and fail the build if it does not, rather than setting incorrect versions.

sebastienvas: Good point. Done

PR #737, opened 2017-05-19
Customize flow control window size.

costinm:
64k window is too small, even with low RTT latency. 

Pulling the latest grpc also adds framer reuse and other improvements.

Note the 'pprof' change also enables  http://localhost:9093/debug/pprof/goroutine?debug=1,
which may be used if gdb and delve fail (as seem to be the case in some environments),
for tracking deadlocks or other issues.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/737)


WORKSPACE
```diff
@@ -113,9 +120,17 @@ new_git_repository(
...
+new_go_repository(
+    name = "org_golang_google_genproto",
+    commit = "bb3573be0c484136831138976d444b8754777aff",
+    importpath = "google.golang.org/genproto",
+)
+
 new_go_repository(
     name = "org_golang_google_grpc",
-    commit = "cdee119ee21e61eef7093a41ba148fa83585e143",  # Mar 14, 2017 (v1.2.0)
+    #commit = "cdee119ee21e61eef7093a41ba148fa83585e143",  # Mar 14, 2017 (v1.2.0)
```
mandarjog: you can just remove this line.

costinm: Done.

mandarjog:
@costinm can you do "git pull upstream master --rebase"
That will ensure a linear change history.
There are many merge commits in this and the other PR.

WORKSPACE
```diff
@@ -113,9 +120,16 @@ new_git_repository(
     remote = "https://github.com/googleapis/googleapis.git",
 )
 
+# Dependency for org_golang_google_grpc
+new_go_repository(
```
douglas-reid: curious: are there any broader concerns with having this dep and the other, different genproto deps all in our WORKSPACE?

costinm: Manager seems to be doing the same - I believe if we don't define it explicitly an older version will be picked anyways.

IMHO having one big WORKSPACE with all the deps explicitly visible and defined - and not using bzl scripts - would
be much cleaner, and worth the effort in particular if we do it in istio repo and use it across components.

douglas-reid: Ok. Sounds reasonable. Thanks. I bet @ZackButcher is looking forward to merging his WORKSPACE changes in with this ;).

costinm:
Can I get a +1 ?

douglas-reid:
Marking as closed, after offline discussion with costinm@.

PR #744, opened 2017-05-19
Add notion of version to metrics

douglas-reid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/744)


douglas-reid:
yes. i will need to go update those files as well (once this is in).

PR #757, opened 2017-05-23
Mixer config API: PUT adapters and descriptors

mandarjog:
Add config API PUTS for
adapters and descriptors

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/757)

### DIFF
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -252,7 +252,7 @@ func (a *API) getScopes(req *restful.Request, resp *restful.Response) {
 	writeErrorResponse(http.StatusNotImplemented, "Listing scopes not implemented", resp)
 }
 
-// getConfig returns the rules document for the scope and the subject.
+// getRules returns the rules document for the scope and the subject.
 // "/scopes/{scope}/subjects/{subject}/rules"
 func (a *API) getRules(req *restful.Request, resp *restful.Response) {
 	a.getConfig(req, resp, &pb.ServiceConfig{})
@@ -340,7 +340,7 @@ func (a *API) putRules(req *restful.Request, resp *restful.Response) {
 	})
 }
 
-// putAdaptersOrDescriptors creates or replaces adapter configurations.
+// putAdaptersOrDescriptors creates or replaces specified configurations.
 // "/scopes/{scope}/adapters"
 // "/scopes/{scope}/descriptors"
 func (a *API) putAdaptersOrDescriptors(req *restful.Request, resp *restful.Response) {
@@ -355,7 +355,7 @@ func (a *API) putAdaptersOrDescriptors(req *restful.Request, resp *restful.Respo
 	})
 }
 
-// getAdaptersOrDescriptors gets an adapter configurations.
+// getAdaptersOrDescriptors gets specified configurations.
 // "/scopes/{scope}/adapters"
 // "/scopes/{scope}/descriptors"
 func (a *API) getAdaptersOrDescriptors(req *restful.Request, resp *restful.Response) {
@@ -363,11 +363,11 @@ func (a *API) getAdaptersOrDescriptors(req *restful.Request, resp *restful.Respo
 	a.getConfig(req, resp, &ret)
 }
 
-// deleteAdapters gets an adapter configurations.
+// deleteAdaptersOrDescriptors deletes specified configurations.
 // "/scopes/{scope}/adapters"
 // "/scopes/{scope}/descriptors
 func (a *API) deleteAdaptersOrDescriptors(req *restful.Request, resp *restful.Response) {
-	writeErrorResponse(http.StatusNotImplemented, "delete adapters not implemented", resp) // TODO
+	writeErrorResponse(http.StatusNotImplemented, "delete not implemented", resp) // TODO
 }
 
 // from api spec
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -278,12 +278,12 @@ func (a *API) getConfig(req *restful.Request, resp *restful.Response, m interfac
 	var found bool
 
 	if val, _, found = a.store.Get(path); !found {
-		writeErrorResponse(http.StatusNotFound, fmt.Sprintf("no rules for %s", path), resp)
+		writeErrorResponse(http.StatusNotFound, fmt.Sprintf("no config for %s", path), resp)
 		return
 	}
 
 	if err := yaml.Unmarshal([]byte(val), m); err != nil {
-		msg := fmt.Sprintf("unable to parse rules at '%s': %v", path, err)
+		msg := fmt.Sprintf("unable to parse config at '%s': %v", path, err)
 		glog.Warning(msg)
 		writeErrorResponse(http.StatusInternalServerError, msg, resp)
 		return
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -114,6 +114,52 @@ func (a *API) register(c *restful.Container) {
 		Reads(&pb.ServiceConfig{}).
 		Writes(APIResponse{}))
 
+	// Adapters
+	ws.Route(ws.
+		GET("/scopes/{scope}/adapters").
+		To(a.getAdaptersOrDescriptors).
+		Doc("Gets named adapter configurations.").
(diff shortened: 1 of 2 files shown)


pkg/config/apiserver.go
```diff
@@ -114,6 +114,52 @@ func (a *API) register(c *restful.Container) {
...
+
+	ws.Route(ws.
+		PUT("/scopes/{scope}/adapters").
+		To(a.putAdaptersOrDescriptors).
+		Doc("Creates or replaces named adapter configurations.").
+		Param(ws.PathParameter("scope", "scope").DataType("string")).
+		Reads(&pb.GlobalConfig{}).
+		Writes(APIResponse{}))
+
+	ws.Route(ws.
```
douglas-reid: delete ALL adapters seems a bit... unsafe. I'm not sure we should even advertise such a method.

pkg/config/apiserver.go
```diff
@@ -114,6 +114,52 @@ func (a *API) register(c *restful.Container) {
...
+		Param(ws.PathParameter("scope", "scope").DataType("string")).
+		Writes(APIResponse{}))
+	ws.Route(ws.
+		PUT("/scopes/{scope}/descriptors").
+		To(a.putAdaptersOrDescriptors).
+		Doc("Creates or replaces descriptors.").
+		Param(ws.PathParameter("scope", "scope").DataType("string")).
+		Reads(&pb.GlobalConfig{}).
+		Writes(APIResponse{}))
+	ws.Route(ws.
```
douglas-reid: see above.

pkg/config/apiserver.go
```diff
@@ -278,14 +322,55 @@ func (a *API) putRules(req *restful.Request, resp *restful.Response) {
...
+
+// getAdaptersOrDescriptors gets an adapter configurations.
+// "/scopes/{scope}/adapters"
+// "/scopes/{scope}/descriptors"
+func (a *API) getAdaptersOrDescriptors(req *restful.Request, resp *restful.Response) {
+	var ret map[string]interface{}
+	a.getConfig(req, resp, &ret)
+}
+
+// deleteAdapters gets an adapter configurations.
```
douglas-reid: nit: does deleteAdapters "get" anything?

pkg/config/apiserver.go
```diff
@@ -278,14 +322,55 @@ func (a *API) putRules(req *restful.Request, resp *restful.Response) {
...
+// "/scopes/{scope}/adapters"
+// "/scopes/{scope}/descriptors
+func (a *API) deleteAdaptersOrDescriptors(req *restful.Request, resp *restful.Response) {
+	writeErrorResponse(http.StatusNotImplemented, "delete adapters not implemented", resp) // TODO
```
douglas-reid: nit: "delete not implemented for adapters or descriptors"

pkg/config/apiserver.go
```diff
@@ -278,14 +322,55 @@ func (a *API) putRules(req *restful.Request, resp *restful.Response) {
...
+		writeErrorResponse(http.StatusBadRequest, "put only supports global scope", resp)
+		return
+	}
+
+	a.storeConfigIfValid(req, resp, func(vd *Validated) interface{} {
+		return vd.adapter[scope]
+	})
+}
+
+// getAdaptersOrDescriptors gets an adapter configurations.
```
douglas-reid: nit: gets "adapter OR descriptor configurations"

pkg/config/apiserver.go
```diff
@@ -278,14 +322,55 @@ func (a *API) putRules(req *restful.Request, resp *restful.Response) {
...
+// "/scopes/{scope}/subjects/{subject}/rules"
+func (a *API) putRules(req *restful.Request, resp *restful.Response) {
+	a.storeConfigIfValid(req, resp, func(vd *Validated) interface{} {
+		return vd.rule[rulesKey{
+			Scope:   req.PathParameter("scope"),
+			Subject: req.PathParameter("subject")}]
+	})
+}
+
+// putAdaptersOrDescriptors creates or replaces adapter configurations.
```
douglas-reid: nit: maybe just drop "adapter" from all of these comments and just say "configurations."

pkg/config/apiserver.go
mandarjog: I can get rid of it. The actual method returns "not implemented".

pkg/config/apiserver.go
```diff
@@ -114,6 +114,52 @@ func (a *API) register(c *restful.Container) {
...
+
+	ws.Route(ws.
+		DELETE("/scopes/{scope}/adapters").
+		To(a.deleteAdaptersOrDescriptors).
+		Doc("Deletes named adapter configurations.").
+		Param(ws.PathParameter("scope", "scope").DataType("string")).
+		Writes(APIResponse{}))
+
+	// Descriptors
+	ws.Route(ws.
```
mandarjog: There are `PATCH` methods to come later (they are in the spec), that let you update individual descriptors and adapters.

pkg/config/apiserver.go
```diff
@@ -206,12 +252,10 @@ func (a *API) getScopes(req *restful.Request, resp *restful.Response) {
 	writeErrorResponse(http.StatusNotImplemented, "Listing scopes not implemented", resp)
 }
 
-// getRules returns the rules document for the scope and the subject.
+// getConfig returns the rules document for the scope and the subject.
```
mandarjog: These are API methods. They `return` from the http caller perspective by writing to the `resp` struct.

pkg/config/apiserver.go
```diff
@@ -114,6 +114,52 @@ func (a *API) register(c *restful.Container) {
 		Reads(&pb.ServiceConfig{}).
 		Writes(APIResponse{}))
 
+	// Adapters
+	ws.Route(ws.
```
mandarjog: yes, we will add query params at that time, but that goes hand in hand with `PATCH` operations.

This method is used when the operator is editing entire files, (like the config is checked into source control), and you want to push it into Mixer.

mandarjog:
Fixed comments and an error message. PTAL.

pkg/config/apiserver.go
```diff
@@ -228,27 +272,30 @@ func (a *API) deleteRules(req *restful.Request, resp *restful.Response) {
...
-func getRules(store KeyValueStore, path string) (statusCode int, msg string, data *pb.ServiceConfig) {
+func (a *API) getConfig(req *restful.Request, resp *restful.Response, m interface{}) {
+	path := req.Request.URL.Path[len(a.rootPath):]
 	var val string
 	var found bool
 
-	if val, _, found = store.Get(path); !found {
-		return http.StatusNotFound, fmt.Sprintf("no rules for %s", path), nil
+	if val, _, found = a.store.Get(path); !found {
+		writeErrorResponse(http.StatusNotFound, fmt.Sprintf("no rules for %s", path), resp)
```
ayj: This method is also used for descriptors/adapters but potentially returns an error about "no rules". This might not be a huge deal since the embedded mixer apiserver is likely to go away with the unified/enhanced config server?

mandarjog:
@ayj  fixed it.

pkg/config/apiserver.go
douglas-reid: OK. Sounds good.

pkg/config/apiserver.go
douglas-reid: What would we do other than return "not implemented"?

It seems like in all but the rarest case deleting all adapters _should_ fail validation if mixer is configured in any way to use any adapter. If that is the case, that just makes me think we don't need this lever anyway.

`PUT` of an empty file would accomplish the same thing if this behavior was needed, correct?  That has its own issues, but...

mandarjog: In the global scope it is unwise to do.

However, if an operation fails validation it will not be allowed to proceed anyway.

Delete may be allowed in the future, so this returns a "Not implemented" vs a 404 which is confusing.

douglas-reid: ok. i'm still skeptical that delete all will ever be needed (so we could remove it from the spec and not have a 404 or unimplemented), but this seems mostly harmless for now.

PR #759, opened 2017-05-23
introduce importer command for setting up the initial config data.

jmuk:
See https://github.com/jmuk/istio/commit/fe0073b9b34c452ea25550d2d26659d0f99a223a for an example of using this tool.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/759)

### DIFF
--- a/cmd/importer/main.go
+++ b/cmd/importer/main.go
@@ -1,0 +1,88 @@
+// Copyright 2016 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


cmd/importer/main.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2016 Istio Authors
```
mandarjog: 2017

douglas-reid:
Can we expand on the documentation here (and in the code)? I'd like some write-up of the use case and explanation of the why to persist -- perhaps even in an issue against this repo.

ZackButcher:
I'd prefer to see this written using `Cobra` like the other tools in the `cmd` dir.

PR #770, opened 2017-05-30
Update mixer's dev doc

xiaolanz:
Update mixer's dev doc on setting up personal token to be in-sync wit…

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/770)


doc/dev/development.md
```diff
@@ -77,6 +79,13 @@ To run some of Mixer's examples and tests, you need to set up Docker server.
 Please follow [these instructions](https://docs.docker.com/engine/installation/)
 for how to do this for your platform.
 
+### Setting up personal access token
+
+Since Istio requires two-factor authentication, you must setup personal access
```
ldemailly: you must setup **a** personal...

doc/dev/development.md
```diff
@@ -77,6 +79,13 @@ To run some of Mixer's examples and tests, you need to set up Docker server.
...
+token to enable push via HTTPS. Please follow [these
+instructions](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/)
+for how to create a token.
+
```
ldemailly: Maybe add a note this is only necessary for pushes by core dev, you don't need this (afaik) for a PR from your own fork ?

xiaolanz: My understanding is that this instruction is for pushing into the istio/mixer instead of the fork.

doc/dev/development.md
```diff
@@ -143,11 +165,19 @@ git remote set-url --push upstream no_push
...
 ### Committing changes to your fork
 
-When you're happy with some changes, you can commit them and push them to your fork:
+When you're happy with some changes, you can commit them to your repo:
 
 ```shell
 git add .
 git commit
+```
+Then push the change to the fork. When prompted for authentication, using your
```
ldemailly: it says fork here... that's why I was wondering - also again I think people can contribute without two factor (I think)

xiaolanz: You cannot join Istio without 2-factor being setup. With 2-factor, you have to push using token if you use HTTPS, which is what this flow is about.

ldemailly: see https://github.com/istio/mixer/pull/772

PR #774, opened 2017-05-31
Moving common dev setup to istio/istio/devel

ldemailly:
See https://github.com/istio/istio/pull/352

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/774)


doc/dev/development.md
```diff
@@ -1,336 +1,12 @@
 # Development Guide
```
geeknoid: Let's remove this file completely and just point the README.md (along with any other references) directly to the istio/istio location.

ldemailly: it references local files like 
 - [Writing Mixer adapters](./adapters.md)
(hard to spot in the sea of red but it's there)
so I don't think we should do that

ldemailly:
please re-review/accept - tia

PR #777, opened 2017-05-31
add the change notifications support to redis.

jmuk:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/777)


pkg/config/redisstore.go
```diff
@@ -64,29 +128,56 @@ func newRedisStore(u *url.URL) (rs *redisStore, err error) {
...
 func (rs *redisStore) String() string {
 	return fmt.Sprintf("redisStore: %v", rs.url)
 }
 
+func (rs *redisStore) index(increase int) int {
+	if rs.redisSubscriber == nil {
+		return indexNotSupported
+	}
+	rs.redisSubscriber.mu.Lock()
+	idx := rs.redisSubscriber.lastIndex + increase
```
mandarjog: Where is the persistent index updated?
This looks like `rs.redisSubscriber.lastIndex` is never modified.

pkg/config/redisstore.go
```diff
@@ -18,12 +18,36 @@ import (
...
+	"github.com/mediocregopher/radix.v2/pubsub"
 	"github.com/mediocregopher/radix.v2/redis"
 )
 
+const (
+	// The name of the config for redis to store the 'keyspace events' availability.
+	keyspaceEventsConfigKey = "notify-keyspace-events"
+
+	// The timeout for receiving the next keyspace event (i.e. the next change on the DB).
+	redisSubscriberTimeout = time.Millisecond * 200
```
ZackButcher: nit: I find this more readable flipping it around to `200 * time.Millisecond`

pkg/config/redisstore.go
```diff
@@ -163,4 +256,89 @@ func (rs *redisStore) Close() {
...
+func (sub *redisSubscriber) Close() {
+	sub.mu.Lock()
+	defer sub.mu.Unlock()
+	if sub.client == nil || !sub.running {
+		return
+	}
+	sub.running = false
+}
+
+func (sub *redisSubscriber) RegisterStoreChangeListener(s StoreListener) {
```
ZackButcher: RegisterStoreChangeListener -> RegisterListener

No need to mention StoreListener, this func is already unambiguous.

pkg/config/store.go
```diff
@@ -78,6 +78,10 @@ type ChangeNotifier interface {
 	// The client should issue ReadChangeLog to see what has changed if the call is available.
 	// else it should re-read the store, perform diff and apply changes.
 	RegisterStoreChangeListener(s StoreListener)
+
+	// IsStoreChangeAvailable returns true only when the backend store is configured
+	// to emit change notifications.
+	IsStoreChangeAvailable() bool
```
ZackButcher: I'm not convinced we need this. Why does it matter if my backend emits change notifications? This is a `ChangeNotifier`, the thing implementing this interface is responsible for producing change notifications, the fact that the backend this is abstracting over may or may not natively support change notifications seems like a detail that doesn't need to be exposed - my implementation of this interface is responsible for synthesizing notifications as appropriate.

Plus, as far as I can tell it's only used in a test.

pkg/config/redisstore.go
```diff
@@ -18,12 +18,36 @@ import (
...
+	"strings"
+	"sync"
+	"time"
 
 	"github.com/golang/glog"
+	"github.com/mediocregopher/radix.v2/pubsub"
 	"github.com/mediocregopher/radix.v2/redis"
 )
 
+const (
```
ZackButcher: I think this whole file should move into `pkg/config/redis`, `package redis`. Then the names of essentially all the functions here can be simplified:

`config.doesRedisConfigSupportsChangeNotifications` -> `redis.supportsChangeNotifications`
`config.setupRedisConnection` -> `redis.NewConnection`

pkg/config/redisstore.go
```diff
@@ -163,4 +256,89 @@ func (rs *redisStore) Close() {
...
+	if rs.redisSubscriber != nil {
+		rs.redisSubscriber.Close()
+	}
+}
+
+func (rs *redisStore) IsStoreChangeAvailable() bool {
+	return rs.redisSubscriber != nil
+}
+
+func (sub *redisSubscriber) startListening(dbNum uint64) error {
```
ZackButcher: `startListening` -> `listen`

pkg/config/redisstore.go
jmuk: Done, but it results in a relatively big refactoring. Please take a look.

pkg/config/redisstore.go
jmuk: done

pkg/config/redisstore.go
jmuk: done

pkg/config/redisstore.go
jmuk: done (it's actually a method name of `ChangeNotifier` -- but anyways. It's renamed).

pkg/config/store.go
ZackButcher: I think I'm misunderstanding how this is intended to be used.

In my mind:
- `Redis` implements the `ChangeNotifier` interface
- the `manager` registers its own `StoreListener`
- When there's a change to the underlying store, the `Redis` notifies `manager` the registered `StoreListener`
- the `manager`'s `StoreListener` impl calls `Redis`'s `ReadChangeLog` func, retrieving the new configs from the store

is that correct, and if so, how is `IsStoreChangeAvailable` intended to be used?

pkg/config/redisstore.go
ZackButcher: Yeah, it wound up being larger than I expected. I think the resulting code is cleaner though, thanks!

pkg/config/store.go
ZackButcher: Basically, because:
```go
var _ store.ChangeNotifier = &redisStore{}
var _ store.ChangeLogReader = &redisStore{}
```
in server:
```go
redis := redis.newStore(...)
listener := cfg.NewListener(redis) // redis as store.ChangeLogReader, returning a store listener
redis.RegisterListener(listener) // redis as store.ChangeNotifier
```

jmuk: What I'm thinking is same as yours. But the point is, how `manager` reacts upon `NotifyStoreChange`. I uploaded two WIP PRs to clarify the approaches:
1. https://github.com/jmuk/mixer/commit/f9c680836ef4d30a7be7e74344c1b63383fc9aa2
   - `NotifyStoreChange` directly invokes `fetchAndNotify`. otherwise `go c.loop()` will do the polling.
2. https://github.com/jmuk/mixer/commit/f574f180dbf818033ccf66e0c4b01f72b6b0ec19
   - `NotifyStoreChange` just remembers the index, the polling loop invokes `fetchAndNotify` (but if it knows no changes, it can skip `fetchAndNotify` invocation)

I was thinking the former approach originally, and this needs `IsStoreChangeAvailable` method so that manager can fall back to `loop` method when the change events are not enabled on the current storage.

ZackButcher: Ok, I thought that was your intended usecase (deciding programmatically whether to register a listener or do the polling yourself).

> so that manager can fall back to loop method when the change events are not enabled on the current storage.

That's exactly what I want us to avoid: it should not be the responsibility of the manager to choose how it interacts with the config store. The `store` package provides interfaces abstracting over being notified when changes occur and reading those changes when they do occur. It is the responsibility the thing implementing the `store` interfaces to work out the nitty gritty of _how_ it produces change notifications. If the manager has to decide then we haven't gotten the abstractions right. Instead, we should have an active polling implementation of `ChangeNotifier` as well as an implementation using notification mechanism native to the backend store. Then the operator selects the notification method when constructing a Mixer binary.

mandarjog: You just need `ChangeLogReader` to read and apply changes from the backing store.

Change notification is an optimization. So I suggest we simply make implementing `ChangeNotifier` interface optional. That way, if the a store implements it, we register to receive notifications.

We should be reading change log for new changes periodically anyway. If we get notified them we read it, but if the notification mechanism breaks down, we should still read it.

So we don't need the `IsStoreChangeAvailable()` method on that interface.

jmuk: Alright, removed. I should have sent the PR for `manager` of use of change notifier first... thanks for the comments.

PR #780, opened 2017-06-01
Switch Mixer to a unary gRPC API.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/780)


pkg/adapterManager/manager.go
```diff
@@ -390,13 +390,15 @@ func newCacheKey(kind config.Kind, cfg *cpb.Combined) (*cacheKey, error) {
 
 	if cfg.Builder.Params != nil {
 		if err := enc.Encode(cfg.Builder.Params); err != nil {
+			pool.PutBuffer(b)
 			return nil, err
 		}
 		ret.builderParamsSHA = sha1.Sum(b.Bytes())
 	}
 	b.Reset()
```
ZackButcher: We don't need to remove in this PR, but this `Reset` isn't needed - `pool.PutBuffer` calls `Reset` on the buffer and is hit unconditionally after this is called.

pkg/attribute/mutableBag.go
```diff
@@ -206,198 +203,209 @@ func (mb *MutableBag) Merge(bags ...*MutableBag) error {
...
-		}
+	if index >= int32(len(globalDict)) {
+		return "", me.Append(err, fmt.Errorf("attribute index %d is not defined in the available dictionaries", index))
 	}
 
-	return nil
+	return globalDict[index], err
+}
+
+func log(buf *bytes.Buffer, format string, args ...interface{}) {
```
ZackButcher: We'll need a `// nolint: interfacer` here, right? Also, occurs to me now that it'd be cleaner to declare this as an anonymous function inside of `GetBagFromProto` since it's only used there.

WORKSPACE
```diff
@@ -182,7 +182,7 @@ new_git_or_local_repository(
     path = "../api",
     remote = "https://github.com/istio/api.git",
     # Change this to True to use ../api directory
-    use_local = False,
+    use_local = True,
```
geeknoid: Done, you caught my PR right before I had pushed this.

cmd/client/cmd/report.go
```diff
@@ -53,36 +53,13 @@ func report(rootArgs *rootArgs, printf, fatalf shared.FormatFn) {
...
 	defer deleteAPIClient(cs)
 
-	span, ctx := cs.tracer.StartRootSpan(context.Background(), "mixc Report", ext.SpanKindRPCClient)
-	_, ctx = cs.tracer.PropagateSpan(ctx, span)
-
-	var stream mixerpb.Mixer_ReportClient
-	if stream, err = cs.client.Report(ctx); err != nil {
-		fatalf("Report RPC failed: %v", err)
-	}
+	span, ctx := ot.StartSpanFromContext(context.Background(), "mixc Check", ext.SpanKindRPCClient)
```
geeknoid: Done

pkg/adapterManager/manager.go
geeknoid: I think the reset is needed because the bytes.Buffer is used by the encoder, which is still used below this point.

pkg/attribute/mutableBag.go
geeknoid: No, I mangled the implementation enough to not require the nolint stuff.

ZackButcher: :+1:

pkg/api/grpcServer.go
```diff
@@ -14,314 +14,204 @@
...
-	"io"
-	"sync"
+	"fmt"
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/golang/protobuf/proto"
 	rpc "github.com/googleapis/googleapis/google/rpc"
-	"github.com/opentracing/opentracing-go/log"
+	context2 "golang.org/x/net/context"
```
douglas-reid: this is super ugly. Can we add a TODO for resolving this duplicative context mess?

pkg/api/grpcServer.go
```diff
@@ -14,314 +14,204 @@
...
 
-			responseBag := attribute.GetMutableBag(nil)
+	out := s.aspectDispatcher.Preprocess(ctx, requestBag, responseBag)
+	if status.IsOK(out) {
 
+		if err := requestBag.Merge(responseBag); err != nil {
+			// TODO: better error messages that push internal details into debuginfo messages
+			glog.Errorf("Could not merge mutable bags for request: %v", err)
+			out = status.WithInternal("The results from the request preprocessing could not be merged.")
+		} else {
```
douglas-reid: do we need this else ?  can we just return from within the `if` clause?  it'd be nice to eliminate some of the nesting here.

could we maybe defer the bits at the end and flatten by changing `if status.IsOk(out)` to `if ! status.IsOk(out) { return }`, etc. ?

cmd/server/cmd/server.go
```diff
@@ -100,14 +102,14 @@ func serverCmd(printf, fatalf shared.FormatFn) *cobra.Command {
 	serverCmd.PersistentFlags().Uint16Var(&sa.monitoringPort, "monitoringPort", 9093, "HTTP port to use for the exposing mixer self-monitoring information")
 	serverCmd.PersistentFlags().Uint16VarP(&sa.configAPIPort, "configAPIPort", "", 9094, "HTTP port to use for Mixer's Configuration API")
 	serverCmd.PersistentFlags().UintVarP(&sa.maxMessageSize, "maxMessageSize", "", 1024*1024, "Maximum size of individual gRPC messages")
-	serverCmd.PersistentFlags().UintVarP(&sa.maxConcurrentStreams, "maxConcurrentStreams", "", 32, "Maximum supported number of concurrent gRPC streams")
-	serverCmd.PersistentFlags().IntVarP(&sa.apiWorkerPoolSize, "apiWorkerPoolSize", "", 1024, "Max # of goroutines in the API worker pool")
-	serverCmd.PersistentFlags().IntVarP(&sa.adapterWorkerPoolSize, "adapterWorkerPoolSize", "", 1024, "Max # of goroutines in the adapter worker pool")
+	serverCmd.PersistentFlags().UintVarP(&sa.maxConcurrentStreams, "maxConcurrentStreams", "", 1024, "Maximum number of outstanding RPCs per connection")
```
douglas-reid: do we want to add some comment or explanation for this choice (and others) somewhere?

pkg/api/grpcServer.go
```diff
@@ -14,314 +14,204 @@
...
+		outAttrs: &resp.Attributes,
+		result:   &resp.Status,
 	}
 
-	*args.result = s.aspectDispatcher.Check(ctx, args.requestBag, args.responseBag)
-	// TODO: this value needs to initially come from config, and be modulated by the kind of attribute
-	//       that was used in the check and the in-used aspects (for example, maybe an auth check has a
-	//       30s TTL but a whitelist check has got a 120s TTL)
-	resp.Expiration = 5 * time.Second
+	err := s.dispatch(ctx, &dState, func(ctx context.Context, requestBag *attribute.MutableBag, responseBag *attribute.MutableBag) rpc.Status {
```
douglas-reid: here we have two contexts of different types both named `ctx`. can we update one or the other to be clearer about what is used later?

pkg/api/grpcServer.go
geeknoid: I don't know what to do to fix things. gRPC would need to make a breaking change to update to the new context definition. Otherwise, we have to pinch our nose and accept this annoyance.

ZackButcher: Std library `context` and `x/net/context` can be used interchangeably except for method signatures, I thought.

pkg/api/grpcServer.go
geeknoid: What's the problem with the current structure? It's pretty clear and simple, no? 'else' isn't an evil word is it?

pkg/api/grpcServer.go
geeknoid: I renamed legacy context variables to legacyCtx to help with this.

cmd/server/cmd/server.go
geeknoid: Added a TODO comment in this file to figure out rational numbers for all these values. Right now, it's really just picked out of thin air.

pkg/api/grpcServer.go
geeknoid: Yeah, so the method stubs generated for gRPC use the x/net/context stuff, while our code tries to use the std library version. We only see the old type in the signature of the Check/Report/Quota methods, and we immediately switch to the std lib version when these 3 methods call dispatch.

pkg/api/grpcServer.go
douglas-reid: there is nothing wrong per-se. it obviously works. this is more about readability.

See: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

Dealing with the error cases, etc., up front and indenting only for them makes it easier (imho) to quickly grasp the happy path. This is why I was suggesting check for !OK first to start with.

I don't want this to become a blocker for this PR, but I think it would improve the code. Up to your discretion.

pkg/api/grpcServer.go
douglas-reid: Ugh. Does gRPC 1.3 switch, or are we waiting for 2.0 ?

ZackButcher: I think it's 2.0 (AFAIK it's a breaking change?) - I certainly have had no problems compiling mixer as-is with grpc 1.3 (I finally have it working in my zipkin tracing branch).

PR #783, opened 2017-06-01
use glob pattern instead of listing files for BUILD

### LINKED ISSUE #781: Swap BUILD files to use glob rather than explicitly listing all files in a package
I forget to include new files in a package's BUILD file all the time - especially test files (which I run in my IDE while writing, no bazel involved). On multiple occasions I've found test files not listed in a BUILD rule in code that's already been checked in.

We should start using globs in the `src` list of our BUILD rules to avoid this; they should look like:

```
go_library(
    name = "go_default_library",
    srcs = glob([
        include = ["*.go"],
        exclude = ["*_test.go"],
    ]),
    visibility = ["<snip>"],
    deps = [
        # deps here
    ],
)

go_test(
    name = "<package name>_test",
    size = "small",
    srcs = glob(["*_test.go"]),
    library = ":go_default_library",
    visibility = ["//visibility:public"],
    deps = [
        # deps here
    ],
)

```

rkpagadala:
fixes #781

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/783)

### DIFF
--- a/adapter/BUILD
+++ b/adapter/BUILD
@@ -4,9 +4,10 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
 go_library(
     name = "go_default_library",
-    srcs = [
-        "inventory.go",
-    ],
+    srcs = glob(
+        include = ["*.go"],
+        exclude = ["*_test.go"],
+    ),
     deps = [
         "//adapter/denyChecker:go_default_library",
         "//adapter/genericListChecker:go_default_library",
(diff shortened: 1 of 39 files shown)


mandarjog:
I think it is better to be explicit and consistent.

ZackButcher:
If we're going to go for explicit lists of files then we need a pre-commit hook that verifies all new files are mentioned in a BUILD file.

douglas-reid:
@ZackButcher agreed. Let's get a pre-commit hook for gazelle working. It'd be awesome to couple that with codecov checks too (which currently are enforced for 3?? packages).

rkpagadala:
Closing this pull request create issue #784

PR #790, opened 2017-06-02
Add support for delta encoding of attributes in ReportRequest

geeknoid:
Also, stop returning APA-produced attributes with every request.

This bumps code coverage of pkg/api to 100%.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/790)

### DIFF
--- a/pkg/api/grpcServer.go
+++ b/pkg/api/grpcServer.go
@@ -49,19 +49,27 @@ type (
 		wordMap map[string]int32
 	}
 
-	// dispatchState holds the set of information used for dispatch and
+	// dispatchState holds the information used for dispatch and
 	// request handling.
 	dispatchState struct {
 		inAttrs, outAttrs *mixerpb.Attributes
 		result            *rpc.Status
+		requestBag        *attribute.MutableBag // optional
 	}
 
 	dispatchFn func(ctx context.Context, requestBag *attribute.MutableBag, responseBag *attribute.MutableBag) rpc.Status
 )
 
+var (
+	// TODO: need full list
+	globalWordList = []string{
+		"request.size",
+	}
+)
+
 // NewGRPCServer creates a gRPC serving stack.
 func NewGRPCServer(aspectDispatcher adapterManager.AspectDispatcher, gp *pool.GoroutinePool) mixerpb.MixerServer {
-	words := []string{"request.size"} // TODO: need full list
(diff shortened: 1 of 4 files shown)


pkg/attribute/mutableBag.go
```diff
@@ -305,73 +319,79 @@ func GetBagFromProto(attrs *mixerpb.Attributes, globalDict []string) (*MutableBa
 	// TODO: fail if the proto carries multiple attributes by the same name (but different types)
 
 	var buf *bytes.Buffer
+	log := func(format string, args ...interface{}) {}
```
ZackButcher: Nice, much cleaner.

pkg/api/grpcServer.go
```diff
@@ -87,33 +96,27 @@ func (s *grpcServer) dispatch(ctx context.Context, dState *dispatchState, worker
 		return makeGRPCError(out)
 	}
 
-	responseBag := attribute.GetMutableBag(nil)
+	preprocResponseBag := attribute.GetMutableBag(requestBag)
```
douglas-reid: does this get serialized anywhere? it'd be nice to dump this all in one place, perhaps close to where we log the response (for debug levels).

pkg/attribute/mutableBag.go
```diff
@@ -305,73 +319,79 @@ func GetBagFromProto(attrs *mixerpb.Attributes, globalDict []string) (*MutableBa
...
+
 	if glog.V(2) {
 		buf = pool.GetBuffer()
-		log(buf, "Creating bag from wire attributes:\n")
+		log = func(format string, args ...interface{}) {
+			fmt.Fprintf(buf, format, args...)
```
douglas-reid: suggestion: all the log() calls end with "\n". perhaps add that detail here, and save having to add it to each call.

maybe something like:
```
if buf.Bytes()[buf.Len()-1] != '\n' {
  buf.WriteByte('\n')
}
```

pkg/api/grpcServer.go
geeknoid: I added code to dump out all the attributes before invoking the normal adapters.

pkg/attribute/mutableBag.go
geeknoid: Done.

PR #793, opened 2017-06-05
Introduce the Mixer IL, interpreter and compiler.

ozevren:
This CL introduces Mixer IL, interpreter and compiler.

The Mixer IL is an intermediate-language, geared towards efficient processing of incoming attributes to Mixer. It is accompanied by an interpreter that takes the IL based programs and attribute bags to perform evaluations. There is also a compiler that can convert Mixer expressions to IL based programs.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/793)

### DIFF
--- a/pkg/il/builder.go
+++ b/pkg/il/builder.go
@@ -1,0 +1,278 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package il
+
+import (
+	"fmt"
+)
+
+// Builder is used for building function bodies in a programmatic way.
+type Builder struct {
+	s      *StringsTable
+	b      []uint32
+	labels map[string]uint32
+	fixups map[string][]uint32
+}
+
+// NewBuilder creates a new Builder instance. It uses the given StringsTable when it needs to
+// allocate ids for strings.
+func NewBuilder(s *StringsTable) *Builder {
+	return &Builder{
+		s:      s,
+		b:      make([]uint32, 0),
+		labels: make(map[string]uint32),
+		fixups: make(map[string][]uint32),
+	}
+}
+
+// Build completes building a function body and returns the accumulated byte code.
+func (f *Builder) Build() []uint32 {
+	return f.b
+}
+
+// Nop appends the "nop" instruction to the byte code.
+func (f *Builder) Nop() {
+	f.op0(Nop)
+}
+
+// Ret appends the "ret" instruction to the byte code.
+func (f *Builder) Ret() {
+	f.op0(Ret)
+}
+
+// Call appends the "call" instruction to the byte code.
+func (f *Builder) Call(fnName string) {
+	f.op1(Call, f.id(fnName))
+}
+
+// ResolveInt appends the "resolve_i" instruction to the byte code.
+func (f *Builder) ResolveInt(n string) {
+	f.op1(ResolveI, f.id(n))
+}
+
+// TResolveInt appends the "tresolve_i" instruction to the byte code.
+func (f *Builder) TResolveInt(n string) {
+	f.op1(TResolveI, f.id(n))
+}
+
+// ResolveString appends the "resolve_s" instruction to the byte code.
+func (f *Builder) ResolveString(n string) {
+	f.op1(ResolveS, f.id(n))
+}
+
+// TResolveString appends the "tresolve_s" instruction to the byte code.
+func (f *Builder) TResolveString(n string) {
+	f.op1(TResolveS, f.id(n))
+}
+
+// ResolveBool appends the "resolve_b" instruction to the byte code.
+func (f *Builder) ResolveBool(n string) {
+	f.op1(ResolveB, f.id(n))
+}
+
+// TResolveBool appends the "tresolve_b" instruction to the byte code.
+func (f *Builder) TResolveBool(n string) {
+	f.op1(TResolveB, f.id(n))
+}
+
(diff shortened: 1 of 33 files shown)


douglas-reid:
Should this go on a branch for evaluation before we decide to mainline it ?

ozevren:
I can do that. ( I don't think the instructions mentions branches/evalution
though, or I missed it).

"il" branch?

kyessenov:
is there a design doc for the IL?

ozevren:
There is package level write-up on the IL/instructions and interpreter in
code.

douglas-reid:
@ozevren maybe: `il-experiment` or `mixer-il` for the branch ?

We probably haven't documented this workflow, because we haven't really encountered it yet :). For introducing a component this large and potentially complex, I think we want to keep it separate for a bit. I think there is likely going to be a fair amount of discussion around adding an IL.

ozevren:
Ok. I'll go with mixer-il.

This is my first attempt at contribution in a long while, so do please let
me know if you guys want me go through some steps. I am happy to go through
them, just let me know. :)

ZackButcher:
Yeah, it'd be nice to see a design - the package comments describe "what" but not "why". I'd also really like to see this split up into multiple, incremental PRs (or at least multiple commits so each commit can be looked at independently). It's impossible to review this much code and actually find the problems.

ozevren:
Will do. Should I still send the CLs, or do you want to take a look at a
design doc first?

ZackButcher:
Doesn't matter to me - maybe split things up so we can start the reviews first, since they'll take a while, and produce the design while you're blocked on reviews?

mandarjog:
@ozevren I think the best way to go about it to open a "mixer issue" as
"Proposal: Create Mixer IL, interpreter and compiler for expression language".
You can answer the why / how questions there.
Thanks.

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,227 @@
...
+	pTypes []il.Type
+	rType  il.Type
+
+	v reflect.Value
+}
+
+// ExternFromFn creates a new reflection based Extern, based on the given function. It panics if the
+// function signature is incompatible to be an extern.
+//
+// A function can be extern under following conditions:
```
geeknoid: under the following

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,227 @@
...
+// - Input parameter types are one of the supported types: string, bool, int64, float64, map[string]string.
```
geeknoid: We don't support time.Duration, time.Time, and []byte?

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,227 @@
...
+	outs := e.v.Call(ins)
+	var rv reflect.Value
+	switch len(outs) {
+	case 1:
+		if e.rType != il.Void {
+			rv = outs[0]
+			break
+		}
+		i := outs[0].Interface()
+		if i != nil {
```
geeknoid: Combine with previous line.

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,227 @@
...
+			return 0, 0, i.(error)
+
+		}
+	case 2:
+		rv = outs[0]
+		i := outs[1].Interface()
+		if i != nil {
```
geeknoid: Combine with previous line.

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	switch e.rType {
+	case il.String:
+		str := rv.String()
+		id := s.GetID(str)
+		return id, 0, nil
+
+	case il.Bool:
+		b := rv.Bool()
```
geeknoid: Simplify:

if rb.Bool() {
  return 1, 0, nil
}
return 0, 0, nil

pkg/il/read.go
```diff
@@ -0,0 +1,369 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// ReadText parses the given il assembly text and converts it into a Program.
+func ReadText(text string) (Program, error) {
+	p := NewProgram()
+	err := MergeText(text, p)
+	if err != nil {
```
geeknoid: Combine with previous line

pkg/il/builder.go
```diff
@@ -0,0 +1,278 @@
...
+	f.op0(Lookup)
+}
+
+// ILookup appends the "ilookup" instruction to the byte code.
+func (f *Builder) ILookup(v string) {
+	f.op1(ILookup, f.id(v))
+}
+
+// AllocateLabel allocates a new label value for use within the code.
+func (f *Builder) AllocateLabel() string {
```
ozevren: You may need to reference a label before you need to declare it. i.e.:
jmp Foo
...
Foo:

In that case, you need to allocate the label first, use it in jmp and then set its position.

pkg/il/function.go
```diff
@@ -0,0 +1,88 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package il
+
+// Function contains metadata about an IL-based function that is implemented in a program.
+type Function struct {
+	// Name is the name of the function. It needs to be globally unique.
+	Name string
```
ozevren: Interesting. Yes that works. Thanks for the suggestion.

pkg/il/generate.sh
```diff
@@ -0,0 +1,42 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+// =================================================================================================
+// WARNING! This file is auto-generated. Do not hand-edit. Instead, edit interpreterRun.got
```
ozevren: It is not a valid .go file and makes editing really hard when intellisense barfs due to preprocessor directives. 
I can still rename it if this is big concern though.

pkg/il/interpreter/extern.go
ozevren: Not yet? We can add support for these. time.Duration is int64, so effectively we can already support it. time.Time seems a bit heavy weight (64bits + 32bits + pointer), which needs to be kept in "heap". Similarly for bytes.

PR #794, opened 2017-06-05
Add more unit tests for pkg/attribute

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/794)

### DIFF
--- a/pkg/attribute/bag_test.go
+++ b/pkg/attribute/bag_test.go
@@ -115,51 +115,6 @@ func TestBag(t *testing.T) {
 	}
 }
 
-/*
-func TestStringMapEdgeCase(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/attribute/bag_test.go
```diff
@@ -373,6 +432,30 @@ func withPanic(f func()) (ret interface{}) {
...
+		v1, _ := b1.Get(name)
+		v2, _ := b2.Get(name)
+
+		match, _ := compareAttributeValues(v1, v2)
+		if !match {
+			return false
+		}
+	}
+
+	for _, name := range b2.Names() {
```
ZackButcher: nit: You could check `len(b1) == len(b2)` then only loop over one of the bags.

PR #805, opened 2017-06-08
Fix bad handling of error cases in Check path.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/805)


ldemailly:
if this fixes the test can you remove the skip?

geeknoid:
This is completely unrelated to the E2E test failure.

PR #809, opened 2017-06-09
Add words from our new global dictionary.

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/809)


pkg/api/grpcServer.go
```diff
@@ -61,9 +61,119 @@ type (
 )
 
 var (
-	// TODO: need full list
+	// global word list as defined in https://github.com/istio/api/blob/master/mixer/v1/global_dictionary.yaml
 	globalWordList = []string{
```
rkpagadala: why is globalWordList a var and not a const?

pkg/api/grpcServer.go
```diff
@@ -61,9 +61,119 @@ type (
 )
 
 var (
-	// TODO: need full list
+	// global word list as defined in https://github.com/istio/api/blob/master/mixer/v1/global_dictionary.yaml
```
qiwzhang: add comment to state which version of the file. now it is the first version.
we can add a version comment in the global dictionary file next time we update it.

pkg/api/grpcServer.go
ZackButcher: You can't declare const arrays in golang, unfortunately.

pkg/api/grpcServer.go
```diff
@@ -61,9 +61,119 @@ type (
 )
 
 var (
```
douglas-reid: would it be worth having this in a separate file -- something like //pkg/api/globalwordlist.go ?  It would be cool if we could autogen from the global_dictonary.yaml too.

ZackButcher: Is there a compelling reason to hard code this over ingesting it as a yaml file? Then we can just use the global_dictonary.yaml directly. Eventually we'll need to manage rolling out changes to this list (and probably handle having multiple versions to serve with at the same time), so we'll want it to not be compiled in anyway.

pkg/api/grpcServer.go
geeknoid: The URL now includes the specified commit SHA.

pkg/api/grpcServer.go
geeknoid: https://docs.google.com/document/d/1NcGgEzSEt0A0_qqJfA3x6ETNyomyMrhm1sBlpdorbHY/edit#heading=h.la7bf3j6tcb5 explains the update procedures for the dictionary. The reason we want to build this into the code is that it's not really config. I don't want operators to mess around with this stuff willy nilly.

geeknoid:
I added a reference to the specific commit the list comes from, and I split the word list into its file such that we can potentially code gen the thing directly from the yaml definition file.

PTAL.

pkg/api/grpcServer.go
rkpagadala: Ahh.. Thanks Zack

PR #814, opened 2017-06-12
Use proto.Marshal instead of gob.Encode to compute executor cache keys

ZackButcher:
Use proto.Marshal instead of gob.Encode to produce a byte array of the input params. This is because gob requires reflection to determine a type for the value it's encoding; for some reason subfields on a proto oneof don't play well with this. Our config ingestion code already requires that the params fields be protos, so this change is safe.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/814)

### DIFF
--- a/pkg/adapterManager/BUILD
+++ b/pkg/adapterManager/BUILD
@@ -20,6 +20,7 @@ go_library(
         "//pkg/expr:go_default_library",
         "//pkg/pool:go_default_library",
         "//pkg/status:go_default_library",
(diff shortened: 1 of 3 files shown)


ZackButcher:
Hold off on reviewing this for a bit: I didn't read enough of the proto API and there is a way we can specify our own buffers to marshal in to. I'll update this PR to use [`proto.Buffer`](https://godoc.org/github.com/golang/protobuf/proto#NewBuffer).

ZackButcher:
Updated to use the other API, PTAL.

PR #818, opened 2017-06-13
fix the registration flow of redis backend.

jmuk:
It was registered through init(), but it turned out wrong,
since the code is not linked well because it looks like a dead code.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/818)


pkg/config/manager.go
```diff
@@ -227,3 +228,7 @@ func (c *Manager) Start() {
 	c.ticker = time.NewTicker(c.loopDelay)
 	go c.loop()
 }
+
+func init() {
```
mandarjog: Can we put all the different storage backends in an `inventory()` similar to the `/adapter/inventory` instead?

jmuk: done. PTAL.

PR #825, opened 2017-06-16
Treat empty strings as nil in expr

mandarjog:
If "user" header was not specified, previously the expression below used to evaluate to
```
request.header["user"] | "unknown"   ---> ""
```
Now we treat empty strings as nil so 
```
request.header["user"] | "unknown"   ---> "unknown"
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/825)

### DIFF
--- a/pkg/expr/eval_test.go
+++ b/pkg/expr/eval_test.go
@@ -164,6 +164,19 @@ func TestGoodEval(tt *testing.T) {
 			true, "",
 		},
 		{
+			`origin.name | "users"`,
+			map[string]interface{}{
+				"origin.name": "",
+			},
+			"users", "",
+		},
+		{
+			`origin.name | "users"`,
+			map[string]interface{}{
+			},
+			"users", "",
+		},
+		{
 			`(x/y) == 30`,
 			map[string]interface{}{
 				"x": int64(20),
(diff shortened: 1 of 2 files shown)


mandarjog:
Now we will hope that no e2e tests rely on the old behaviour :-)

pkg/expr/eval_test.go
```diff
@@ -223,12 +244,12 @@ func TestGoodEval(tt *testing.T) {
 				if tst.err == "" {
 					t.Errorf("[%d] unexpected error: %v", idx, err)
 				} else if !strings.Contains(err.Error(), tst.err) {
-					t.Errorf("[%d] got %s\nwant %s", idx, err, tst.err)
+					t.Errorf("[%d] got <%s>\nwant <%s>", idx, err, tst.err)
```
rkpagadala: suggesstion use %q instead of %s

pkg/expr/eval_test.go
```diff
@@ -223,12 +244,12 @@ func TestGoodEval(tt *testing.T) {
...
 				}
 				return
 			}
 			if res != tst.result {
-				t.Errorf("[%d] %s got %s\nwant %s", idx, exp.String(), res, tst.result)
+				t.Errorf("[%d] %s got <%s>\nwant <%s>", idx, exp.String(), res, tst.result)
```
rkpagadala: same here

pkg/expr/func.go
```diff
@@ -213,7 +213,9 @@ func newOR() Func {
 func (f *orFunc) Call(attrs attribute.Bag, args []*Expression, fMap map[string]FuncBase) (interface{}, error) {
 	for _, arg := range args {
 		ret, _ := arg.Eval(attrs, fMap)
-		if ret != nil {
+		// treating empty strings as nil, since
+		// go strings cannot be nil
+		if ret != nil && ret != "" {
```
rkpagadala: suggestion use
if len(ret) != 0 {

mandarjog: Updated the other 2...

ret is of type `interface{}` is it still better to check  length. I will have to type assert first.
`ret != ""` seems succinct.

douglas-reid:
build mixer/e2e-smoketest

douglas-reid:
build mixer/e2e-smoketest

PR #848, opened 2017-06-28
Introduce bazel rule for doing mixer codegen.

douglas-reid:
This PR bazel-ifies the code in //tools, adding a generate.bzl file that includes a new build rule `mixer_proto_library`.

Example usage:
```
mixer_proto_library(
    name = "go_default_library",
    protos = ["my_new_aspect.proto"],
)
```

The PR also enables the previously commented out tests and rearranges their setup to make them work with bazel (and exercise the new build rule). Part of that work included adding in programmatic gofmt and goimport organization to our generated code.  The rest of that work included adding build rules for all of the testdata to generate the file descriptors in advance of the tests.

There were edits to the `bazel_to_go.py` script to include the template generated code in the builds, as well.

There are probably several aspects of the generate script that can be cleaned up, but I anticipate having to revisit this when the code generation story is fleshed out a bit more. So, I'm holding off on trying to make it shiny for now.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/848)


tools/codegen/pkg/interfacegen/generator_test.go
```diff
@@ -15,159 +15,89 @@
...
-	return mvCmd.Run()
-}
+		b2 := make([]byte, chunkSize)
+		s2, err2 := f2.Read(b2)
 
-func getBaseFileNameWithoutExt(filePath string) string {
-	tmp := filepath.Base(filePath)
-	return tmp[0 : len(tmp)-len(filepath.Ext(tmp))]
+		if err1 != nil || err2 != nil {
+			if err1 == io.EOF && err2 == io.EOF {
```
guptasu: can this be just 
s2, err2 := f2.Read(b2)
if err1 == io.EOF && err2 == io.EOF {
  return true
}
if !bytes.Equal(b1, b2) {
	logf("bytes don't match (sizes: %d, %d):\n%s\n%s", s1, s2, string(b1), string(b2))
	return false
}

tools/codegen/pkg/interfacegen/generator_test.go
```diff
@@ -15,159 +15,89 @@
...
-		"-I=.",
-		"-I=api",
+	b, fileErr := ioutil.ReadFile("testdata/ErrorTemplate.baseline")
+	if fileErr != nil {
+		t.Fatalf("Could not read baseline file: %v", err)
 	}
+	want := fmt.Sprintf("%s", b)
+	got := err.Error()
+	if got != want {
+		t.Fatalf("Generate(%s) => '%s'\nwanted: '%s'", "testdata/error_template.descriptor_set", got, want)
```
guptasu: Can we print the file where the got file will be placed. Will be easy to go and copy/replace when creating new tests  ?

tools/codegen/pkg/interfacegen/testdata/BUILD
```diff
@@ -0,0 +1,84 @@
...
+    srcs = [
+        "ListTemplateProcessorInterface.golden.go",
+        "LogTemplateProcessorInterface.golden.go",
+        "MetricTemplateProcessorInterface.golden.go",
+        "NestedMessageProcessorInterface.golden.go",
+        "QuotaTemplateProcessorInterface.golden.go",
+    ],
+)
+
+proto_compile(
```
guptasu: Can't we make this test also like the one above ? It will generate a baseline that has errors

tools/codegen/pkg/interfacegen/testdata/BUILD
```diff
@@ -0,0 +1,84 @@
+package(default_visibility = ["//tools/codegen/pkg/interfacegen:__pkg__"])
+
+load("@org_pubref_rules_protobuf//protobuf:rules.bzl", "proto_compile")
+load("//tools/codegen:generate.bzl", "mixer_proto_library")
+
+mixer_proto_library(
+    name = "list_template_library",
+    testonly = True,
+    protos = ["ListTemplate.proto"],
```
guptasu: Should we add a tests that wants to import things outside the well known imports (which mixer_proto_library will include for them)?

tools/codegen/pkg/modelgen/model_test.go
```diff
@@ -34,14 +30,13 @@ func TestErrorInTemplate(t *testing.T) {
...
 		expectedError string
 	}{
-		{"testdata/MissingPackageName.proto", "package name missing"},
-		{"testdata/MissingTemplateNameExt.proto", "Contains only one of the following two options"},
-		{"testdata/MissingTemplateVarietyExt.proto", "Contains only one of the following two options"},
-		{"testdata/MissingBothRequiredExt.proto", "one proto file that has both extensions"},
-		{"testdata/MissingTypeMessage.proto", "message 'Type' not defined"},
-		{"testdata/MissingConstructorMessage.proto", "message 'Constructor' not defined"},
-		{"testdata/ReservedFieldInConstructor.proto", "proto:15: Constructor message must not contain the reserved filed name 'Name'"},
-		{"testdata/Proto2BadSyntax.proto", "Proto2BadSyntax.proto:3: Only proto3 template files are allowed."},
```
guptasu: Can we include the line numbers ? Were they causing issues ?

tools/codegen/generate.bzl
```diff
@@ -0,0 +1,120 @@
...
+}
+
+PROTO_IMPORTS = ["external/com_github_google_protobuf/src"]
+PROTO_INPUTS = [ "@com_github_google_protobuf//:well_known_protos" ]
+
+
+def mixer_proto_library(
+    name,
+    protos = [],
+    importmap = {},
```
guptasu: May be I missed it. But I was not able to find a test that exercises the fields below protos (example importmap, imports, inputs, deps)

tools/codegen/cmd/mixgenproc/main.go
```diff
@@ -40,17 +40,17 @@ func withArgs(args []string, errorf func(format string, a ...interface{})) {
 			if len(args) == 0 {
 				errorf("Must specify a file descriptor set protobuf file.")
 			}
-			if len(args) != 1 {
-				errorf("Only one input file is allowed.")
-			}
+			//if len(args) != 1 {
```
geeknoid: Delete?

tools/codegen/generate.bzl
```diff
@@ -0,0 +1,120 @@
...
+      "deps": MIXER_DEPS + GOGO_DEPS + deps,
+      "imports": imports + MIXER_IMPORTS + PROTO_IMPORTS,
+      "importmap": dict(dict(MIXER_IMPORT_MAP, **GOGO_IMPORT_MAP), **importmap),
+      "inputs": inputs + MIXER_INPUTS + PROTO_INPUTS,
+      # "output_to_workspace": True,
+      "verbose": verbose,
+   }
+
+   # we run this proto library to get the generated pb.go files to link
+   # in wit the mixer generated files for a go library
```
guptasu: with

tools/codegen/generate.bzl
```diff
@@ -0,0 +1,120 @@
...
+   proto_compile(**proto_compile_args)
+
+   gogoslick_args += {
+      "name": name + "_gogo_proto",
+      "protos": protos,
+      "deps": MIXER_DEPS + GOGO_DEPS + deps,
+      "imports": imports + MIXER_IMPORTS + PROTO_IMPORTS,
+      "importmap": dict(dict(MIXER_IMPORT_MAP, **GOGO_IMPORT_MAP), **importmap),
+      "inputs": inputs + MIXER_INPUTS + PROTO_INPUTS,
+      # "output_to_workspace": True,
```
guptasu: Do we need this comment ?

guptasu: There are few more commented out lines.

tools/codegen/pkg/interfacegen/testdata/BUILD
douglas-reid: no, we cannot. mixer_proto_library will fail because it tries to build a library. there is no mixer_proto_compile step at the moment. I'm not quite sure what the expected output of mixer_proto_compile should be.  I'd prefer to punt on adding that functionality for now.

tools/codegen/generate.bzl
douglas-reid: There probably is not one. I used those explicitly during development and then moved them all into generate.bzl default sets, as I figured they would be widely used.

If it is OK with you, we can add that sort of testing in a subsequent PR.

guptasu: SGTM

tools/codegen/pkg/interfacegen/generator_test.go
```diff
@@ -15,159 +15,89 @@
...
-		{"testdata/QuotaTemplate.proto", "testdata/QuotaTemplateProcessorInterface.go.baseline"},
-		{"testdata/LogTemplate.proto", "testdata/LogTemplateProcessorInterface.go.baseline"},
-		{"testdata/ListTemplate.proto", "testdata/ListTemplateProcessorInterface.go.baseline"},
-		{"testdata/NestedMessage.proto", "testdata/NestedMessageProcessor.go.baseline"},
-		{"testdata/ErrorTemplate.proto", "testdata/ErrorTemplate.baseline"},
+		{"Metrics", "testdata/metric_template_library_processor.gen.go", "testdata/MetricTemplateProcessorInterface.golden.go"},
+		{"Quota", "testdata/quota_template_library_processor.gen.go", "testdata/QuotaTemplateProcessorInterface.golden.go"},
+		{"Logs", "testdata/log_template_library_processor.gen.go", "testdata/LogTemplateProcessorInterface.golden.go"},
+		{"Lists", "testdata/list_template_library_processor.gen.go", "testdata/ListTemplateProcessorInterface.golden.go"},
+		{"Nested Message", "testdata/nested_message_library_processor.gen.go", "testdata/NestedMessageProcessorInterface.golden.go"},
```
douglas-reid: bazel rule for the test defines data deps, which is how these files are generated.

I have not tried to ensure that any of this works from gogland. Nor have I tried to trigger bazel build rules from with go code. I'm not sure we want to execute bazel build rules from the go code (we would have this weird bazel -> go -> bazel setup which might make it hard to access generated artifacts, etc.).

A different approach would be to have a separate test with known file descriptor protos as inputs, requiring only a call to generate.  That would not exercise the generate.bzl rules, but we could add separate tests for that.

Thoughts?

tools/codegen/pkg/interfacegen/generator.go
```diff
@@ -17,13 +17,16 @@ package interfacegen
...
 	"fmt"
+	"go/format"
 	"io/ioutil"
 	"os"
 	"text/template"
 
-	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/gogo/protobuf/proto"
+	"github.com/gogo/protobuf/protoc-gen-gogo/descriptor"
```
douglas-reid: here: because we use gogo to compile the other protos. having both gogo and golang protobuf deps causes all sorts of issues. I think we want everything to be gogo at the moment.

tools/codegen/pkg/modelgen/model_test.go
douglas-reid: This seems like an issue with merging (i did not intentionally strip those out). I'll investigate.

tools/codegen/generate.bzl
douglas-reid: Good catch, thanks.

tools/codegen/generate.bzl
douglas-reid: We needed this uncommented actually (so that `bin/bazel_to_go.py` can put things in the right place)

tools/codegen/pkg/interfacegen/generator_test.go
douglas-reid: I've updated to use *.descriptor_set files and direct calls to Generate(). I've also updated the `bin/bazel_to_go.py` to put everything in the right place for gogland/IDE/go testing.  Look for it in the next commit I push up.

tools/codegen/pkg/interfacegen/generator_test.go
douglas-reid: Good catch! I've simplified the logic here a bit.

tools/codegen/pkg/modelgen/model_test.go
douglas-reid: I still need to investigate this.

douglas-reid: Fixed in latest commit.

tools/codegen/pkg/interfacegen/testdata/BUILD
douglas-reid: I've filed: https://github.com/istio/mixer/issues/859 to track this outside of this PR (to avoid making this PR gigantic).

douglas-reid:
Thanks for the thorough review - caught a number of important issues. I think the current set of commits now addresses all concerns.

PTAL.

douglas-reid:
build mixer/e2e-smoketest

tools/codegen/generate.bzl
mandarjog: it's not clear if `output_to_workspace` should be uncommented here? otherwise bazel_to_go.py won't find stuff, like you say in your comment

douglas-reid: oh sorry, we needed the `output_to_genfiles` bit in the `mixer_gen` rule. `output_to_workspace` is not needed here (and was included only for validation during development).

PR #856, opened 2017-06-29
Implement the new combined Check+Quota API.

geeknoid:
This is a breaking change.

We'll do a quick coordinated set of checkins. First, the change in the API repo will be submitted. Then this PR will be updated to point to the right SHA of the API repo and be pushed in. The mixer client will then be pushed in. And finally, the proxy will be updated to use the new mixer client.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/856)


costinm:
It would be great if we can avoid 'breaking changes' - add the new feature
without breaking the old one, and after it
gets enough testing and it is stable remove the old.

And big thank you for combining quota and check :-)

It might be useful to allow (via some config) separate quota calls: just
like report, quota can use
batching optimizations, and in some cases it can use sampling and other
techniques to reduce RT.
However this may not be needed - a developer may use a different quota
plugin independently.

geeknoid:
The API is designed to allow prefetching of quota tokens, so we do effectively have batching semantics for quota.

Once caching and prefetching are fully enabled, the proxy should be able to run autonomously for extended periods of time without ever talking to Mixer. After this PR is in, getting caching and prefetching fully plumbed through is the next order of business.

douglas-reid:
Can we restore the kubernetes runtime config in: `testdata/configroot/scopes/global/subjects/global/rules.yml` ?

Unfortunately, we use that to generate a bunch of attributes which are used elsewhere.

Its removal does not seem important to this PR.

geeknoid:
Yes, it can be restored. I dont think I intended to check those changes in.
I had to make 'em in order for stuff to work for me locally.

douglas-reid:
OK. Thanks.

We really need to separate out our "default deployment config" from our "local testing config."  I think I may be the only one that uses local kubeconfig files to run Mixer locally.

I'm sure that other developers with other preprocessors will have similar issues.

PR #874, opened 2017-07-01
Remove custom script for munging cfg.proto.

### LINKED ISSUE #871: Replace build_cfg.sh with bazel rules

douglas-reid:
This PR moves the logic of the script into a bazel genrule(). This
will allow repeatable generation of the *.pb.go file and related
library.

Also, this PR finds and destroys all competing references to istio/api
version of the config proto library. This forces all code to use the
generated and fixed library version.

There was an issue with gogoslick_proto_compile that prevented use.
While I investigate that, this PR uses gogo_proto_compile as a stand-in.

Fixes #871.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/874)


pkg/config/proto/BUILD
```diff
@@ -5,8 +5,8 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library")
 go_library(
     name = "go_default_library",
     srcs = [
-        "cfg.pb.go",
         "combined.go",
+        "@com_github_istio_api//:mixer/v1/config_fixed",
```
guptasu: nice

cmd/server/cmd/BUILD
```diff
@@ -31,7 +31,6 @@ go_library(
         "@com_github_grpcecosystem_opentracing//go/otgrpc:go_default_library",
         "@com_github_grpcecosystem_prometheus//:go_default_library",
         "@com_github_istio_api//:mixer/v1",
-        "@com_github_istio_api//:mixer/v1/config",
```
guptasu: Does all the files from where you have removed "@com_github_istio_api//:mixer/v1/config"', do we need to reference mixer/v1/config_fixed ?

douglas-reid: Everywhere that we used "@com_github_istio_api//:mixer/v1/config", we also used: "//pkg/config/proto:go_default_library". Since `cfg.proto` is the primary inclusion in both libraries, this should not be necessary.

BUILD.api
```diff
@@ -109,3 +104,17 @@ filegroup(
...
+# TODO: when gogoslick_proto_compile rule is fixed for use, remove the
+# VALUE_TYPE enum fix part of the sed.
+genrule(
+    name = "mixer/v1/config_fixed",
+    srcs = [":mixer/v1/config_gen"],
+    outs = ["fixed_cfg.pb.go"],
+    cmd = "sed " +
+          "-e 's/*google_protobuf.Struct/interface{}/g' " +
+          "-e 's/ValueType_VALUE_TYPE_UNSPECIFIED/VALUE_TYPE_UNSPECIFIED/g' " +
+          "$(location :mixer/v1/config_gen) | $(location @org_golang_x_tools_imports//:goimports) >> $@",
```
mandarjog: why `>> $@` not `> $@`

douglas-reid: fat fingers. fixed.

PR #875, opened 2017-07-01
Add pprof support to mixer.

douglas-reid:
With this PR, we can use commands like the following to improve our understanding of Mixer's runtime:
- `go tool pprof http://localhost:9093/debug/pprof/heap`
- `go tool pprof http://localhost:9093/debug/pprof/profile`
- `wget http://localhost:6060/debug/pprof/trace?seconds=5`

More info available here: https://blog.golang.org/profiling-go-programs

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/875)

### DIFF
--- a/cmd/server/cmd/server.go
+++ b/cmd/server/cmd/server.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	_ "net/http/pprof" // For profiling / performance investigations
 	"os"
 	"strings"
 	"time"


douglas-reid:
From: http://artem.krylysov.com/blog/2017/03/13/profiling-and-optimizing-go-web-applications/

> One of the biggest pprof advantages is that it has low overhead and can be used in a production environment on a live traffic without any noticeable performance penalties.

If we find that isn't true, we can always remove it.

PR #880, opened 2017-07-05
Switch stdioLogger adapter to use zap

douglas-reid:
This PR transitions the stdioLogger adapter implementation from using the standard json encoder to serialize logs to [zap](https://github.com/uber-go/zap).

About zap:
> It includes a reflection-free, zero-allocation JSON encoder, and the base Logger strives to avoid serialization overhead and allocations wherever possible.

For comparision, using the newly added (extremely raw) benchmark, I saw the following on my laptop:

#### Original implementation, using a `logStream` of `ioutil.Discard`:
```
BenchmarkLogger_Log-8             200000             10930 ns/op            2801 B/op         76 allocs/op
PASS
ok      istio.io/mixer/adapter/stdioLogger      2.337s
```
#### With zap:
```
BenchmarkLogger_Log-8            3000000               461 ns/op             384 B/op          1 allocs/op
PASS
ok      istio.io/mixer/adapter/stdioLogger      1.899s
```

The reduction in `allocs/op` from 76 to 1 convinced me that this may be worth pursuing.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/880)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -496,12 +496,6 @@ new_go_repository(
     importpath = "go.uber.org/atomic",
 )
 
-new_go_repository(
-    name = "org_uber_go_multierr",
-    commit = "3c4937480c32f4c13a875a1829af76c98ca3d40a",  # Jun 30, 2017 (v1.1.0)
-    importpath = "go.uber.org/multierr",
-)
-
 ##
 ## Testing
 ##
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -484,6 +484,24 @@ new_go_repository(
     importpath = "golang.org/x/tools",
 )
 
+new_go_repository(
+    name = "org_uber_go_zap",
+    commit = "9cabc84638b70e564c3dab2766efcb1ded2aac9f",  # Jun 8, 2017 (v1.4.1)
+    importpath = "go.uber.org/zap",
+)
+
+new_go_repository(
+    name = "org_uber_go_atomic",
+    commit = "4e336646b2ef9fc6e47be8e21594178f98e5ebcf",  # Apr 12, 2017 (v1.2.0)
+    importpath = "go.uber.org/atomic",
+)
+
+new_go_repository(
+    name = "org_uber_go_multierr",
+    commit = "3c4937480c32f4c13a875a1829af76c98ca3d40a",  # Jun 30, 2017 (v1.1.0)
+    importpath = "go.uber.org/multierr",
+)
+
 ##
 ## Testing
 ##
--- a/adapter/stdioLogger/stdioLogger.go
+++ b/adapter/stdioLogger/stdioLogger.go
@@ -30,6 +30,7 @@ import (
 type (
 	builder struct{ adapter.DefaultBuilder }
 	logger  struct {
+		omitEmpty   bool
 		outputPaths []string
 		impl        *zap.Logger
 	}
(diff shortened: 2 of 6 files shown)


WORKSPACE
```diff
@@ -484,6 +484,24 @@ new_go_repository(
...
+)
+
+new_go_repository(
+    name = "org_uber_go_atomic",
+    commit = "4e336646b2ef9fc6e47be8e21594178f98e5ebcf",  # Apr 12, 2017 (v1.2.0)
+    importpath = "go.uber.org/atomic",
+)
+
+new_go_repository(
+    name = "org_uber_go_multierr",
```
ZackButcher: This is a transitive dep? (I ask because in `stdioLogger` you still import hashicorp's `multierror`.)

douglas-reid: Yes. Transitive dep.  I'll test with leaving it out.

ZackButcher: I assumed. It'd be really nice if Bazel provided a way for us to split this stuff up. Like a top level WORKSPACE with sub-WORKSPACE files for different repo's deps (e.g. zipkin's opentracing is gonna pull in like 13 new repos; it'd be nice to have them tucked away somewhere zipkin specific).

douglas-reid:
PTAL. I shuffled some stuff around to preserve 100% test coverage and removed the unnecessary and problematic Sync() calls (see: https://github.com/uber-go/zap/issues/370).

adapter/stdioLogger/stdioLogger.go
```diff
@@ -17,19 +17,36 @@
...
+	logger  struct {
+		outputPaths []string
+		impl        *zap.Logger
+	}
+	zapBuilderFn func(outputPaths ...string) (*zap.Logger, error)
+)
+
+const (
+	stdErr = "stderr"
+	stdOut = "stdout"
```
ZackButcher: Super-nit: I question whether these actually add value: we only use them once in stdioLogger, and only a few times in the tests. It's not like stdout/stderr are likely to change any time soon.

adapter/stdioLogger/stdioLogger.go
```diff
@@ -72,18 +94,43 @@ func (l *logger) LogAccess(entries []adapter.LogEntry) error {
 }
 
 func (l *logger) log(entries []adapter.LogEntry) error {
-	var errors *me.Error
+	var errors *multierror.Error
+	fields := make([]zapcore.Field, 0, 6)
 	for _, entry := range entries {
-		if err := writeJSON(l.logStream, entry); err != nil {
-			errors = me.Append(errors, err)
+		if entry.LogName != "" {
```
ZackButcher: Does Zap not handle all of these zero values correctly already? In other words, could we do this without all the checks, like we would for proto3?

adapter/stdioLogger/stdioLogger.go
douglas-reid: They may not, really. I just tend to screw up typing a bunch and it is an easy thing to miss when (self-) reviewing, so i was trying to put them in one spot to eliminate my own tendency towards typos.

When this whole thing is rewritten in two weeks, we can revisit ;).

adapter/stdioLogger/stdioLogger.go
douglas-reid: It does not.  For instance, eliminating the checks leads to test failures like:

```
got: [{"logName":"istio_log","timestamp":"2017-Jan-09","severity":"INFO","labels":{},"textPayload":"","structPayload":null}],
want: [{"logName":"istio_log","timestamp":"2017-Jan-09","severity":"INFO"}]
```

douglas-reid: I guess you could argue that _might_ be preferable ??  Thoughts?

PR #883, opened 2017-07-06
Update dot generation to double-quote node names

douglas-reid:
See: https://github.com/istio/issues/issues/43

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/883)


douglas-reid:
Great suggestion!

PR #885, opened 2017-07-07
Introduce the MixerIL interpreter and beefing up the IL tests.

ozevren:
This is part two of three part check-in. It builds the interpreter
on top of the MixerIL. The Interpreter can run programs in bytecoe
mode and evaluate them against property bags.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/885)

### DIFF
--- a/pkg/il/interpreter/extern.go
+++ b/pkg/il/interpreter/extern.go
@@ -1,0 +1,225 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package interpreter
+
+import (
+	"reflect"
+
+	"istio.io/mixer/pkg/il"
+)
+
+// Extern represents an external, native function that is callable from within the interpreter,
+// during program execution.
+type Extern struct {
+	name       string
+	paramTypes []il.Type
+	returnType il.Type
+
+	v reflect.Value
+}
+
+// ExternFromFn creates a new, reflection based Extern, based on the given function. It panics if the
+// function signature is incompatible to be an extern.
+//
+// A function can be extern under the following conditions:
+// - Input parameter types are one of the supported types: string, bool, int64, float64, map[string]string.
+// - The return types can be:
+//   - none                                 (i.e. func (...) {...})
+//   - a supported type                     (i.e. func (...) string {...})
+//   - error                                (i.e. func (...) error {...})
+//   - suported type and error              (i.e. func (...) string, error {...})
+//
+func ExternFromFn(name string, fn interface{}) Extern {
+	t := reflect.TypeOf(fn)
+	if t.Kind() != reflect.Func {
+		panic("interpreter.ExternFromFn: not a function")
+	}
+
+	// Validate and calculate return types.
+	iErr := reflect.TypeOf((*error)(nil)).Elem()
+	returnType := il.Void
+	switch t.NumOut() {
+	case 0:
+
+	case 1:
+		if !t.Out(0).Implements(iErr) {
+			returnType = ilType(t.Out(0))
+		}
+
+	case 2:
+		returnType = ilType(t.Out(0))
+		if !t.Out(1).Implements(iErr) {
+			panic("interpreter.ExternFromFn: the second return value is not an error")
+		}
+
+	default:
+		panic("interpreter.ExternFromFn: more than two return values are not allowed")
+	}
+
+	if returnType == il.Unknown {
+		panic("interpreter.ExternFromFn: incompatible return type")
+	}
+
+	// Calculate parameter types.
+	paramTypes := make([]il.Type, t.NumIn())
+	for i := 0; i < t.NumIn(); i++ {
+		pt := t.In(i)
+		ilt := ilType(pt)
+		if ilt == il.Unknown {
+			panic("interpreter.ExternFromFn: incompatible parameter type")
+		}
+		paramTypes[i] = ilt
+	}
+
+	v := reflect.ValueOf(fn)
+
+	return Extern{
+		name:       name,
+		paramTypes: paramTypes,
+		returnType: returnType,
+		v:          v,
+	}
+}
+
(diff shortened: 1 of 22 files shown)


pkg/il/interpreter/interpreterRun.got
```diff
@@ -0,0 +1,756 @@
...
+	var ti64 int64
+	var tu64 uint64
+	var tf64 float64
+	var tVal interface{}
+	var tStr string
+	var tBool bool
+	var tFound bool
+	var tErr error
+
+	opstack = make([]uint32, opStackSize)
```
geeknoid: Since these are of constant size, you could actually allocate them as arrays instead of slices which would be faster.

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,225 @@
...
+
+import (
+	"reflect"
+
+	"istio.io/mixer/pkg/il"
+)
+
+// Extern represents an external, native function that is callable from within the interpreter,
+// during program execution.
+type Extern struct {
```
ozevren: Absolutely, and I think that is the right way to go. Initially, I had such a model here, but the glue code looked extremely ugly and I didn't want to spend any time trying to come up with a decent model for this as it is not needed yet. Think of this as a basic/generalized/proof/fallback implementation for an extern infrastructure.

pkg/il/interpreter/extern.go
```diff
@@ -0,0 +1,225 @@
...
+// - The return types can be:
+//   - none                                 (i.e. func (...) {...})
+//   - a supported type                     (i.e. func (...) string {...})
+//   - error                                (i.e. func (...) error {...})
+//   - suported type and error              (i.e. func (...) string, error {...})
+//
+func ExternFromFn(name string, fn interface{}) Extern {
+	t := reflect.TypeOf(fn)
+	if t.Kind() != reflect.Func {
+		panic("interpreter.ExternFromFn: not a function")
```
ozevren: I don't mind going either way but we should probably discuss first. I've been going back and forth on the error v.s. panic, and finally I've tried to settle on a model that Go seems to be following as well: https://github.com/golang/go/blob/master/src/reflect/type.go
Notice there are many methods here that simply panics. I believe the rationale is, most of these methods get called in a context where the caller either inherently knows that this is the right method to call (i.e. if you're calling reflect.Type.In(), you better make sure it is a function first), and/or have facilities to ensure before making the call (i.e. reflect.Type.Kind()).

Using a similar tact, I've tried to reduce the code paths to use panic where panicking indicates programming error (on our part). For this particular instance, my assumption is things like ExternFromFn (or its non-reflection counterpart) will be mostly called either by us, or by generated code. Trying to bubble these up will complicate the code.

Having said that, I don't mind making the change, I just want to make sure we establish the principle first.

pkg/il/interpreter/interpreterRun.got
ozevren: That is what I started with. :) Then I tried using slices and it turned out to be faster. :)
I will run the benchmark again with arrays to double check and will update if it was only my memory playing tricks on me and arrays were indeed faster.

pkg/il/interpreter/extern.go
ozevren: So what is the final verdict on this? panic or error?

pkg/il/interpreter/interpreterRun.got
ozevren: Ok, here are the results I have. The top one is using slices, the bottom one is using arrays.
I think the results are a bit too noisy to give a resounding victory, but it looks like using arrays either have roughly the same performance, or slightly worse. (BTW, this doesn't measure extern call performance, which I think will be much more sensitive to this change).

I'd prefer to keep what I have right now, but revisit this as we add more perf benchmarks for different scenarios.

ozben-macbookpro2:interpreter ozben$ ./generate.sh
ozben-macbookpro2:interpreter ozben$ go test -run=^$  -bench=.  -benchmem -benchtime=5s
BenchmarkIL/ExprBench/ok_1st-8         	50000000	       119 ns/op	       0 B/op	       0 allocs/op
BenchmarkIL/ExprBench/ok_2nd-8         	50000000	       169 ns/op	       0 B/op	       0 allocs/op
BenchmarkIL/ExprBench/not_found-8      	50000000	       179 ns/op	       0 B/op	       0 allocs/op
BenchmarkExpr/ExprBench/ok_1st-8       	50000000	       139 ns/op	       2 B/op	       2 allocs/op
BenchmarkExpr/ExprBench/ok_2nd-8       	20000000	       318 ns/op	      19 B/op	       4 allocs/op
BenchmarkExpr/ExprBench/not_found-8    	20000000	       317 ns/op	      19 B/op	       4 allocs/op
PASS
ok  	istio.io/mixer/pkg/il/interpreter	44.484s
ozben-macbookpro2:interpreter ozben$ ./generate.sh
ozben-macbookpro2:interpreter ozben$ go test -run=^$  -bench=.  -benchmem -benchtime=5s
BenchmarkIL/ExprBench/ok_1st-8         	50000000	       125 ns/op	       0 B/op	       0 allocs/op
BenchmarkIL/ExprBench/ok_2nd-8         	50000000	       168 ns/op	       0 B/op	       0 allocs/op
BenchmarkIL/ExprBench/not_found-8      	50000000	       186 ns/op	       0 B/op	       0 allocs/op
BenchmarkExpr/ExprBench/ok_1st-8       	50000000	       141 ns/op	       2 B/op	       2 allocs/op
BenchmarkExpr/ExprBench/ok_2nd-8       	20000000	       323 ns/op	      19 B/op	       4 allocs/op
BenchmarkExpr/ExprBench/not_found-8    	20000000	       329 ns/op	      19 B/op	       4 allocs/op
PASS

ozevren:
It looks like the linters are complaining about missing copyright in the generated .go file. I've removed the copyright from that file based on your previous comment on not having copyright on generated files.

It looks like there is no easy way to "suppress", as the check is implemented as a shell scripts. Any advice on how to resolve the situation? I can reintroduce the copyright notice, or I can do an exclusion check on the scripts (or, I can make them check for an auto-generate warning if the copyright doesn't exist).

pkg/il/interpreter/extern.go
geeknoid: Leave it as-is. Using panic means that error paths are eliminated from callers, which is a good  complexity/coverage win.

geeknoid:
You can manually add "exclusions" for known cases directly in linters.sh. If there's no way to do this in a clean way, just go ahead and insert the copyright headers in the generated output.

pkg/il/interpreter/interpreterRun.got
geeknoid: OK, let's leave as-is. Maybe a TODO comment suggesting that using arrays might be a good thing.

ozevren:
Updated the bash-based copyright check linters to filter out things that have "auto-generated" at the beginning of file.

Issue #882, opened 2017-07-06
Add ca-certificates to mixer base image


douglas-reid:
Needed for https calls, etc.

xiaolanz:
Try use https://github.com/GoogleCloudPlatform/distroless

PR #886, opened 2017-07-10
Update Mixer base docker image

douglas-reid:
The current Mixer base image does not include the ca-certificates package. This prevents use of any code that needs to make HTTPS calls out, etc.

This PR adds bazel machinery necessary to install ca-certs and ssl libs on top of the xenial base image.

Fixes: https://github.com/istio/mixer/issues/882

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/886)


douglas-reid:
Adding some more eyes to this PR. It seems there may be a strong preference to remain consistent across Istio components. Let's use this PR thread to discuss.

elevran:
Are there benefits to using [distroless](https://github.com/GoogleCloudPlatform/distroless/) over [Alpine](https://hub.docker.com/_/alpine/) base images? 
I'm sure there are differences (e.g., musl vs glibc), but trying to understand requirements and basis for comparison (e.g., remain consistent across all of Istio, as noted by @douglas-reid above). That requires a separate discussion and possibly creation of shared base image for Istio components as a whole.

Also, just adding CA certificates (as in #882) does not require a new base image. 
```RUN apt-get ca-certificates && update-ca-certificates``` should have been sufficient and, IMO, we should not conflate these two issues.

**Comment by douglas-reid** [comment]
@elevran part of this is complicated by how our docker images are generated currently. There isn't an easy way to force a `RUN ...` command. The new rules_docker stuff does seem to suggest that we can select debian packages, however. Maybe that is the simplest path forward. 

fwiw, distroless was suggested by the GKE team, and was mentioned in #882 -- which is why I selected it here. I have no strong preference.

**Comment by douglas-reid** [comment]
PTAL. This PR now takes a different approach. It uses the distroless bazel rules to pull in and unpack the ca-certs and ssl stuff and add them to the base xenial image.

In this way, Mixer reuses the same base image that we use elsewhere, but adds on the certs.

**Comment by douglas-reid** [comment]
Any objections to submitting this PR at this point in time?

It is blocking some related work which requires HTTPS for outbound calls in adapters. We can revisit the base image choice across Istio in subsequent PRs and issues.

File: docker/extract.sh [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+
+CERTS=$(find usr/share/ca-certificates -type f | sort)
+for cert in $CERTS; do
+  cat $cert >> $CERT_FILE
+done
+
+tar -cf $2 etc/ssl/certs/ca-certificates.crt usr/share/doc/ca-certificates/copyright
+
+rm data.tar.xz
+rm -rf usr/share/ca-certificates
```
xiaolanz: add new line here.

docker/cacerts.bzl
```diff
@@ -0,0 +1,32 @@
...
+            executable = True,
+            allow_files = True,
+        ),
+    },
+    executable = False,
+    outputs = {
+        "out": "%{name}.tar",
+    },
+    implementation = _impl,
+)
```
xiaolanz: add new line

docker/extract.sh
douglas-reid: done.

docker/cacerts.bzl
douglas-reid: done.

PR #891, opened 2017-07-11
Resurrect the attribute based trace context carrier.

ZackButcher:
Getting Envoy to propagate trace metadata in the gRPC metadata sidechannel will likely take a bit of work. Envoy already propagates the trace metadata to us, in the `request.headers` attribute. This PR implements an opentracing `Carrier`, which is opentracing's abstraction over the trace propagation mechanism. Using this, we'll be able to make use of the Zipkin trace headers Envoy propagates to us via attributes.

This is adding back, with slight modification, code that was in Mixer before the shift to our unary API (we thought we'd get away with just using the existing opentracing interceptors, unfortunately not).

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/891)

### DIFF
--- a/pkg/tracing/BUILD
+++ b/pkg/tracing/BUILD
@@ -5,9 +5,11 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "go_default_library",
     srcs = [
+        "attributes.go",
         "basictracing.go",
     ],
     deps = [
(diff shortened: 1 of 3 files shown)


ZackButcher:
Turns out the fix in mixerclient/envoy was much faster than I anticipated: it should be done in the next few days, so I'm gonna delete this and we'll carry on using interceptors.

PR #928, opened 2017-07-20
still broken, want to validate approach

geeknoid:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/928)

### DIFF
--- a/adapter/memQuota/memQuota.go
+++ b/adapter/memQuota/memQuota.go
@@ -73,7 +73,11 @@ func (builder) ValidateConfig(cfg adapter.Config) (ce *adapter.ConfigErrors) {
 	return
 }
 
-func (builder) NewQuotasAspect(env adapter.Env, c adapter.Config, d map[string]*adapter.QuotaDefinition) (adapter.QuotasAspect, error) {
+func (builder) Configure(quotas map[string]*adapter.QuotaDefinition) error {
+	return nil
+}
+
+func (builder) NewQuotasAspect(env adapter.Env, c adapter.Config) (adapter.QuotasAspect, error) {
 	return newAspect(env, c.(*config.Params))
 }
 
--- a/adapter/noop/noop.go
+++ b/adapter/noop/noop.go
@@ -25,8 +25,9 @@ import (
 )
 
 type (
-	builder struct{}
-	aspect  struct{}
+	builder  struct{}
+	qbuilder struct{} // used for quotas, because of duplicate method names w/ diff signatures between metrics and quotas
+	aspect   struct{}
 )
 
 // Register registers the no-op adapter as every aspect.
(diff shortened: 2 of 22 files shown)


adapter/memQuota/memQuota.go
```diff
@@ -73,7 +73,11 @@ func (builder) ValidateConfig(cfg adapter.Config) (ce *adapter.ConfigErrors) {
 	return
 }
 
-func (builder) NewQuotasAspect(env adapter.Env, c adapter.Config, d map[string]*adapter.QuotaDefinition) (adapter.QuotasAspect, error) {
+func (builder) Configure(quotas map[string]*adapter.QuotaDefinition) error {
```
geeknoid: And presumably QuotaDefinition will be replaced with some other type (I forgot the suffix we chose)

adapter/noop/noop.go
```diff
@@ -25,8 +25,9 @@ import (
 )
 
 type (
```
geeknoid: We should be auto-generating this adapter.

@guptasu What do you think?

douglas-reid:
what is this PR for?  can it be broken down into smaller PRs?

adapter/noop/noop.go
guptasu: I think this is just for testing purpose, representing someone writing an adapter that supports a bunch of templates(aspect) that our tests depend on. I think we can leave this as is. BTW, there is a noop2 that is equivalent of this for the template world and I think that has to be hand written because tests depend on it.

adapter/memQuota/memQuota.go
guptasu: Do we need to do this change ? This interface does not represent the eventual Builder interface for the template based adapters. We will end of moving this configure method into Builder.Build method eventually when porting these adapters to the new world. If we can can somehow have this PR not touch any of the existing adapters, it will be ideal. Unless it is really needed.

ZackButcher:
Sorry, this was't supposed to actually be opened. I sent the diff over to Martin to look at the basic shape before I finished it.

No, it can't be readily broken into pieces: I need new interfaces to make the new stuff work with the old, and we can't have the interface changes without updating the things implementing those interfaces. I don't think this PR is overly large compared to our normal ones.

pkg/config/manager.go
```diff
@@ -193,7 +193,7 @@ func (c *Manager) fetchAndNotify() {
 
 	glog.Infof("Loaded new config %s", c.store)
 	for _, cl := range c.cl {
-		cl.ConfigChange(rt, df)
+		cl.ConfigChange(rt, df, handlers)
```
ZackButcher: RT resolves config, which is orthogonal to providing handler instances. I don't think they should be merged ever, really.

PR #938, opened 2017-07-24
Support dispatching to new adapter model Handlers

ZackButcher:
This PR allows new adapter model Handlers to be called by the existing adapterManager dispatch code. It accomplishes this by abstracting over aspect creation (the new `aspect.CreateAspectFunc` type), and in `adapterManager/manager.go` checking whether the resolved config references either a builder (old style) or a handler (new style). It then constructs a `CreateAspectFunc` from the builder or handler.

In order to test the new behavior, I needed an `adapter.Builder` instance implementing many Aspect interfaces; as a result the `adapter/noop.Builder` struct is now public. Due to changing our core Aspect interfaces I wound up having to touch a lot of files; most of the changes are not significant.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/938)

### DIFF
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -492,7 +492,8 @@ func newCacheKey(kind config.Kind, cfg *cpb.Combined) (*cacheKey, error) {
 }
 
 // cacheGet gets an aspect executor from the cache, use adapter.Manager to construct an object in case of a cache miss
-func (m *Manager) cacheGet(cfg *cpb.Combined, mgr aspect.Manager, createAspect aspect.CreateAspectFunc, df descriptor.Finder) (executor aspect.Executor, err error) {
+func (m *Manager) cacheGet(
+	cfg *cpb.Combined, mgr aspect.Manager, createAspect aspect.CreateAspectFunc, df descriptor.Finder) (executor aspect.Executor, err error) {
 	var key *cacheKey
 	if key, err = newCacheKey(mgr.Kind(), cfg); err != nil {
 		return nil, err
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -416,10 +416,10 @@ func (m *Manager) execute(ctx context.Context, cfg *cpb.Combined, requestBag att
 
 	var createAspect aspect.CreateAspectFunc
 	if builder, found := m.builders.FindBuilder(cfg.Builder.Impl); found {
-		var e error
-		createAspect, e = aspect.FromBuilder(builder, kind)
-		if e != nil {
-			return status.WithError(e)
+		var err error
+		createAspect, err = aspect.FromBuilder(builder, kind)
+		if err != nil {
+			return status.WithError(err)
 		}
 	} else if handler, found := m.getHandlers()[cfg.Builder.Impl]; found {
 		createAspect = aspect.FromHandler(handler.Instance)
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -432,7 +432,7 @@ func (m *Manager) execute(ctx context.Context, cfg *cpb.Combined, requestBag att
 
 	executor, err := m.cacheGet(cfg, mgr, createAspect, df)
 	if err != nil {
-		return status.WithError(fmt.Errorf("Failed to construct executor for adapter '%s': %v", cfg.Builder.Impl, err))
+		return status.WithError(fmt.Errorf("failed to construct executor for adapter '%s': %v", cfg.Builder.Impl, err))
 	}
 
 	// TODO: plumb ctx through asp.Execute
--- a/pkg/adapterManager/manager.go
+++ b/pkg/adapterManager/manager.go
@@ -108,8 +108,9 @@ type Manager struct {
 	adapterGP         *pool.GoroutinePool
 
 	// Configs for the aspects that'll be used to serve each API method.
-	cfg atomic.Value
-	df  atomic.Value
+	resolver atomic.Value // config.Resolver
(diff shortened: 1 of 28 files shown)


ZackButcher:
And yet again we have failing linters due to things that didn't change in this PR. I'll send another PR fixing the issues.

pkg/aspect/common.go
```diff
@@ -21,10 +21,98 @@ import (
...
 	"istio.io/mixer/pkg/adapter"
+	"istio.io/mixer/pkg/adapter/config"
 	"istio.io/mixer/pkg/attribute"
+	cfg "istio.io/mixer/pkg/config"
 	"istio.io/mixer/pkg/expr"
 )
 
+// FromHandler creates a CreateAspectFunc from the provided handler instance.
+func FromHandler(handler config.Handler) CreateAspectFunc {
+	return func(adapter.Env, adapter.Config, ...interface{}) (adapter.Aspect, error) {
```
guptasu: This works because the adapter.handler interface and the adapter.Aspect interface are identical right. Once we have things working, we can work on cleaning all of that up.

pkg/aspect/metricsManager.go
```diff
@@ -55,7 +55,7 @@ func newMetricsManager() ReportManager {
 	return &metricsManager{}
 }
 
-func (m *metricsManager) NewReportExecutor(c *cpb.Combined, a adapter.Builder, env adapter.Env, df descriptor.Finder) (ReportExecutor, error) {
+func (m *metricsManager) NewReportExecutor(c *cpb.Combined, createAdapater CreateAspectFunc, env adapter.Env, df descriptor.Finder) (ReportExecutor, error) {
```
guptasu: I always get confused, this is basically "CreateAdapter" right ? Can we rename the CreateAspectFunc to CreateAdapterFunc

pkg/aspect/common_test.go
```diff
@@ -146,3 +149,76 @@ func TestValidateTemplateExpressions(t *testing.T) {
...
+
+func TestFromBuilder(t *testing.T) {
+	tests := []struct {
+		name         string
+		kind         config.Kind
+		expectedType reflect.Type // we only compare the type of this value against the one returned by the builder
+		err          string
+		args         []interface{}
+	}{
+		{"access logs", config.AccessLogsKind, reflect.TypeOf((*adapter.AccessLogsAspect)(nil)).Elem(), "", []interface{}{}},
```
douglas-reid: do we need an invalid kind test (for completeness)?

pkg/aspect/common.go
```diff
@@ -21,10 +21,98 @@ import (
...
+	case cfg.MetricsKind:
+		b, ok := builder.(adapter.MetricsBuilder)
+		if !ok {
+			return nil, fmt.Errorf("invalid builder - kind MetricsKind expected builder implementing MetricsBuilder, got builder: %v", builder)
+		}
+		return func(env adapter.Env, c adapter.Config, cfg ...interface{}) (adapter.Aspect, error) {
+			if len(cfg) < 1 {
+				return nil, fmt.Errorf("metric builders must have configuration args")
+			}
+			metrics, ok := cfg[0].(map[string]*adapter.MetricDefinition)
```
guptasu: Should we assert that the length should be exactly 1, we do indirectly assume by doing cfg[0]. right ?

pkg/aspect/common_test.go
ZackButcher: Ah, now that I'm driving this off of Kind I can add one - I didn't have a test case for it originally because my original `switch` statement was over `builder.(type)`. I'll add one.

pkg/adapterManager/manager.go
```diff
@@ -418,21 +420,29 @@ func (m *Manager) execute(ctx context.Context, cfg *cpb.Combined, requestBag, re
 		return status.WithError(fmt.Errorf("could not find aspect manager %#v", cfg.Aspect.Kind))
 	}
 
-	var builder adapter.Builder
-	if builder, found = m.builders.FindBuilder(cfg.Builder.Impl); !found {
+	var createAspect aspect.CreateAspectFunc
+	if builder, found := m.builders.FindBuilder(cfg.Builder.Impl); found {
+		var e error
+		createAspect, e = aspect.FromBuilder(builder, kind)
```
ZackButcher: Can't do that: e isn't pre-declared, and if we use `:=` it redeclares `createAspect` in a smaller scope.

pkg/aspect/common.go
ZackButcher: Yup

pkg/aspect/common_test.go
ZackButcher: Added tests for the error cases, thanks for catching that, I completely forgot after refactoring that func.

pkg/aspect/metricsManager.go
ZackButcher: Adapters implement Aspects, Executors execute Aspects (i.e. Executors care about behaviors, not implementations). So I think CreateAspectFunc is appropriate here.

pkg/aspect/common.go
ZackButcher: Done

ZackButcher:
jenkins run mixer/e2e-smoketest

ZackButcher:
Tests passed but Jenkins failed. Running again.

ZackButcher:
jenkins run mixer/e2e-smoketest

guptasu:
@ZackButcher I think this PR is set to go in (approved and pass all tests).

ZackButcher:
There has not yet been a passing jenkins run of mixer/e2e-smoketest, even though on multiple runs all tests in that suite have passed.

PR #939, opened 2017-07-24
Fix linter issues from PR 938

ZackButcher:
Fixes linter issues in unchanged files from https://github.com/istio/mixer/pull/938

We really need to figure out what is happening with our linters. It's getting ridiculous - every PR fails due to linter issues in code the PR didn't touch, and I have never been able to repro any of the failures locally.

This PR may need a few iterations since I can't repro linter failures locally.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/939)

### DIFF
--- a/pkg/config/apiserver.go
+++ b/pkg/config/apiserver.go
@@ -251,7 +251,7 @@ func (a *API) Run() {
 
 // getScopes returns the scopes
 // "/scopes"
-func (a *API) getScopes(req *restful.Request, resp *restful.Response) {
+func (a *API) getScopes(_ *restful.Request, resp *restful.Response) {
 	writeErrorResponse(http.StatusNotImplemented, "Listing scopes not implemented", resp)
 }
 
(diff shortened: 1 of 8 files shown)


douglas-reid:
can we run the updated linters on the entire codebase and fix all the issues at once somehow? we shouldn't be incrementally dealing with these issues, right?

pkg/config/apiserver.go
```diff
@@ -251,7 +251,7 @@ func (a *API) Run() {
 
 // getScopes returns the scopes
 // "/scopes"
-func (a *API) getScopes(req *restful.Request, resp *restful.Response) {
+func (a *API) getScopes(_ *restful.Request, resp *restful.Response) {
```
geeknoid: Why is this using _ while the other cases above just removed the parameter name completely?

ZackButcher: If any parameter is named, all parameters must be named - you can't mix and match. Since we use `resp` in the body, the first parameter must have a name. In the other cases they were fakes used in tests that ignored their parameters, so we could remove the names of all of them.

PR #941, opened 2017-07-25
Fix the build on OSX

ZackButcher:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/941)


ZackButcher:
IDC either way - this isn't a super common problem, but I also don't really see any downside. Is there a global way to set the default?

PR #953, opened 2017-07-27
Make explicit some of the external deps

douglas-reid:
This PR adds some of the deps that get include through the `rules_protobuf` dependency and rules explicitly to the Mixer `WORKSPACE` file.

There was some concern expressed that Mixer itself is not in control of all the deps that it is using. This attempts to make explicit the deps and give Mixer control over their handling (and upgrade lifecycle). This comes with some potential downsides (upgrading Mixer's versions could cause issues in the dependent repos and rules that are used to build Mixer; we are copying deps, but can't ensure that we are fully covering all imported deps; etc.).

A few other notes:
- Ideally, this work should allow us to remove calls like `proto_repositories()`. Unfortunately, those rules do more than just install deps and cannot be easily avoided. This results in duplicate `git_repository` declarations ultimately.
- I did not attempt to complete the transition of the `cpp_proto_repositories()` rule, as one of the deps pulled in via that rule uses a custom BUILD script, and that seemed a bit excessive to me -- and I'm completely uncertain as to the reason for a cpp BUILD rule to be in the Mixer `WORKSPACE`.
- I moved the new deps above the corresponding rules (`proto_repositories()`, etc.) that also include them. I believe that bazel uses the first declared dep in a WORKSPACE, so this should force usage of the Mixer declared dep.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/953)


douglas-reid:
build mixer/e2e-smoketest

douglas-reid:
build mixer/e2e-smoketest

PR #955, opened 2017-07-27
Remove problematic alternate target service derivation

douglas-reid:
This PR removes the problematic code in the k8s adapter related to `targetService` derivation based on inputs other than `target.uid`.

The k8s adapter was previously taking an input named `targetService` and attempting to derive the FQDN for the target service based on this input. At best, operator config would supply the attribute `target.service` for that input, rendering the derivation useless. At worst, the operator would use a non-namespace-aware input, such as the attribute `request.host`, leading to overriding the derived FQDN with a service name in the `default` namespace (leading to incorrect application of config by Mixer).

We expect `target.uid` to always be available (and likely to have `target.service` as well). This eliminates a hard-to-debug surprise derivation.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/955)

### DIFF
--- a/adapter/kubernetes/config/config.proto
+++ b/adapter/kubernetes/config/config.proto
@@ -40,6 +40,7 @@ option (gogoproto.gostring_all) = false;
 // sourceUID and the output value of pod ip, this adapter will output a map
 // that includes a key of "sourcePodIP" (assuming parameter defaults).
 message Params {
+    reserved 17;
     // next field id: 19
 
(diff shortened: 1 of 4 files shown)


kyessenov:
Hah, smoketest failed! Currently, only `target.uid` is available, but `target.service` is 100% guaranteed to be available, will add it at some point.

douglas-reid:
build mixer/e2e-smoketest

douglas-reid:
build mixer/e2e-smoketest

PR #961, opened 2017-07-27
Introduce the ProtoBag type.

geeknoid:
ProtoBag is an attribute bag whose storage is an Attributes proto.
Use of ProtoBag elides much setup work that Mixer was doing for
every incoming request.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/961)

### DIFF
--- a/pkg/api/perf_test.go
+++ b/pkg/api/perf_test.go
@@ -130,7 +130,7 @@ func (bs *benchState) cleanupBenchState() {
 	bs.deleteGRPCServer()
 }
 
-func (bs *benchState) Check(ctx context.Context, bag *attribute.MutableBag, output *attribute.MutableBag) rpc.Status {
+func (bs *benchState) Check(ctx context.Context, bag attribute.Bag, output *attribute.MutableBag) rpc.Status {
 	output.Set("kubernetes.pod", "a pod name")
 	output.Set("kubernetes.service", "a service name")
 	output.Set("kubernetes.id", int64(123456))
@@ -138,18 +138,18 @@ func (bs *benchState) Check(ctx context.Context, bag *attribute.MutableBag, outp
 	return status.OK
 }
 
-func (bs *benchState) Report(ctx context.Context, bag *attribute.MutableBag) rpc.Status {
+func (bs *benchState) Report(ctx context.Context, bag attribute.Bag) rpc.Status {
 	return status.WithPermissionDenied("Not Implementd")
 }
 
-func (bs *benchState) Quota(ctx context.Context, requestBag *attribute.MutableBag,
+func (bs *benchState) Quota(ctx context.Context, requestBag attribute.Bag,
 	qma *aspect.QuotaMethodArgs) (*aspect.QuotaMethodResp, rpc.Status) {
 
 	qmr := &aspect.QuotaMethodResp{Amount: 42}
 	return qmr, status.OK
 }
 
-func (bs *benchState) Preprocess(ctx context.Context, requestBag *attribute.MutableBag, responseBag *attribute.MutableBag) rpc.Status {
+func (bs *benchState) Preprocess(ctx context.Context, requestBag attribute.Bag, responseBag *attribute.MutableBag) rpc.Status {
 	return status.OK
 }
 
--- a/pkg/attribute/protoBag.go
+++ b/pkg/attribute/protoBag.go
@@ -1,0 +1,264 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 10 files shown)


pkg/attribute/protoBag.go
```diff
@@ -0,0 +1,159 @@
...
+	return &ProtoBag{
+		proto:          proto,
+		globalDict:     globalDict,
+		globalWordList: globalWordList,
+	}
+}
+
+// Get returns an attribute value.
+func (pb *ProtoBag) Get(name string) (interface{}, bool) {
+	index, ok := pb.globalDict[name]
```
ZackButcher: We might want to look in to a map from index to type, since it's a known mapping for the global dictionary. That way you only do three map lookups: name -> index, index -> type, type[index] -> value. Otherwise in the worst case we have to do 9 lookups. In particular, `request.headers` is frequently used and the last that are hit walking through this method.

pkg/api/perf_test.go
```diff
@@ -130,26 +130,26 @@ func (bs *benchState) cleanupBenchState() {
 	bs.deleteGRPCServer()
 }
 
-func (bs *benchState) Check(ctx context.Context, bag *attribute.MutableBag, output *attribute.MutableBag) rpc.Status {
+func (bs *benchState) Check(ctx context.Context, bag attribute.Bag, output *attribute.MutableBag) rpc.Status {
```
guptasu: Specific reason to not use pointer ?

pkg/attribute/protoBag.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return names
+}
+
+func (pb *ProtoBag) Done() {
+	// NOP
+}
+
+// Lazily produce the message-level dictionary
+func (pb *ProtoBag) getMessageDict() map[string]int32 {
```
geeknoid: getMessageDict is called lazily. If no attributes are used that require the message-level dictionary, than getMessageDict never runs.

pkg/attribute/protoBag.go
```diff
@@ -0,0 +1,159 @@
...
+	if pb.messageDict == nil {
+		// build the message-level dictionary
+
+		d := make(map[string]int32, len(pb.proto.Words))
+		for i, name := range pb.proto.Words {
+			d[name] = -int32(i) - 1
```
geeknoid: Done.

pkg/api/perf_test.go
geeknoid: MutableBag is a struct, whereas Bag is an interface. Don't want pointer to an interface in general in Go.

pkg/attribute/protoBag.go
```diff
@@ -0,0 +1,159 @@
...
+	if ok {
+		return value, true
+	}
+
+	return nil, false
+}
+
+// Names return the names of all the attributes known to this bag.
+func (pb *ProtoBag) Names() []string {
+	var names []string
```
geeknoid: Nope, this works as is.

pkg/attribute/protoBag.go
geeknoid: Excellent idea, if my plan A doesn't work.

My plan A is to bring back strongly-typed getters (GetInt64, GetString, etc) and have the expression parser use those. Since it knows the logical type from the DescriptorFinder, it can derive the transport-level type and invoke the right getter.

In the meantime, I will reorg the order of evaluation and move StringMap up.

costinm: +1 on strongly-typed getters - avoids the interface{} and the repeated maps. 
Another option is to change the proto to send a single map of int to bytes, and convert on server
side ( only when attribute is used - avoiding client-side conversions for unused attrs.)

geeknoid:
The use of dictionaries substantially reduces the payload size of all
messages and my prior perf tests show it has a fairly positive impact on
overall perf.

Once we have real-world traffic data for the Envoy->Mixer communication
channel, we can evaluate further optimizations to the protocol. It's quite
possible that we could communicate at a slightly lower level and avoid a
lot of canonicalization overhead.

costinm:
Payload reduction has a positive impact on perf on lowbandwith/high-latency networks, 
as proven by H2. However I've seen benchmarks showing a negative impact on high-BW/low-latency,
as is typical in a cluster.

So far all tests I've seen including envoy - with all the extra allocations and copies - seem to show 
the compression has a significant cost, and may be larger than savings on mixer side. It is visible
in particular on high load, where alloc/free and copy end up dominating. 

One middle ground option would be to use the fixed H2 built-in dictionary, to still compress all
headers ( which probably can be optimized-away - since Envoy will need to do this before
sending over http2). 

Alternatively, we could have it both ways - by adding a map<string,string> rawHeaders, and
have mixer process it on demand ( so we can easily compare side-by-side, and let the
client chose to do compression or not).

geeknoid:
The perf tests I conducted where single-machine tests. So infinite
bandwidth and negligible latency. I'm pretty sure this is a net win.

On Fri, Jul 28, 2017 at 11:35 AM, Costin Manolache <notifications@github.com

pkg/attribute/protoBag.go
guptasu: Make sense. Get(s) method is not always going to get called.

PR #962, opened 2017-07-27
Introduce gazelle

jmuk:
This also regenerates the BUILD files through gazelle.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/962)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -178,7 +178,7 @@ new_go_repository(
 )
 
 new_go_repository(
-    name = "com_github_opentracing_basictracer",
+    name = "com_github_opentracing_basictracer_go",
     commit = "1b32af207119a14b1b231d451df3ed04a72efebf",  # Sep 29, 2016 (no releases)
     importpath = "github.com/opentracing/basictracer-go",
 )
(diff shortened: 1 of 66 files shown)


ZackButcher:
It's a hygiene issue: it's better to have build targets that are not accessible outside of their intended scope than to have all targets be globally visible.

Plus, the specific instances I called out are all proto build targets, so we don't have a choice w.r.t. what the golang visibility looks like.

ayj:
[Internal packages](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit) may be appropriate in this case, e.g. 

```
adapter/<name>/internal/config/    # only usable by adapter/<name>
adapter/<name>/
```

ZackButcher:
Oh sorry, I misread: gazelle did not touch the targets under `adapter/*/config/BUILD`. Some of the subtargets in `cmd/` and `example/` maybe shouldn't be public but not a big deal.

douglas-reid:
should we hold off on this until we upgrade to the latest `rules_go` (so that we only pay the cost once)?  or does it not matter?

@ayj but but... that's why we have `bazel`.

also, i don't know why but I strongly dislike `go_default_tests` for the test targets, but such is life.

kyessenov:
latest rules_go don't build client-go. waiting for a fix.

jmuk:
re hold off until `rules_go` upgrade: sure. I just did this thing today just because I got frustrated by some deps management.

re `go_default_tests`: unfortunately gazelle doesn't support customizing the default test name 😢. Maybe, `bin/gazelle` can add another `sed` line to rename them if that's the agreement on mixer hackers.

douglas-reid:
ok. let's do this now. i'll deal with the 0.5.2 conversion when `rules_go` gets sorted out properly in a week or so.

PR #972, opened 2017-07-31
Update Makefile to force commands through bash

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/972)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 
+SHELL := /bin/bash
+
 build:
 	@bazel build //...:all
 
@@ -32,3 +34,5 @@ coverage:
 
 racetest:
 	@bazel test --features=race //...
+
+.PHONY: build clean test lint fmt coverage racetest


Makefile
```diff
@@ -1,4 +1,4 @@
-## Copyright 2016 Istio Authors
+## Copyright 2017 Istio Authors
```
ldemailly: no need to change that

Makefile
```diff
@@ -12,23 +12,32 @@
...
+
+.PHONY: build
 build:
-	@bazel build ...:all
+	@bazel build //...
 
+.PHONY: clean
 clean:
-	@bazel clean
+	@bazel clean //...
```
ldemailly: clean does not need an argument

Makefile
```diff
@@ -12,23 +12,32 @@
...
 
+.PHONY: fmt
 fmt:
 	@bin/fmt.sh
 
+.PHONY: coverage
 coverage:
 	@bin/codecov.sh
 
+.PHONY: racetest
```
ldemailly: maybe put all the .PHONY in 1 line

Makefile
```diff
@@ -12,23 +12,32 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 
+SHELL := /bin/bash
+
+.PHONY: build
 build:
-	@bazel build ...:all
```
ldemailly: is ...:all exactly the same as ... ?

ZackButcher: They're equivalent but some shells (e.g. zsh) barf on `...` but don't barf on `//...`

Makefile
cmluciano: Do you mean something like ```.PHONY: build test lint``` ? If so this would have the side-effect of running test and lint whenever you call ```make build```

Makefile
cmluciano: I can remove this. Do we have requirements on when/if this date should ever be changed?

ldemailly: a copyright doesn't go forward; it can be a range like 2016-present but usually the first date matters

Makefile
ldemailly: not if you don't put it as the first target e.g. try:
```
foo:
	echo foo

bar:
	echo bar

blah:
	echo blah

.PHONY: foo bar blah
```

cmluciano: nice, trying now

cmluciano:
ok I think I got it now :)

cmluciano:
@ldemailly @ZackButcher Is this good to go?

ZackButcher:
I think so, just needs to be updated to latest HEAD and it can be merged in.

cmluciano:
rebased

ldemailly:
@cmluciano github still says your branch is out of date

cmluciano:
*fingers crossed*

Issue #966, opened 2017-07-28
Orphaned adapter cleanup on config change


mandarjog:
Mixer maintains a cache of initialized adapters. When configuration changes, an adapter instance may become orphaned. 

Orphaned adapters should `close()` and be evicted from the cache.

guptasu:
@mandarjog in the new model controller is already doing that. We good with this bug

PR #963, opened 2017-07-27
Update to more recent k8s api, matching pilot version

costinm:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/963)

PR #975, opened 2017-07-31
update kubernetes client library package

jmuk:
This is similar to https://github.com/istio/pilot/pull/966, updating k8s.io/client-go and its dependencies. adapters/kubernetes also needs to update some module name due to this update.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/975)


douglas-reid:
How does this compare to #963  ?

jmuk:
oh, didn't notice that. Closing.

adapter/kubernetes/cache.go
```diff
@@ -20,10 +20,11 @@ import (
 	"reflect"
 	"time"
 
+	"k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
douglas-reid: can we just use `metav1` instead (no underscore)?

jmuk:
reopened: #963 is referring to a bit old version of client-go. This refers to the same revision as istio/pilot#966

adapter/kubernetes/cache.go
jmuk: done

PR #985, opened 2017-08-01
Track which attributes are used during request processing.

geeknoid:
- We track which attributes are referenced for each Mixer Check call and report this to Envoy. Envoy uses this information to properly cache the Check response.

- Fixes a bug in the proto bag code which didn't correctly handle the case where a word was both in the global and the message-level word lists. The message-level word list now takes precedence in all cases.

- Adds the needed support to enable upgrades of the global word list by limiting used of the global word list in generated Attributes messages to what the caller indicated is its maximum support word count.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/985)

### DIFF
--- a/pkg/attribute/bag_test.go
+++ b/pkg/attribute/bag_test.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -171,10 +172,10 @@ func TestEmptyRoundTrip(t *testing.T) {
 	attrs0 := mixerpb.Attributes{}
 	attrs1 := mixerpb.Attributes{}
 	mb := GetMutableBag(nil)
-	mb.ToProto(&attrs1, nil)
+	mb.ToProto(&attrs1, nil, 0)
 
 	if !reflect.DeepEqual(attrs0, attrs1) {
-		t.Error("Expecting equal attributes, got a delta")
+		t.Errorf("Expecting equal attributes, got a delta: original #%v, new #%v", attrs0, attrs1)
 	}
 }
 
@@ -267,7 +268,7 @@ func TestProtoBag(t *testing.T) {
 			}
 
 			var a2 mixerpb.Attributes
-			mb.ToProto(&a2, globalDict)
+			mb.ToProto(&a2, globalDict, len(globalDict))
 
 			pb.Done()
 		}
@@ -288,7 +289,7 @@ func TestMessageDict(t *testing.T) {
 	b.Set("M3", "M2")
 
 	var attrs mixerpb.Attributes
-	b.ToProto(&attrs, globalDict)
+	b.ToProto(&attrs, globalDict, len(globalDict))
 	b2, _ := GetBagFromProto(&attrs, globalWordList)
 
 	if !compareBags(b, b2) {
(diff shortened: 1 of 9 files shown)


pkg/attribute/protoBag.go
```diff
@@ -50,9 +64,56 @@ func (pb *ProtoBag) Get(name string) (interface{}, bool) {
...
+func (pb *ProtoBag) GetReferencedAttributes() mixerpb.ReferencedAttributes {
+	output := mixerpb.ReferencedAttributes{}
+
+	ds := newDictState(nil)
+
+	output.AttributeMatches = make([]mixerpb.ReferencedAttributes_AttributeMatch, len(pb.referencedAttrs))
+	i := 0
+	for k, v := range pb.referencedAttrs {
+		output.AttributeMatches[i] = mixerpb.ReferencedAttributes_AttributeMatch{
+			Name:      ds.getIndex(k),
```
ZackButcher: Oh wow. It took me a while to understand what this is doing. I'm not a fan of `getIndex` mutating the underlying store, I think that kind of mutation might bite us down the road. In Java I'd call this method `getGlobalIndexOrCreateMessageIndex`. Maybe change this to `synthesizeIndex` or `getOrCreateIndex`, or similar, to denote that this is _not_ just a lookup in the dicts that the dictState was created with.

pkg/attribute/bag_test.go
```diff
@@ -567,6 +568,119 @@ func TestMessageDictEdge(t *testing.T) {
...
+				var err error
+				b, err = GetBagFromProto(&attrs, globalWordList)
+				if err != nil {
+					t.Error("Expecting success, got %v", err)
+				}
+			}
+
+			v, ok := b.Get("HELLO")
+			if !ok {
+				t.Error("Expecting true, got false")
```
douglas-reid: nit: error messages should be `got`, then `want`

pkg/attribute/bag_test.go
```diff
@@ -567,6 +568,119 @@ func TestMessageDictEdge(t *testing.T) {
...
+
+func TestReferenceTracking(t *testing.T) {
+	globalWordList := []string{"G0", "G1", "G2"}
+	globalDict := map[string]int32{globalWordList[0]: 0, globalWordList[1]: 1, globalWordList[2]: 2}
+	messageWordList := []string{"N1", "N2", "N3"}
+
+	attrs := mixerpb.Attributes{
+		Words: messageWordList,
+		Strings: map[int32]int32{
+			0:  0,
```
douglas-reid: can we add comments about the intended affect of these maps? I find it hard to follow.

pkg/attribute/dictState.go
```diff
@@ -0,0 +1,55 @@
...
+		return index
+	}
+
+	if ds.messageDict == nil {
+		ds.messageDict = make(map[string]int32)
+	} else if index, ok := ds.messageDict[word]; ok {
+		return index
+	}
+
+	index := -int32(len(ds.messageDict)) - 1
```
douglas-reid: i feel like we need some helper method around index calculation. I think coming back to this later will leave us scratching our heads.

pkg/attribute/protoBag.go
```diff
@@ -50,9 +64,56 @@ func (pb *ProtoBag) Get(name string) (interface{}, bool) {
...
+	return result, ok
+}
+
+// Get the set of attributes that have been referenced through this bag.
+func (pb *ProtoBag) GetReferencedAttributes() mixerpb.ReferencedAttributes {
+	output := mixerpb.ReferencedAttributes{}
+
+	ds := newDictState(nil)
+
+	output.AttributeMatches = make([]mixerpb.ReferencedAttributes_AttributeMatch, len(pb.referencedAttrs))
```
douglas-reid: any reason not use use `make([]..., 0, len(pb.referencedAttrs))` and `append`s (instead of `i`) ?

pkg/attribute/protoBag.go
```diff
@@ -33,14 +33,28 @@ type ProtoBag struct {
 	messageDict         map[string]int32
 	convertedStringMaps map[int32]map[string]string
 	stringMapMutex      sync.RWMutex
+
+	// to keep track of attributes that are referenced
+	referencedAttrs      map[string]mixerpb.ReferencedAttributes_Condition
```
douglas-reid: candidate for `syncmap`? https://godoc.org/golang.org/x/sync/syncmap

pkg/attribute/bag_test.go
```diff
@@ -567,6 +568,119 @@ func TestMessageDictEdge(t *testing.T) {
...
+	globalWordList := []string{"HELLO"}
+	globalDict := map[string]int32{globalWordList[0]: 0}
+	messageWordList := []string{"HELLO", "GOOD", "BAD"}
+
+	attrs := mixerpb.Attributes{
+		Words:   messageWordList,
+		Strings: map[int32]int32{0: -3, -1: -2},
+	}
+
+	for i := 0; i < 2; i++ {
```
geeknoid: Cleaned up and commented, hopefully clearer.

pkg/attribute/bag_test.go
```diff
@@ -567,6 +568,119 @@ func TestMessageDictEdge(t *testing.T) {
...
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			var b Bag
+
+			if i == 0 {
+				b = NewProtoBag(&attrs, globalDict, globalWordList)
+			} else {
+				var err error
+				b, err = GetBagFromProto(&attrs, globalWordList)
+				if err != nil {
+					t.Error("Expecting success, got %v", err)
```
geeknoid: Done

pkg/attribute/dictState.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func newDictState(globalDict map[string]int32) *dictState {
+	return &dictState{
+		globalDict: globalDict,
+	}
+}
+
+func (ds *dictState) getIndex(word string) int32 {
+	if index, ok := ds.globalDict[word]; ok {
```
geeknoid: Nope, this is going in the opposite direction (producing an Attributes message) and there we want to aggressively use the global dictionary when possible.

pkg/attribute/dictState.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	index := -int32(len(ds.messageDict)) - 1
+	ds.messageDict[word] = index
+	return index
+}
+
+func (ds *dictState) getMessageWordList() []string {
+	if len(ds.messageDict) == 0 {
+		return nil
```
geeknoid: I want nil here such that it matches a blank Attributes message.

pkg/attribute/protoBag.go
geeknoid: Renamed to assignIndex. Added big comment too.

pkg/attribute/bag_test.go
geeknoid: Done

pkg/attribute/bag_test.go
geeknoid: Done. Added several test comments.

pkg/attribute/dictState.go
geeknoid: Done.

pkg/attribute/protoBag.go
geeknoid: Eventually. For now, syncmap doesn't support a Len method which is really a bummer.

geeknoid:
PTAL

pkg/attribute/dictState.go
```diff
@@ -0,0 +1,82 @@
...
+		globalDict:      globalDict,
+		globalWordCount: globalWordCount,
+	}
+}
+
+// assignDictIndex determines whether a word is in the allowed subset of the
+// global word list or whether it should be added to the message word list
+// being accumulated. It then returns the appropriate dictionary index given
+// the selected word list.
+func (ds *dictState) assignDictIndex(word string) int32 {
```
ZackButcher: Much clearer, thanks.

Issue #983, opened 2017-08-01
Use k8s CRD as the config store


jmuk:
This is a task related to our recent decision on Galley; first we will use kubernetes custom resource definitions (CRDs) to store configs for istio. However, this new config data model is different from what mixer currently has, and it is not obvious how we make this transition.

I think we should take the following steps:

step 1: implement k8s client as a "store" implementation
----

The initial attempt would be to build the k8s client as an implementation of `pkg/config/store` (especially a `Watcher`). This will be inefficient because it needs to reformat the k8s data into the serialized yaml data and then re-validate them. No external webhooks.

step 2: introduce the validation webhook / reformat apiserver
----

Currently `pkg/config/apiserver.go` receives the requests from `istioctl`, validates the config data, and stores it to the server.

We want to replace the validation by the external admission webhook. During the transition, first webhooks are introduced. Also apiserver.go will be reformed to store the config data directly into CRD and let webhooks do the rest.

step 3: refactor validation / storage abstraction
----

Related to https://github.com/istio/mixer/issues/910. k8s client can emit updates per "kind" (i.e. type of configs), it is better to refactor the validator and storage abstraction to fit with the new config data model.

step 4: the retirement of the old config data
----

Updates on e2e test scenario, installation scripts for the predefined data (to `istio/istio`), and removal of the old code.

----

This issue is for tracking tasks for myself. Here's the notes:
- [ ] step1: k8s client as a store implementation
- step2
   - [ ] a: validation webhooks
   - [ ] b: reformat apiservers
- [ ] step3: refactoring of validator and storage
- step4
   - [ ] a: e2e update
   - [ ] b: installation data and scripts
   - [ ] c: removal of old code

geeknoid:
Can we close this?

jmuk:
yes.

PR #989, opened 2017-08-02
implement Kubernetes CRDs watcher as pkg/config/store interface.

jmuk:
This is an inefficient version to consume k8s CRDs as the config
data -- would be used for the step 1 of #983

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/989)


pkg/config/crd/resources.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, def := range defs.(*unstructured.UnstructuredList).Items {
+		name := def.Object["metadata"].(map[string]interface{})["name"].(string)
+		delete(crds, name)
+	}
+	if len(crds) == 0 {
+		return nil
+	}
+	for _, crd := range crds {
+		glog.V(7).Infof("Declaring %+v", crd)
+		_, err := client.Create(crd)
```
mandarjog: We either shouldn't declare them in this path normally.
If we want to create it, we should not `fail` if the crd definition already exists.

jmuk: fixed to pass when the resource definition already existed.

Also removed the code path to use this code from `crdwatcher.go`, it won't declare it anymore. Maybe I can remove this file entirely.

jmuk:
PTAL

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+type watcher struct {
+	namespace string
+	stoppers  []func()
+
+	mu       sync.Mutex
+	data     map[string]string
+	listener store.Listener
+
+	// kind to resourceMap
+	resources map[string]resourceMap
```
ZackButcher: nit: move this field up in the struct so it's obvious that `resources` is not guarded by `mu`.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+
+// newStore creates a new watcher instance with watching resources.
+func newStore(u *url.URL) (store.KeyValueStore, error) {
+	namespace := u.Host
+	if namespace == "" {
+		namespace = v1.NamespaceDefault
+	}
+	configFilePath := u.Path
+	if configFilePath == "" {
+		configFilePath = filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
ZackButcher: This should be config driven. Not a blocker for this PR, but the value should probably originate from a flag.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// Mapping from the resource name to the resource spec.
+type resourceMap map[string]*resource
+
+type watcher struct {
+	namespace string
+	stoppers  []func()
+
+	mu       sync.Mutex
```
ZackButcher: Please comment which fields this mutex guards (just `data`? `data` and `listener`?).

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// Set implements a KeyValueStore method.
+func (w *watcher) Set(key, value string) (int, error) {
+	return store.IndexNotSupported, errors.New("not implemented")
+}
+
+// List implements a KeyValueStore method.
+func (w *watcher) List(key string, recurse bool) ([]string, int, error) {
+	w.mu.Lock()
```
ZackButcher: Aquire the lock after fixing the key, on line 119.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+	res := make([]interface{}, 0, len(data))
+	for _, d := range data {
+		res = append(res, d.spec)
+	}
+	enc, err := yaml.Marshal(map[string]interface{}{field: res})
+	if err != nil {
+		glog.Errorf("Failed to marshal: %v", err)
+		return
+	}
+	w.data[key] = string(enc)
```
ZackButcher: Don't you need to hold `w.mu` to make this change? If not (e.g. because this func must only be called when the lock is already held) then comment the method to that effect.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func unstructuredToResource(uns *unstructured.Unstructured) (string, *resource) {
+	md := uns.Object["metadata"].(map[string]interface{})
+	name := md["name"].(string)
+	version := md["resourceVersion"].(string)
+	return name, &resource{version, uns.Object["spec"]}
+}
+
+func (w *watcher) appendAndEncode(data resourceMap, field, key string) {
```
ZackButcher: Please rename `data` to something else - maybe `resources`. It's confusing in context compared to `w.data`, especially because access to `w.data` needs to be guarded.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+		if err != nil {
+			glog.Errorf("Failed to marshal: %v", err)
+			return
+		}
+		w.data[key] = string(enc)
+	}
+}
+
+func (w *watcher) buildDataAndNotify() {
+	w.mu.Lock()
```
ZackButcher: If `resources` is not guarded by `mu` then this should be re-written - you could hold `mu` for a substantially shorter time period (just the updates at the end of `appendAndEncode`, `encodeData`) while still making the update look atomic because you don't notify listeners until construction is finished.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+	defer w.mu.Unlock()
```
ZackButcher: Do you need to be holding `mu` while notifying everyone? Is that done to make things look atomic? I worry because `NotifyStoreChanged` is executed in thread, so you're holding the lock until every single listener finishes executing whatever arbitrary code that they have.

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+	dynamic, err := dynamic.NewClient(config)
+	if err != nil {
+		return err
+	}
+
+	list, err := discovery.ServerResourcesForGroupVersion("config.istio.io/v1alpha2")
+	if err != nil {
+		return err
+	}
+	for _, k := range []string{"Adapter", "Descriptor", "Constructor", "Handler", "Rule", "Action"} {
```
ZackButcher: Promote this list up to a var at the top of the file - maybe also declare vars for each of the strings since they're used in at least one other place as well (`buildDataAndNotify`).

pkg/config/crd/crdwatcher.go
```diff
@@ -0,0 +1,273 @@
...
+// Mapping from the resource name to the resource spec.
+type resourceMap map[string]*resource
+
+type watcher struct {
+	namespace string
+	stoppers  []func()
+
+	mu       sync.Mutex
+	data     map[string]string
+	listener store.Listener
```
ZackButcher: This should be `[]store.Listener`, right? As implemented if I call `RegisterListener` twice I only get notified on the last listener.

jmuk:
We will take a different approach with #1042 and another PR. Closing.

PR #1014, opened 2017-08-04
Introduce and integrate IL based evaluator.

ozevren:
Also introduce a command-line flag to enable the IL based evaluator, optionally.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1014)


pkg/il/evaluator/evaluator.go
```diff
@@ -0,0 +1,158 @@
...
+func (e *ILEvaluator) EvalPredicate(expr string, attrs attribute.Bag) (bool, error) {
+	var result interpreter.Result
+	var err error
+	if result, err = e.evalResult(expr, attrs); err != nil {
+		return false, err
+	}
+	return result.Bool(), nil
+}
+
+// EvalType evaluates expr using the attr attribute bag and returns the result as string.
```
geeknoid: Doesn't return a string, right?

ozevren: Oops. Copy paste mistake. Fixed.

cmd/server/cmd/server.go
```diff
@@ -308,6 +323,9 @@ func runServer(sa *serverArgs, tmplRepo template.Repository, printf, fatalf shar
 	grpcOptions = append(grpcOptions, grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(interceptors...)))
 
 	configManager.Register(adapterMgr)
+	if sa.useIL {
+		configManager.Register(ilEval)
```
ZackButcher: I really wish there was a convenient way for us to get rid of this circular dependency (eval used to construct the config manager which the eval needs to listen to). Nothing springs to mind though.

pkg/il/compiler/compiler.go
```diff
@@ -31,18 +31,24 @@ type generator struct {
 	err     error
 }
 
+// CompilationResult is returned as the result of compilation.
+type CompilationResult struct {
```
ZackButcher: Just `Result`, you're already in `package compiler`, this is references as `compiler.CompilationResult` which is redundant; `compiler.Result` works just as well/more idiomatic. Plus the variable name that you store this into is always called `result`, which is another signal that the type can by named `Result` w/o losing anything.

pkg/il/evaluator/evaluator.go
```diff
@@ -0,0 +1,158 @@
...
+	"istio.io/mixer/pkg/config"
+	"istio.io/mixer/pkg/config/descriptor"
+	"istio.io/mixer/pkg/expr"
+	"istio.io/mixer/pkg/il/compiler"
+	"istio.io/mixer/pkg/il/interpreter"
+)
+
+// ILEvaluator is an implementation of expr.Evaluator that also exposes specific methods.
+// Specifically, it can listen to config change events.
+type ILEvaluator struct {
```
ZackButcher: We're in package `evaluator`, naming this struct `IL` doesn't lose anything since you refer to it as `evaluator.IL` (vs the redundant `evaluator.ILEvaluator`).

pkg/il/evaluator/evaluator.go
```diff
@@ -0,0 +1,158 @@
...
+	cache  *lru.Cache
+	finder expr.AttributeDescriptorFinder
+	fMap   map[string]expr.FuncBase
+}
+
+var _ expr.Evaluator = (*ILEvaluator)(nil)
```
ZackButcher: `&ILEvaluator{}` rather than `(*ILEvaluator)(nil)`

cmd/server/cmd/server.go
ozevren: Why does config manager need eval? I think it would be nice to tease out these into their own layers as appropriate. the Evaluator interface seems to be highly overloaded.
(I won't take any action for this CL, but may look into it in the future).

ZackButcher: Yeah, I didn't expect you to take action. I don't recall why it needs eval, but IIRC it's to look at selectors? Unclear, and seems unnecessary. We should look into removing that dependency.

pkg/il/evaluator/evaluator.go
ozevren: Ok, applied pretty much all the feedback. Thanks

PR #1016, opened 2017-08-04
fix the warning message on pre-commit check.

jmuk:
bin/use_bazel_go.sh tries to set GOROOT, however the dependency
path name has been changed -- now it doesn't start with golang_.
It's go1_8_3_linux_amd64 on my environment.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1016)


sebastienvas:
I already fixed this when moving to Prow.

PR #1019, opened 2017-08-04
Reverse the polarity of the useIL flag.

ozevren:
+ Return nil instead of panic'ing for empty Result instances.
+ Also rename .String() etc to .AsString() to avoid implicit .String() invocations.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1019)


ozevren:
Huh, I have no idea how I "approved" this PR. On the contrary, DO NOT APPROVE THIS PR!

yutongz:
Sorry for the inconvenience, we are using a new system of reviewing and merging. Please take a look at [mungegithub](https://github.com/istio/test-infra/blob/master/mungegithub/deployment/README.md) for more info.

PR #1021, opened 2017-08-04
Refactor stuff in order to pass an adapter.Env to the HandlerBuilder.…

geeknoid:
…Build method.

For now, the env always comes in as nil. I'll create the actual Env to pass down in a
subsequent PR. I just want to get all this refactoring stuff out of the way first.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1021)


adapter/noop2/noop2.go
```diff
@@ -32,8 +31,8 @@ type (
 
 ///////////////// Configuration time Methods ///////////////
 
-func (noop2Builder) Build(cnfg proto.Message) (adapter_cnfg.Handler, error) {
-	fmt.Println("Build in noop Adapter called with", cnfg)
+func (noop2Builder) Build(config proto.Message, _ *adapter.Env) (adapter.Handler, error) {
```
ZackButcher: `adapter.Env` is an interface, this shouldn't be passed as a pointer.

ZackButcher: (Here and in all the other interfaces in the PR)

geeknoid:
Good catch!

Fixed.

PR #1023, opened 2017-08-05
Provide minimal server impl with attributes handling

douglas-reid:
In a couple of contexts, we've been asked to provide a minimal implementation of the Mixer API that provides only the attributes handling part of the protocol -- and, critically, requires the fewest deps possible.

This PR provides one approach to dealing with those requests. It creates a `test` package with a stripped down set of dependencies.

To achieve this, I removed the `globalWordList.go` file from the `//pkg/api` package and added a bunch of rules to `BUILD.api` to generate such a file from the source at build time. I'm not proud of the way in which that generation is done, but... well, it works for now.

```release-note
Transitions static global word list to generated list from istio/api repo.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1023)


douglas-reid:
build mixer/e2e-smoketest

test/server.go
```diff
@@ -0,0 +1,123 @@
...
+	a.Attributes <- attribute.CopyBag(requestBag)
+
+	resp := &mixerpb.CheckResponse{
+		Precondition: mixerpb.CheckResponse_PreconditionResult{
+			Status:        rpc.Status{Code: int32(rpc.OK)},
+			ValidUseCount: 1,
+			Attributes:    req.Attributes,
+		},
+	}
+	return resp, nil
```
qiwzhang: Istio/proxy integration test expects  test_mix to perform followings
1) all attributes have been processed into attributeBug.
2) for Report batch,   delta update has been processed, multiple Report should be called for its own attributeBug
3) Quota is called separately from Check.

We basically intercept at adapterManager.AspectDispatcher interface.

https://github.com/istio/mixer/blob/master/pkg/api/grpcServer.go#L45

Here is istio/ proxy intergration code.

https://github.com/istio/proxy/blob/master/src/envoy/mixer/integration_test/mixer_server.go

douglas-reid: @qiwzhang 

1. This stub mirrors the exact logic in the current impl in Mixer. The bags sent on the channel are exactly as would be transmitted down to dispatch.

2. The deltas for Report() are applied and the new bags are sent on the channel (notic the `requestBag.UpdateBagFromProto` calls in the loop)

3. There is no more `Quota()` in the API.

qiwzhang: Not exactly.   Check and Quota are combined into one call in API.  but aspectDispatch has Quota call.  

This is how it is called.  https://github.com/istio/mixer/blob/master/pkg/api/grpcServer.go#L134

I feel this is part of transport.  it is better to be handled by this mock server.

Ideally, this mock server use api/grpcServer.go directly.   Just replace aspectDispatcher.Dispatch interface.

qiwzhang: For proxy integrate test, it need to set CheckResponse status too to test negative case.   Same for Quota,  it need to set quota allocated amount and response status.

test/server.go
```diff
@@ -0,0 +1,123 @@
...
+	"istio.io/api/mixer/v1/attributes"
+	"istio.io/mixer/pkg/attribute"
+)
+
+// AttributesServer implements the Mixer API to send mutable attributes bags to
+// a channel upon API requests. This can be used for tests that want to exercise
+// the Mixer API and validate server handling of supplied attributes.
+type AttributesServer struct {
+	GlobalDict map[string]int32
+	Attributes chan attribute.Bag
```
qiwzhang: we need to know which calls the attribute is coming from; check, report or quota.   We may need 3 channels, one for each.
Also for each call,  testing code may want to set response status too.  some calls are failed with specific error status.

For testing client cache,  testing code also need to return RelevantAttributes to tell client which attributes to use for cache.

test/server.go
douglas-reid: we don't want to import grpcServer directly. It involves all sorts of tracing and preprocess logic that are internal Mixer details that external tests should not depend on, nor need to know about.

I will add a way to set response values and separate out the channels for Check, Quota, and Response.

qiwzhang: It is fine with me.   If your interface is similar to 

adapterManager.AspectDispatcher

it will be much easier for me to use.

douglas-reid:
@qiwzhang PTAL. The fake now has three channels for receiving processed attributes and info (check, report, and quota(attributes + method args) ). You can also set the server to return GRPC errors as well as control the status code in `CheckResponse`. You can also override the default `QuotaResult` per request.

BUILD.api
```diff
@@ -118,3 +118,55 @@ genrule(
...
+genrule(
+    name = "mixer/v1/attributes_list",
+    srcs = [":mixer/v1/attributes_file"],
+    outs = ["attributes_list.txt"],
+    cmd = "grep -e '^- ' $(location :mixer/v1/attributes_file) | sed -e 's/- /\"/g' -e 's/$$/\",/g' > $@",
+    message = "Extracting strings",
+    visibility = ["//visibility:private"],
+)
+
+PRE = """package attributes; func GlobalList() ([]string) { tmp := make([]string, len(globalList)); copy(tmp, globalList); return tmp;}; var ( globalList = []string{"""
```
costinm: Would it be possible to at least write the code in proper python or shell, and just call that from bzl ? 
It is easier to maintain, test, read, debug, understand.

Bazel is a great build tool - but not a programming language :-)

douglas-reid: it certainly seems possible to create some python to write the source before using bazel to run goimports over it.

would you be opposed to handling that in a follow-on CL (with an issue tracking it)?

costinm:
It was just a suggestion, not a big issue :-)

test/server.go
```diff
@@ -0,0 +1,253 @@
...
+
+	if a.GenerateGRPCError {
+		return nil, errors.New("error handling check call")
+	}
+
+	requestBag := attribute.NewProtoBag(&req.Attributes, a.GlobalDict, attributes.GlobalList())
+	defer requestBag.Done()
+
+	// avoid blocked go-routines in testing
+	select {
```
qiwzhang: we need a flag not to send data to channel (in stress test mode). 

But I really think such customization should be in the testing code,  not in this mock stub.

If AttrbuteServer allows testing code to provide an interface, test server just passes the data to interface and get its response.   It will be much flexible for testing code to customize the server behavior.

test/server.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, errors.New("timeout trying to send attribute bag")
+	}
+
+	status := a.CheckResponseStatus
+
+	resp := &mixerpb.CheckResponse{
+		Precondition: mixerpb.CheckResponse_PreconditionResult{
+			Status:               status,
+			ValidUseCount:        DefaultValidUseCount,
+			Attributes:           req.Attributes,
```
qiwzhang: we don't echo attributes back to proxy.

test/server.go
qiwzhang: What I like is for AttributeServer to allow testing code to provide aspectDispatcher interface.

test/server.go
qiwzhang: My testing code may want to control returning attributes. we should have a way for testing code to set it.

test/server.go
douglas-reid: The `aspectDispatcher` interface is internal to Mixer. We should not be exposing it to testing code outside of Mixer. It is not appropriate for anything other than Mixer to be using for any purpose.

I've added a flag to exclude publishing of data to the channels.

There already exists a mechanism for providing a custom implementation. Nothing prevents users from supplying an implementation of `mixerpb.MixerServer` to the `NewMixerServer()` method if specialized behavior is desired. `AttributesServer` is just an example that may prove useful.

test/server.go
douglas-reid: I've updated to no longer echo and to take a full `mixerpb.CheckResponse`. Customize away!

BUILD.api
douglas-reid: moved to a py script and build refactored.

test/server.go
```diff
@@ -0,0 +1,252 @@
...
+				select {
+				case a.QuotaDispatches <- QuotaDispatchInfo{attribute.CopyBag(requestBag), qma}: //no-op
+				case <-time.After(3 * time.Second):
+					return nil, errors.New("timeout trying to send quota dispatch")
+				}
+			}
+
+			qr := mixerpb.CheckResponse_QuotaResult{
+				GrantedAmount:        param.Amount,
+				ValidDuration:        DefaultValidDuration,
```
qiwzhang: I wish my testing code could control granted_amount and valid_duration.   I have a rate limiting cache testing need that.

qiwzhang:
I can work with this mock server.   But I need to make a lot of changes in my testing code to use it.

Please looking at my testing code at:  https://github.com/istio/proxy/blob/master/src/envoy/mixer/integration_test/mixer_server.go

If your mock server allow testing code to provide

adapterManager.AspectDispatcher

interface,  it will be much easier for me to use it.

test/server.go
douglas-reid: good news: you can. If you set the `CheckResponse` field in the `AttributesServer`, then all of this will be ignored and overridden.

```
attrSrv.CheckResponse = &mixerpb.CheckResponse{ ... Quotas: ... "quota1": mixerpb.CheckResponse_QuotaResult{ GrantedAmount: myCustomAmt, ... } }
...
client.Check(...)
```

kyessenov: is there an example of how to import this? I want to get a fake mixer, too, in pilot.

douglas-reid: import the CheckResponse stuff?  you should just need the protos (+ codegen). I started on an effort to extract out the mixer proto codegen deps to make that easy to add to a WORKSPACE. you could copy the `BUILD.api` script from the Mixer repo, i guess. I'm open to ideas here. Mixer itself doesn't have a use case for this yet, so I need feedback as to what is useful/not useful and what the problems with use are.

Or were you talking about something else?

qiwzhang: I will be the first one to use it.   Ideally, it should be just adding a couple lines in my WORKSPACE,  and then I can use it in my go testing projects.  This is the goal we are shooting for.

douglas-reid:
@qiwzhang interface added as requested.

PR #1028, opened 2017-08-07
Add initial set of templates.

geeknoid:
Also improves comments in generated code.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1028)


geeknoid:
BTW: checkNothing is what the denial adapter consumes. reportNothing is intended for an eventual void adapter (think /dev/null) for Report calls.

geeknoid:
Also note that the code gen stuff has some issues at the moment, so these templates aren't quite right just yet. See embedded comments for details.

ZackButcher:
Is the plan to submit all the existing aspect configs as templates then trim them down as we've discussed  (removing all of the non-runtime fields from the template and pushing them into adapter config)? E.g. in the log adapters, severity, log_name, log_template, and payload_format should all make their way to adapter config, since they aren't derived at runtime, and we need to move a `severity_expression` field into the template as a `ValueType`.

guptasu:
Answering @ZackButcher's question: IMO, we are not planning to create templates in two steps (first templates containing details of adapter config and then moving that away in the adpater config). The templates that Martin is creating must, ideally, not contain any fields that are suppose to be eventually inside the adapter configuration.

geeknoid:
My expectation is that these are the real final templates.

In the case of LogEntry, don't you think severity is a property of an individual log entry? Maybe PayloadFormat should not be there though?

And for timestamp, in the old world it was filed in by the aspect manager. In the new world, I guess we could drop the field and have the adapter sample the current time itself? Or do we leave it in the template and have the operator specify something like request.time, or maybe a built-in time.now function?

template/logEntry/logEntry.proto
```diff
@@ -0,0 +1,94 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package logEntry;
+
+// TODO: STUFF THAT IS COMMENTED OUT NEEDS TO BE TURNED ON ONCE CODE GEN IS FIXED
```
ZackButcher: What's broken about it today? I'm not convinced we can populate any of these fields at runtime via codegen'd stuff. In particular:
- I don't know how we could produce a string value "log_name" at runtime in code genned code: where would the values come from?
- By the same token, how can we handle severity?
- What do we stick in the log_template string? Again, it's just a string from codegen's perspective, not special, so I don't see how we could ever provide a meaningful value.

These are all the things I thought we'd discussed moving out of templates. The rule of thumb we'd agreed on is that only runtime values belong in templates. Look at the metric template: it _only_ has ValueType fields, all of the non-runtime values (the type of the metric, buckets, etc) are in other configuration (today the aspect, in the future it'll move out to individual adapters).

ZackButcher: Instead, I think the template needs to look like this:

```proto
message Template {
    string severity = 2; // expression-populated value that must have ValueType `STRING`
    map<string, istio.mixer.v1.config.descriptor.ValueType> labels = 3;
}
```
(and the generated Instance has a `name` that we use as the log name field)

Then a logging adapter's configuration might look like:

```proto
message Params {
    map<string, my.backend.Severity> severity_map = 1; // maps string expression result to a severity value for this backend
    map<string, my.backend.Format> format_map = 2; // maps from instance name to format for that instance
    map<string, string> log_templates = 3; // maps from instance name to log template for that instance
}
```

We don't need to handle timestamps specifically as they can just be dimensions and derived from attributes (which is what you'd want to do for the log entry anyway; use `request.time` for start time and `request.duration` to compute the end time).

geeknoid: I'm not seeing things the same way at this point.

Think of a template as a generic type definition: Foo<T>. When you define Foo<T>, certain fields can have definite types, like string or int, while other fields are of type T. You can then define a type that derives from Foo<T> and defines what T means.

In the same way, fields in templates can be ValueType, indicating these fields' types must be defined by a concrete type. Fields of other types are already fully specified and so are used directly in the Type. So the fact the template has a field of type Severity means the Type will have an identical field, and thus the instance YAML will need to supply a value of type Severity. So at the point where a log entry is created, the severity is assigned on the spot. I think the same thing is true for timestamp and log name.

I agree that the payload format and template stuff should go elsewhere.

ZackButcher: This is a change in behavior over logging as implemented today: the severity is not static but instead varies at runtime based on attributes. You can't just say "any time this is logged it's CRITICAL" when e.g. writing application logs. Sometimes they're critical, sometimes they're info, but either way they use the same dimensions/log templates.

ZackButcher: BTW I'm all for keeping `timestamp` as a primitive we support, following the same rule where the expression populating the value has to have the type `TimeStamp` (or whatever we called time-based attributes).

geeknoid:
The model I'm suggesting makes it so the severity is assigned when the
instance is created. Isn't that what we want? Our expression language isn't
quite up to snuff yet, but you'd eventually be able to do:

severity: response.error == 123 ? WARNING : ERROR

This is basically enriching the types that can be produced by an expression
to include enums.

template/logEntry/logEntry.proto
ZackButcher: Also, different backends have different notions of severity. An adapter will already have to map from our severity enum to their own backend's notion of severity, so I don't see that it gains us anything.

ZackButcher:
Sorry, replied to the code comment, but aside from being able to use severity in an expression, which would be nice, different backends have different notions of severity. An adapter will already have to map from our severity enum to their own backend's notion of severity, so I don't see that it gains us anything.

geeknoid:
The benefit is like any other strong typing:

   - The operator trying to create a config will see a field of type
   Severity with well-known values instead of a field of type string.
   - The adapter implementer will receive a struct with a field of type
   Severity instead of string.

Yes, the adapter will still need to map our severity levels into their. But
their switch statement will be enum-based instead of string-based which I
do think is an improvement all around.

ZackButcher:
If we can make it work in general then I'm not opposed, but getting type checking to work for expressions with arbitrary user defined types will take a lot of work and require substantial changes to how we handle expressions in Mixer. In particular I worry that it implies we now need to start generating code for handling expressions in Mixer framework too. I know you want to do more powerful things in the expression language over time, but I'm still not totally on board yet.

guptasu:
I think we can do the follow to make custom types work..

consider a template
message Template {
  ValueType val = 1
  Severity sev_type = 2
  string someString = 3
  int64 someInt64 = 4
}

we generated
message Type {
  ValueType val = 1
  // DO NOT GENERATE ANYTHING FOR sev and someString and someInt64
}

message ConstructorParam /*aka InstanceParam*/ {
 string val = 1
 // NOTE: NOTE: DO NOT stringify user defined types that our expression eval won't work with.
 Severity sev_type = 2
 string someString = 3
 // NOTE: we stringify someInt64 because expression are allowed here..
 string someInt64
}
// I AM HOPING, Yaml parser might be able to parse instance config with Severity enum

type Instance {
  val interface{}
  sevType istio_mixer_tempalte_metric.Severity
  someString string
  someInt64 int64
}

Any concerns if we do this to support custom enums for things like severity.

I am hoping when we

ZackButcher:
But how do we actually handle Severity in our expression parsing code? It can't know about arbitrary types defined in templates without us generating at least some of the expression handling code. That's the part that scares me; generating the instances/constructors with the type name is easy.

guptasu:
We don't handler severity in expression parsing...

The generated code will look like the following
```
generatedCreateInstanceForMyTemplate(cstrParam ConstructorParam) {
  inst Instance
  inst.val = mapper.Eval(cstrParam.val)

  // Leave custom types alone...do not have them as expression evaluable.
  // **NOTE** We can do this inside codegen by having a whitelist of types for which mixer config 
  // parser will accept expression. This list include ValueType, int64, string, double, bool. May be
  // more in the future.
  inst.sevType = cstrParam.sevType

  inst.someString = mapper.Eval(cstrParam.someString).(string)
  inst.someInt64 = mapper.Eval(cstrParam.someInt64).(int64)  
}
```

ZackButcher:
That doesn't address that severity is dynamic; it becomes far less useful when the severity is hard coded in the action. Nor does it address what Martin wants to do, which _does_ involve having to handle Severity (or <insert arbitrary user defined type here>) at the expression level. I know he's not proposing that we start handling it now, but my concern that we'll wind up having to generate code to handle arbitrary types is unaddressed. I think this sets a bad precedence. If we can't support it now then don't put it in the template now, do something that works. When we can support it at the expression level, _then_ move it into the land of expressions.

The whole overarching approach towards templates, which I thought we had agreement on, is that anything in the template is a runtime value derived from attributes. Anything else should be adapter configuration.

ZackButcher:
Imagine the case where severity is specified directly (statically) in the constructor: then to log the same message at different severities I have to duplicate that constructor multiple times, changing only the severity. But, to actually get that to work as intended at runtime, I have to create multiple rules, each with different predicates, that determine which one gets called so that my severity ends up at the correct level. That seems horrible, and far easier to get wrong than the strawman of a mapping from some string to an actual severity that my backend understands.

geeknoid:
PTAL

tools/codegen/pkg/interfacegen/template/interface.go
```diff
@@ -29,40 +29,46 @@ import (
...
   Name string
   {{range .TemplateMessage.Fields}}
   {{.Comment}}
   {{.GoName}} {{replaceGoValueTypeToInterface .GoType}}
   {{end}}
 }
 
-// {{.Name}}HandlerBuilder must be implemented by adapter code if it wants to
-// process data associated with the template.
+// {{.Name}}HandlerBuilder must be implemented by adapters if they want to
```
ZackButcher: nit: remove the "if they want", just "FooHandlerBuilder must be implemented by adapters to process data associated with the Foo template." reads better IMO.

pkg/template/template.gen.go
```diff
@@ -31,6 +31,18 @@ import (
 	"istio.io/mixer/pkg/expr"
```
douglas-reid: curious: do we mean to have `*.gen.go` code in the repo?

template/checkNothing/BUILD
```diff
@@ -0,0 +1,8 @@
+package(default_visibility = ["//visibility:public"])
```
douglas-reid: shouldn't this be: `checknothing`?  no mixed case?

pkg/template/template.gen.go
ZackButcher: My understanding was that we'd leave in the generated code for the well known templates to make it easier for adapter authors.

template/checkNothing/checkNothing.proto
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package checkNothing;
+
+import "pkg/adapter/template/TemplateExtensions.proto";
+
+option (istio.mixer.v1.config.template.template_variety) = TEMPLATE_VARIETY_CHECK;
```
douglas-reid: in our design review, we were asked to move the annotations onto the messages themselves, and not at the package level. Can we do that before committing all of these templates?

template/listEntry/listEntry.proto
```diff
@@ -0,0 +1,29 @@
...
+import "pkg/adapter/template/TemplateExtensions.proto";
+
+option (istio.mixer.v1.config.template.template_variety) = TEMPLATE_VARIETY_CHECK;
+option (istio.mixer.v1.config.template.template_name) = "ListEntry";
+
+// ListEntry is used to verify the presence/absence of a string
+// within a list.
+message Template {
+    // Specifies the entry to verify in the list.
+    string value = 1;
```
douglas-reid: any reason to not make this a repeated field?

geeknoid: It's a good idea, but we don't have support for that in our data model. And then you'd have to also have an option for OR vs. AND semantics.

template/checkNothing/checkNothing.proto
geeknoid: I don't remember being asked to do this. But it would probably make sense. If we do this, could we also call the message by the desired name instead of calling 'em all Template? So the message would be called ListEntry.

@guptasu Sunny?

pkg/template/template.gen.go
geeknoid: Hmm, I thought this was temporary until we have the full build story in place.

template/checkNothing/BUILD
geeknoid: All right.

template/checkNothing/checkNothing.proto
guptasu: I don't remember this ask of moving annotations inside message (just created a new issue for this https://github.com/istio/mixer/issues/1040). However, it should be a trivial change. we can make this change.. it is a find and replace issue on the template itself (and codegen implementation) and nothing else would change.. I would prefer to do this as a separate refactor step (I can get to it within 2-3 days once I finish the task in hand) and bulk fix the existing template 

Regarding renaming of the Template message to a desired name message. I think calling it Template is very explicit and come handy once we start supporting sub messages and all the sub messages will be also defined inside the same proto file. It will help in finding which is the uber container root message. If we replace Template with any user defiend name, then the presence of template_variety will dictate the main root message. I personally like this explicit 'Template' declaration schema.. what do others think @ZackButcher @mandarjog @douglas-reid  ? Note if we rename 'Template' to 'ListEntry' then we also ditch the template_name option and simply just reuse the message name.

geeknoid:
Lower cases directory and package names and renamed protos to
template.proto.

PTAL.

template/listEntry/listEntry.proto
guptasu: Supported types are like a white list we don't support all types in the template. Main reason for supporting finite set is we want to create this magic Instance object and the generated code that wire up the things inside mixer. The generated code is data type specific (for example if it is a map<x,y> when we generated code in go we want to do a `for x, y := range fooMap, for non map we tackle different for list we will have to write the for loop slightly differently). It is definitely doable to extend the vocab of supported types in Template message if we have a use case. We just started with a small set.

The currently supported types are:
string, double, float, bool, ValueType and map<string, ValueType>

I am currently working on supporting few more types 
protobuf.datetime, protobuf.duration.

pkg/template/template.gen.go
guptasu: The generated file is copied to this dir via a script. 

In the future this is what we should be doing when we have the Build that would create a build of Mixer from all templates:
The build rule would invoke the tools/codegen/cmd/mixgenbootstrap to generate the template.gen.go... create a go_library out of it and make the mixer/pkg/template reference the on the fly generated go_library (which just contains the template.gen.go)

PR #1035, opened 2017-08-08
Use the zipkin option to ensure that mixer generates its own span, s…

ZackButcher:
…eparate from the client's

**Release note**:

```release-note
- Ensures Zipkin generates new child spans for Mixer rather than using the same span for both client and server.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1035)


ZackButcher:
Oh, I think like we need to turn off requiring approval from reviewers in github as it blocks the bot from merging too.

PR #1039, opened 2017-08-09
Update nil handling in the indexer expression code generator.

ozevren:
+ Original code assumed that the target of the index expression would be non-nil. This change generates code so that the case where the target can also be nil.
+ Fix a compiler bug that failed to surface internal compiler errors.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1039)


yutongz:
You can comment "/release-note XXX" to add release-note or simply comment "/release-note-none"

PR #1059, opened 2017-08-11
Introduce new-style noop adapter.

geeknoid:
We want to auto-generate this adapter, but in the meantime here it is.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1059)


adapter/noop2/noop.go
```diff
@@ -0,0 +1,147 @@
...
+	"istio.io/mixer/template/checknothing"
+	"istio.io/mixer/template/listentry"
+	"istio.io/mixer/template/logentry"
+	"istio.io/mixer/template/metric"
+	"istio.io/mixer/template/quota"
+	"istio.io/mixer/template/reportnothing"
+)
+
+type (
+	handler struct {
```
guptasu: do you want to add all implemented interfaces here too ?

guptasu: and in the handler below

adapter/noop2/noop_test.go
```diff
@@ -0,0 +1,160 @@
+// Copyright 2017 Istio Authors.
```
guptasu: If we auto generate noop, we also autogenerate _test for it ? It seems we should and we might be able to, looking at the content of the test below.

geeknoid: Yep, I added a comment to that effect.

adapter/noop2/noop.go
geeknoid: I actually removed everything here. As far as I can tell, putting those definitions there has zero effect on anything.

PR #1072, opened 2017-08-14
Implement new store API with k8s CRD backend

jmuk:
k8s custom resource definitions (CRDs) support is added although the use of CRDs are limited to the new API which isn't used in the actual code path yet.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1072)


pkg/config/crd/queue.go
```diff
@@ -0,0 +1,69 @@
...
+		select {
+		case <-q.ctx.Done():
+			break loop
+		case ev := <-q.chin:
+			evs := []store.Event{ev}
+			for len(evs) > 0 {
+				select {
+				case <-q.ctx.Done():
+					break loop
+				case ev := <-q.chin:
```
mandarjog: why do we need to connect 2 channels?  synchronous in and async out...

Why can't the client directly use the channel?

pkg/config/crd/resource.go
```diff
@@ -0,0 +1,124 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+//resource type represents the structure of a single custom resource.
+type resource struct {
+	Kind              string `json:"kind"`
+	APIVersion        string `json:"apiVersion"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
mandarjog: why is ObjectMeta embedded? by not meta ObjectMeta  or
Kind / APIVersion is same as  `metav1.TypeMeta`

pkg/config/crd/queue.go
```diff
@@ -0,0 +1,69 @@
...
+	cancel context.CancelFunc
+	chout  chan store.Event
+	chin   chan store.Event
+}
+
+func newQueue(ctx context.Context, cancel context.CancelFunc) *eventQueue {
+	eq := &eventQueue{
+		ctx:    ctx,
+		cancel: cancel,
+		chout:  make(chan store.Event, 10),
```
mandarjog: why 10, define constant with rationale.

pkg/config/crd/resource_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
+
+	cfg "istio.io/mixer/pkg/config/proto"
+)
+
+func deepEqualNonNil(v1 interface{}, v2 interface{}) bool {
```
mandarjog: Can you add comments as to why these methods are implemented?
They are not called from anywhere that I can see.

pkg/config/crd/store.go
```diff
@@ -0,0 +1,374 @@
...
+	dyn, err := dynamic.NewClient(s.conf)
+	if err != nil {
+		return err
+	}
+	s.caches = make(map[string]cache.Store, len(s.kinds))
+	for _, res := range resources.APIResources {
+		if _, ok := s.kinds[res.Kind]; ok {
+			// This is a bit tricky; the variable "res" may refer to the same address
+			// whose content can be updated through the iteration, and that goes
+			// the resource client to lookup different resources. It's safer to make a
```
mandarjog: That goes `to` the resource client  ?  'to' is missing.
Can you specify which line may modify the contents of res?

testdata/kubernetes/BUILD
```diff
@@ -0,0 +1,5 @@
+filegroup(
+    name="kubeconfig",
+    srcs=["config"],
```
mandarjog: where does the `src` come from? It seems there should be a file somewhere called config?

pkg/config/crd/resource_test.go
jmuk: done. It's used in `TestConvert` below. Conversion path of map[string]interface{} -> proto.Message -> map[string]interface{} could drop some keys for nil value, this function special handles the pattern like that.

pkg/config/crd/store.go
jmuk: modified the code a bit, and updated the comment here accordingly. I think this clarifies your questions.

PR #1076, opened 2017-08-14
Support istio label for service lookup from UID

douglas-reid:
To support ingress mixer integration, this PR adds support for deriving `targetService` from the `istio` label on a Pod. This is a short-term fix for a larger issue of doing proper UID --> service lookups (based on labels) meant to unblock ingress work. A follow-on PR will add service info caching and full label-based keys. A longer term effort will be required to handle multiple `target.service` (and `source.service`) derivations from a UID.

**Release note**:

```release-note
Support targetService derivation from "istio" label to enable Mixer-Ingress integration use case.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1076)


adapter/kubernetes/config/config.proto
```diff
@@ -92,6 +92,14 @@ message Params {
     // Default: app
     string pod_label_for_service = 6;
 
+    // In order to extract the service associated with a source, target, or
+    // origin, this adapter relies on pod labels. In particular, it looks for
+    // the value of a specific label for istio services, as specified by this
+    // parameter.
```
mandarjog: nit: istio_service is a bit ambiguous, perhaps istio_component_service. 
Please add examples of "istio=ingress, istio=mixer, istio=pilot" etc.

PR #1077, opened 2017-08-14
Create adapter_deps.bzl for Mixer

douglas-reid:
This PR separates out the core dependencies that Mixer uses to compile protos (primarily, adapter config protos) into a skylark set of repository rules.

The goal of this PR is to establish an easy-to-use rule in WORKSPACE files to import the same deps that Mixer is using internally. The intent is to enable adapter authors to quickly establish the bazel-ification necessary to mirror Mixer setup. This should enable adapter authors to test/build outside of mixer repo (in advance of mixer repo integration).

The set of dependencies was created based on a set of bazel queries to inspect various adapter config `BUILD`s.

Example:

```bash
$ bazel query 'deps("//adapter/serviceControl/config:go_default_library")' | grep -iv "go1_" | grep -iv "local_config" | sed -e "s/\/\/.*//g" | sort | uniq -c
   7
   7 @bazel_tools
  94 @com_github_gogo_protobuf
   4 @com_github_golang_glog
  13 @com_github_golang_protobuf
 430 @com_github_google_protobuf
  23 @io_bazel_rules_go
   2 @io_bazel_rules_go_toolchain
  47 @org_golang_google_grpc
  55 @org_golang_x_net
  22 @org_golang_x_text
   2 @org_pubref_rules_protobuf
```

**Release note**:

```release-note
Create skylark rules for core mixer proto dependencies.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1077)

### DIFF
--- a/adapter/list/list_test.go
+++ b/adapter/list/list_test.go
@@ -37,7 +37,7 @@ import (
 func TestBasic(t *testing.T) {
 	info := GetBuilderInfo()
 
-	if !contains(info.SupportedTemplates, listentry.TemplateName) {
+	if !containsListEntryTemplate(info.SupportedTemplates) {
 		t.Error("Didn't find all expected supported templates")
 	}
 
@@ -63,9 +63,9 @@ func TestBasic(t *testing.T) {
 	}
 }
 
-func contains(s []string, e string) bool {
+func containsListEntryTemplate(s []string) bool {
 	for _, a := range s {
-		if a == e {
+		if a == listentry.TemplateName {
 			return true
 		}
 	}
--- a/adapter_author_deps.bzl
+++ b/adapter_author_deps.bzl
@@ -1,14 +1,13 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_repository")
-load('@bazel_tools//tools/build_defs/repo:git.bzl', 'git_repository')
 
 def mixer_adapter_repositories():
-    git_repository(
+    native.git_repository(
         name = "org_pubref_rules_protobuf",
         commit = "9ede1dbc38f0b89ae6cd8e206a22dd93cc1d5637",  # Mar 31, 2017 (gogo* support)
         remote = "https://github.com/pubref/rules_protobuf",
     )
 
-    git_repository(
+    native.git_repository(
         name = "com_github_google_protobuf",
         commit = "52ab3b07ac9a6889ed0ac9bf21afd8dab8ef0014",  # Oct 4, 2016 (match pubref dep)
         remote = "https://github.com/google/protobuf.git",
--- a/adapter_author_deps.bzl
+++ b/adapter_author_deps.bzl
@@ -1,7 +1,7 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_repository")
 load('@bazel_tools//tools/build_defs/repo:git.bzl', 'git_repository')
 
-def mixer_proto_repositories():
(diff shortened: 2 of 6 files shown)


qiwzhang:
May be related.   Could you also create a mixer_mock_test.bzl  to load dependencies need to build mixer mock test?

douglas-reid:
@ZackButcher I've updated to `adapter_author_deps.bzl` and `mixer_adapter_repositories`. PTAL.

ZackButcher:
Nit: update release note in PR description.

adapter_author_deps.bzl
```diff
@@ -0,0 +1,57 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_repository")
+load('@bazel_tools//tools/build_defs/repo:git.bzl', 'git_repository')
```
lizan: No need to load `git_repository`, it can be called by `native.git_repository` instantly.

douglas-reid: Ah, good point. Thanks!

douglas-reid:
PTAL. I had to update the `bin/bazel_to_go.py` script to handle a new source for repo deps (something to keep in mind for future adapter authors) and fix a linter issue (unparam) in `list_test.go`

bin/bazel_to_go.py
```diff
@@ -127,20 +132,26 @@ def bazel_to_vendor(WKSPC):
         print "WORKSPACE file not found in " + WKSPC
         print "prog BAZEL_WORKSPACE_DIR"
         return -1
+    mixer_adapter_repos = WKSPC + "/adapter_author_deps.bzl"
```
lizan: No necessary in this PR, but this kind of parsing can be done by `bazel query`

Feed the output of `bazel query 'kind(go_repository, //external:*)' --output=build` to the AST here would be better :)

mandarjog: oh wow...
This is amazing !

douglas-reid: agreed that would better, but I didn't want to rewrite the whole thing in this PR.

douglas-reid:
I really don't understand. This PR causes no issues on my system, but generates infinite linter errors on prow.

adapter/list/list_test.go
```diff
@@ -63,9 +63,9 @@ func TestBasic(t *testing.T) {
 	}
 }
 
-func contains(s []string, e string) bool {
+func containsListEntryTemplate(s []string) bool {
```
geeknoid: Why this change?

bin/linters.sh
```diff
@@ -82,6 +82,7 @@ go_metalinter() {
         --enable=vet\
         --enable=vetshadow\
         --skip=testdata\
+        --skip=vendor\
         --vendor\
         $PKGS
 }
```
geeknoid: Should we add this .bzl file and others like it to the list of files over which we run buildifier? See the bottom of linters.sh.

adapter/list/list_test.go
douglas-reid: linter error: the method was only ever called with one string as `e`. `unparam` check.

bin/linters.sh
douglas-reid: i could be mistaken, but I was under the impression that `buildifier` only worked on `BUILD` files and could not be run on skylark files.

PR #1082, opened 2017-08-15
Mixer2 adapter dispatcher

mandarjog:
Adds Manager2 as the new adapter dispatcher.

This is the first part of a multi-PR work item.

Given a new interface `RuntimeState`  and `template.Repository` Manager2 dispatches to the generated code. It collects the results and sends them back the caller.

Next items.
1. Update GrpcServer to use the new dispatcher
2. Implement `RuntimeState` using Jun's CRD config work.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1082)


pkg/adapter/check.go
```diff
@@ -31,3 +31,36 @@ type CheckResult struct {
 	// ValidUseCount represents the number of uses for which this result can be considered valid.
 	ValidUseCount int64
 }
+
+// Result is the common interface supported by Results.
+type Result interface {
```
douglas-reid: should this live in a `result.go` or something file (other than `check.go`)?

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,439 @@
...
+	cpb "istio.io/mixer/pkg/config/proto"
+	"istio.io/mixer/pkg/expr"
+	"istio.io/mixer/pkg/pool"
+	"istio.io/mixer/pkg/status"
+	"istio.io/mixer/pkg/template"
+)
+
+// RuntimeAction is the runtime representation of configured action - cpb.Action.
+// Configuration is processed to hydrate instance names to Instances.
+type RuntimeAction struct {
```
douglas-reid: any reason to prepend `Runtime` here and elsewhere?

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,439 @@
...
+// Dispatcher#Report.
+func (m *Manager2) Report(ctx context.Context, requestBag attribute.Bag) error {
+	rt := m.RuntimeState()
+	// cannot be null by construction.
+	calls, err := fetchDispatchInfo(requestBag, adptTmpl.TEMPLATE_VARIETY_REPORT, rt, m.templateRepo, nil)
+	if err != nil {
+		return err
+	}
+
+	ra := make([]*runArg, 0, 10)
```
douglas-reid: why 10? should this be a const somewhere (as it is also used in `Check` and `Quota`) ?

pkg/adapterManager/manager2_test.go
```diff
@@ -0,0 +1,326 @@
...
+	ti, found := t.t[templateName]
+	return ti, found
+}
+
+type fakeRuntimeState struct {
+	ra  []*RuntimeAction
+	err error
+	hs  *HandlerState
+}
+
```
guptasu: Should there be a separate test for fetch mtd ? It seems to be doing interesting stuff and the result of the fake RuntimeState seems less challenging.

pkg/adapterManager/manager2.go
mandarjog: So as to distinguish it from the config elements of the same name.
If I name it `adapterManager.Action` it is still not clear if this is a runtime representation or not.

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,439 @@
...
+	// Instances to dispatch to the handler.
+	Instances []*cpb.Instance
+}
+
+// HandlerState represents a ready to use handler.
+type HandlerState struct {
+	// ready to use handler.
+	Handler adapter.Handler
+	// builder that produced the above handler.
+	Builder *adapter.BuilderInfo
```
mandarjog: Handler is not self describing so we need this Info.
In the `RuntimeState` implementation I will use this struct as the cache entry, So when a config changes, I can use the builder to create a new handler.

pkg/adapterManager/manager2_test.go
mandarjog: fetch is not a concern of the dispatcher.

Those tests will be covered in  RuntimeState tests.

pkg/adapterManager/manager2.go
guptasu: BuilderInfo will be renamed to Info, it is better to call tis variable into right now. We have a todo on top of BuilderInfo struct to rename it to Info soon.
This struct is static and filled during the mixer init time. We already have adapterinfo_registry.go that contains the lookup from adapter name to this Info, can we use that registry to discovery the info object. I think here we mainly just need the name of the handler right for debugging purpose ? If so, can we change this to just a string handlerName ?

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+// RuntimeState represents the current snapshot of the configuration database
+// and associated, initialized handlers.
+type RuntimeState interface {
+	// Resolve resolves configuration to a list of actions.
+	ResolveConfig(bag attribute.Bag, variety adptTmpl.TemplateVariety) ([]*RuntimeAction, error)
```
guptasu: ResolveConfig resolves

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,438 @@
...
+		gp:           gp,
+	}
+	m.SetRuntimeState(rt)
+	return m
+}
+
+// Manager2 is the 0.2 adapter manager.
+type Manager2 struct {
+	// mapper is the selector and expression evaluator.
+	// It is passed through by Manager.
```
guptasu: Should we remove the second comment? At least to me it is unclear which Manager we are referring to.

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+	}
+	return m.run(ctx, ra)
+}
+
+// Report dispatches to the set of adapters associated with the Report API method
+// Config validation ensures that things are consistent.
+// If they are not, we should continue as far as possible on the runtime path
+// before aborting. Returns an error if any of the adapters return an error.
+// If not the results are combined to a single CheckResult.
```
guptasu: we don't need this last statement. there is no result/CheckResult

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,438 @@
...
+
+	if buf != nil {
+		res.SetStatus(status.WithMessage(code, buf.String()))
+		pool.PutBuffer(buf)
+	}
+	return res, err.ErrorOrNil()
+}
+
+// doFunc is the abstraction used by runAsync to do work.
+type doFunc func(context.Context) *result2
```
guptasu: Should we call this dispatchFn ?

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,438 @@
...
+func (m *Manager2) dispatch(ctx context.Context, requestBag attribute.Bag, variety adptTmpl.TemplateVariety,
+	filterFunc filterFunc, genDoFunc genDoFunc) (adapter.Result, error) {
+	rt := m.RuntimeState()
+	// cannot be null by construction.
+	calls, err := fetchDispatchInfo(requestBag, variety, rt, m.templateRepo, filterFunc)
+	if err != nil {
+		return nil, err
+	}
+
+	ra := make([]*runArg, 0, 2*len(calls))
```
guptasu: Consider adding comment on why number 2.

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,438 @@
...
+				continue
+			}
+			mapKey := action.HandlerName + "/" + inst.Template
+			di := callmap[mapKey]
+			if di == nil {
+				di = &dispatchInfo{
+					processor:    &ti,
+					handlerName:  action.HandlerName,
+					handlerState: hs,
+					instances:    make([]*cpb.Instance, 0, 5),
```
guptasu: consider adding comment about no. 5

pkg/adapterManager/manager2_test.go
guptasu: Sorry, I was referring to fetchDispatchInfo, which does works on the result of RuntimeState.ResolveConfig.

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,439 @@
...
+		if !status.IsOK(st) {
+			if buf == nil {
+				buf = pool.GetBuffer()
+				// the first failure result's code becomes the result code for the output
+				code = rpc.Code(st.Code)
+			} else {
+				buf.WriteString(", ")
+			}
+
+			buf.WriteString(rs.callinfo.handlerName + ":" + st.Message)
```
mandarjog: lets do that later.

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,439 @@
...
+}
+
+// run runArgs using runAsync and return results.
+func (m *Manager2) run(ctx context.Context, runArgs []*runArg) (adapter.Result, error) {
+	nresults := len(runArgs)
+	resultsChan := make(chan *result2, nresults)
+	results := make([]*result2, nresults)
+
+	for _, ra := range runArgs {
+		m.runAsync(ctx, ra.callinfo, resultsChan, ra.do)
```
mandarjog: I have refactored so that Check/Report/Quota are smaller and dispatch does most of the work.

pkg/adapterManager/manager2_test.go
mandarjog: Looks like all the code paths are covered thru Check / Report and Quota.

pkg/adapterManager/manager2.go
douglas-reid: does `adapterManager` deal with anything but runtime representations?  it feels like that package is entirely runtime.

pkg/adapterManager/manager2_test.go
guptasu: I think the example of action containing instances across multiple templates is a interesting case. Sorry, there is a chance I might be missing to look for it,

pkg/adapterManager/manager2.go
guptasu: Sure, I am fine with that. Just to clarify I was referring to adding the code to the string rs.callinfo.handlerName + ":" + st.Code + " :" + st.Message + "\n"

pkg/adapterManager/manager2.go
guptasu: After the refactor , looks slick

pkg/adapterManager/manager2.go
```diff
@@ -0,0 +1,443 @@
...
+	// Quota dispatches to the set of aspects associated with the Quota API method
+	Quota(ctx context.Context, requestBag attribute.Bag,
+		qma *aspect.QuotaMethodArgs) (*adapter.QuotaResult2, error)
+}
+
+// dispatchInfo contains details needed to make a function call
+type dispatchInfo struct {
+	// generated code - processor.ProcessXXX ()
+	processor *template.Info
+	// Name of the handler being called.
```
guptasu: Not related to this PR, but I think we might want to  change the interface that adapter authors implement to 'Adapter' instead of Handler. It confuses me that whenever we refer to handler name, it is not the name of the adapter that is pass in via Adapter.info. Just to clarify this is the name of the handler configuration right ? Do you think we can name this handlerConfigName, I know it is too verbose, but may be will not confuse me later :-)

adapter/list/list_test.go
```diff
@@ -387,7 +387,7 @@ func TestBadUrl(t *testing.T) {
 	info := GetBuilderInfo()
 	builder := info.CreateHandlerBuilder()
 	cfg := config.Params{
-		ProviderUrl:     "http://abadurl.com",
```
geeknoid: Why this change? This needs to be an explicitly bad URL.

pkg/adapter/check.go
```diff
@@ -31,3 +31,24 @@ type CheckResult struct {
...
 }
+
+// GetStatus gets status embedded in the result.
+func (r *CheckResult) GetStatus() rpc.Status { return r.Status }
+
+// SetStatus embeds status in result.
+func (r *CheckResult) SetStatus(s rpc.Status) { r.Status = s }
+
+// Combine combines other result with self.
+func (r *CheckResult) Combine(otherPtr interface{}) interface{} {
```
geeknoid: This doesn't do anything with Status? Is this on purpose? A comment to that effect would be good.

pkg/adapter/quotas.go
```diff
@@ -108,10 +110,34 @@ type (
...
 )
+
+// GetStatus gets status embedded in the result.
+func (r QuotaResult2) GetStatus() rpc.Status { return r.Status }
+
+// SetStatus embeds status in result.
+func (r *QuotaResult2) SetStatus(s rpc.Status) { r.Status = s }
+
+// Combine combines other result with self.
+func (r *QuotaResult2) Combine(otherPtr interface{}) interface{} {
```
geeknoid: I think is a fundamentally invalid concept. You can't combine quotas like this, you end up just throwing away quota tokens.

pkg/adapter/result.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+package adapter
+
+import (
+	rpc "github.com/googleapis/googleapis/google/rpc"
+)
+
+// Result is the common interface supported by Results.
+type Result interface {
```
geeknoid: Given my comment above about QuotaResult2.Combine not being kosher, then it's likely this Result abstraction is not necessary,

pkg/adapterManager/manager2.go
geeknoid: I like both of Doug's suggestions. Dropping the Runtime prefix and renaming the package to runtime. Then we can have runtime.Manager.

adapter/list/list_test.go
mandarjog: It was failing tests because that url resolved to something.
The replacement is more guaranteed to be bad.

pkg/adapter/quotas.go
mandarjog: Do we need to combine results simply in an array then ?
Or is this where we will enforce the single dispatch semantic ?

mandarjog:
Created `pkg/runtime`, moved all code there. QuotaResult is no longer combined. Maintaining Result abstraction is still useful.

PTAL.

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+
+// Action is the runtime representation of configured action - cpb.Action.
+// Configuration is processed to hydrate instance names to Instances and handler.
+type Action struct {
+	// ready to use handler
+	handler adapter.Handler
+	// configuration for the handler
+	handlerConfig *cpb.Handler
+	// instanceConfigs to dispatch to the handler.
+	// instanceConfigs must belong to the same template.
```
geeknoid: So if instances from different templates are used in the same action within configuration, this will be translated into multiple different action objects in Go?

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+
+// Resolver represents the current snapshot of the configuration database
+// and associated, initialized handlers.
+type Resolver interface {
+	// Resolve resolves configuration to a list of actions.
+	Resolve(bag attribute.Bag, variety adptTmpl.TemplateVariety) ([]*Action, error)
+}
+
+// Newdispatcher creates a new dispatcher.
+func Newdispatcher(mapper expr.Evaluator, templateRepo template.Repository,
```
geeknoid: Newdispatcher -> NewDispatcher

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+	templateRepo template.Repository
+
+	// <resolver>
+	resolver atomic.Value
+
+	// gp is used to dispatch multiple adapters concurrently.
+	gp *pool.GoroutinePool
+}
+
+// SetState installs a new runtime state.
```
geeknoid: SetState -> SetResolver.
Can you expand the command to say when this happens? Is SetResolver called when Mixer configuration is updated by the user?

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+func (m *dispatcher) SetResolver(rt Resolver) {
+	m.resolver.Store(rt)
+}
+
+// State gets the current runtime state.
```
geeknoid: State -> Resolver

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+func (m *dispatcher) Resolver() Resolver {
+	return m.resolver.Load().(Resolver)
+}
+
+// Dispatcher dispatches incoming API calls to configured adapters.
```
geeknoid: Can you move this to the top of the file, it's more customary to have interfaces up front.

Please update aspect-related comments to more modern phrasing.

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+	// Report dispatches to the set of aspects associated with the Report API method
+	Report(ctx context.Context, requestBag attribute.Bag) error
+
+	// Quota dispatches to the set of aspects associated with the Quota API method
+	Quota(ctx context.Context, requestBag attribute.Bag,
+		qma *aspect.QuotaMethodArgs) (*adapter.QuotaResult2, error)
+}
+
+// dispatchInfo contains details needed to make a function call
+type dispatchInfo struct {
```
geeknoid: Move to top of file.

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+	"istio.io/mixer/pkg/template"
+)
+
+// Action is the runtime representation of configured action - cpb.Action.
+// Configuration is processed to hydrate instance names to Instances and handler.
+type Action struct {
+	// ready to use handler
+	handler adapter.Handler
+	// configuration for the handler
+	handlerConfig *cpb.Handler
```
geeknoid: If you already have a handler in-hand, than what is this config used for? This doesn't feel right...

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+// dispatchInfo contains details needed to make a function call
+type dispatchInfo struct {
+	// generated code - processor.ProcessXXX ()
+	processor *template.Info
+	// Name of the handler being called. Informational.
+	handlerName string
+	// Name of adapter that created the handler. Informational.
+	adapterName string
+	// handler to call.
+	handler adapter.Handler
```
geeknoid: Can handler and instances in this struct be replaced with an Action struct since that's what they are?

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+	// instance configuration to call the handler with.
+	instances []*cpb.Instance
```
geeknoid: Similar field is called instanceConfig in the Action struct, t'would be good to be consistent.

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+}
+
+// genDispatchFn creates dispatchFn closures based on the given call.
+type genDispatchFn func(call *dispatchInfo) []dispatchFn
+
+// dispatch dispatches to all function based on function pointers.
```
geeknoid: Correct phrasing.

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+							QuotaAmount:     qma.Amount,
+							BestEffort:      qma.BestEffort,
+						})
+					return &result{err, &resp, call}
+				},
+			}
+		},
+	)
+	var res *adapter.QuotaResult2
+	if qres != nil {
```
geeknoid: Is this nil check really necessary (here and in the other methods above)? Can't a nil value be casted?

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+		if rs.err != nil {
+			err = multierror.Append(err, rs.err)
+		}
+		if rs.res == nil { // When there is no return value like ProcessReport().
+			continue
+		}
+		if res == nil {
+			res = rs.res
+		} else {
+			if rc, ok := res.(adapter.ResultCombiner); ok { // only check is expected to supported combining.
```
geeknoid: Just check for CheckResult specifically

pkg/runtime/dispatcher.go
```diff
@@ -0,0 +1,443 @@
...
+		results[i] = <-resultsChan
+	}
+	return combineResults(results)
+}
+
+// runAsync runs the dispatchFn using a scheduler. It also adds a new span and records prometheus metrics.
+func (m *dispatcher) runAsync(ctx context.Context, callinfo *dispatchInfo, results chan *result, do dispatchFn) {
+	m.gp.ScheduleWork(func() {
+		// tracing
+		op := fmt.Sprintf("%s:%s(%s)", callinfo.processor.Name, callinfo.handlerName, callinfo.adapterName)
```
geeknoid: Use string concatenation which would be 100x faster than Sprintf.

pkg/runtime/dispatcher.go
geeknoid: Ah, I see below it's fetching some strings out of this config. Then let's make Action just hold those strings instead of the whole config. Otherwise, it really feels like a layering problem.

pkg/adapter/check.go
```diff
@@ -31,3 +31,25 @@ type CheckResult struct {
...
+
+// GetStatus gets status embedded in the result.
+func (r *CheckResult) GetStatus() rpc.Status { return r.Status }
+
+// SetStatus embeds status in result.
+func (r *CheckResult) SetStatus(s rpc.Status) { r.Status = s }
+
+// Combine combines other result with self. It does not handle
+// Status.
+func (r *CheckResult) Combine(otherPtr interface{}) interface{} {
```
geeknoid: Can make this strongly-typed now.

PR #1083, opened 2017-08-16
Add tool to generate adapter inventory

douglas-reid:
Initial PR to provide inventory generation. PR also cleans up some linter errors.

**Release note**:

```release-note
Adds tool to generate adapter inventory.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1083)


adapter/inventory.bzl
```diff
@@ -0,0 +1,27 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+def _inventory_gen(name, packages, out):
+  args = ""
+  for k, v in packages.items():
+    args += "-p %s:%s " % (k, v)
+  
+  native.genrule(
+      name = name+"_gen",
+      outs = [out],
```
guptasu: Naive bazel question, we are not passing the out to the tool, who is filling in the out file ?

tools/codegen/cmd/mixgeninventory/main.go
```diff
@@ -0,0 +1,63 @@
...
+		Short: "Generates mixer adapter inventory source code",
+		Long: "Generates mixer adapter inventory source code from an input map of adapter packages to include in the inventory.\n" +
+			"Example: mixgeninventory -p prometheus:istio.io/mixer/adapter/prometheus -p stdio:istio.io/mixer/adapter/stdio",
+		Run: func(cmd *cobra.Command, args []string) {
+			packageMap := make(map[string]string)
+			for _, maps := range mappings {
+				m := strings.Split(maps, ":")
+				packageMap[strings.TrimSpace(m[0])] = strings.TrimSpace(m[1])
+			}
+			if err := inventory.Generate(packageMap, os.Stdout); err != nil {
```
guptasu: Does stdout cause the terminal to see the entire output ?

tools/codegen/pkg/inventory/generator.go
```diff
@@ -0,0 +1,119 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// THIS FILE IS AUTOMATICALLY GENERATED.
+
+package adapter
+
+import(
+		{{range .}}"{{.Path}}"; {{end}}
```
guptasu: since we have the package name, should we use it here. {{.Package}} {{.Path}}.
Just in case two adapters have the same package name (but different full paths), the person who is building the mixer will have the responsibility to provide names to help us disambiguate. example 
prometheus:istio.io.template.prometheus, 
prometheus2:istio.io.template.v2.prometheus

guptasu: nit: consider indenting the strings for better readibility
import(
    {{range .}}
        "{{.Path}}"
    {{end}}
)

tools/codegen/pkg/inventory/generator.go
```diff
@@ -0,0 +1,119 @@
...
+	m := make(model, 0, len(packageMap))
+	for pkg, path := range packageMap {
+		m = append(m, Import{path, pkg})
+	}
+
+	sort.Sort(m)
+
+	buf := new(bytes.Buffer)
+	if err = tmpl.Execute(buf, m); err != nil {
+		return fmt.Errorf("could not generate inventory code: %v", err)
```
guptasu: I find it helpful during debugging to print the buffer along with the error. Mainly because the error would refer to a line number in a file that does not exists.

adapter/inventory.bzl
douglas-reid: in this case, the tool itself just dumps to stdout.  the genrule `cmd` redirects it to a file.  this was done for expediency/simplicity.  it should be very easy to add the logic around passing in a file stream to the command, if you think that is clearer.

tools/codegen/cmd/mixgeninventory/main.go
douglas-reid: if you run the tool in your terminal directly, you will see the entire output, yes.  As I said above, this was a quick way of handling the generation.  Adding logic around taking a file argument should be straightforward if you think that would be clearer.

tools/codegen/pkg/inventory/generator.go
douglas-reid: yeah, i was thinking about doing this (mainly to support a package that isn't the same as the path). I'll add it.

tools/codegen/pkg/inventory/generator.go
douglas-reid: OK. I'll add.

tools/codegen/cmd/mixgeninventory/main.go
guptasu: I would prefer -o options to make the tool standalone that takes input and produces output. However, I don't care for this particular case, since the stand alone tool that generates inventory has no use outside the mixer's build pipeline. Up to you.

adapter/inventory.bzl
guptasu: I would prefer -o options to make the tool standalone that takes input and produces output. However, I don't care for this particular case, since the stand alone tool that generates inventory has no use outside the mixer's build pipeline. Up to you.

adapter/BUILD
```diff
@@ -27,3 +28,17 @@ go_library(
         "//pkg/adapter:go_default_library",
     ],
 )
+
+inventory_library(
```
guptasu: Sorry forgot to ask one question in my initial review. Do you think it will be simpler if do not require mixer builders (operators) to edit this file and just ask them to edit the main build mixer_gen (does not exists yet) which will take templates and adapters both. Or may be that is your plan, this is a place holder for now.

geeknoid: In fact, if the user needs to edit the BUILD file anyway, then I don't see the point in generating inventory.go, they can just edit that themselves too.

The general idea I was going for here is that an operator could take an unchanged Mixer repo and augment it with their suite of adapters. If we're having the operator change files in the original Mixer repo, we've somewhat eroded that value prop.

douglas-reid: I don't think this mechanism is going to be the final form.

Having a single build rule named `mixer_binary` that takes deps and adapter package arguments and templates is still the goal. But it doesn't exist yet. This is most definitely a placeholder for such a rule (which would likely dispatch to the inventory.bzl script).

guptasu: That's what I thought. Just wanted to double check. thanks

PR #1087, opened 2017-08-16
Mock mixer test to support CheckResponse.

qiwzhang:
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1087)

### DIFF
--- a/test/handler.go
+++ b/test/handler.go
@@ -27,7 +27,7 @@ import (
 // to testing code for validation of proper transport using the Mixer API.
 type AttributesHandler interface {
 	// Check will be called once per Mixer API Check() request.
-	Check(attribute.Bag, *attribute.MutableBag) rpc.Status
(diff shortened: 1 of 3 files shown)


test/handler.go
```diff
@@ -27,7 +27,7 @@ import (
 // to testing code for validation of proper transport using the Mixer API.
 type AttributesHandler interface {
 	// Check will be called once per Mixer API Check() request.
-	Check(attribute.Bag, *attribute.MutableBag) rpc.Status
+	Check(attribute.Bag, *attribute.MutableBag) (CheckResponse, rpc.Status)
```
douglas-reid: it doesn't matter for this at all, but just to be clear, the real Mixer code doesn't generate `CheckResponse` structs from Check() dispatches. This does provide a convenient override however.

qiwzhang: Yes, my testing code needs this.

PR #1091, opened 2017-08-17
Update k8s adapter (TCP support for IP address pod resolution)

douglas-reid:
This PR adds IP_ADDRESS support to the k8s adapter, as well as to the config eval mechanisms (AST and IL).

The support added to the k8s adapter enables derivation of `*.service` attributes from `*.ip` (for k8s pods). To handle this, a new map is generated/updated based on a k8s watch of Pod resources. 
When updates arrive for pods, event handlers trigger that update the map based on pod metadata. 

As part of this PR, the default config that Mixer deploys with was modified in a few key ways:
- use `target.ip` and `source.ip` in attribute generation when `target.uid` and `source.uid` are not present. This is needed for TCP scenarios to derive `source.service`, etc.
- handle attribute types of `*.ip` as `[]uint8` throughout the Mixer codebase.
- introduce new expr function `ip(STRING)` which returns `IP_ADDRESS` values (implemented in both AST and IL)

Tested:

```
$ bazel-bin/cmd/server/mixs server --configStoreURL=fs://$(pwd)/testdata/configroot --logtostderr -v=6 --useAst=false 
```
```
$ bazel-bin/cmd/client/mixc report --bytes_attributes target.ip=0:0:0:0
Report RPC returned Internal (unable to resolve config: target.service attribute not found)

$ bazel-bin/cmd/client/mixc report --bytes_attributes target.ip=a:c:4:8
Report RPC returned OK

$ bazel-bin/cmd/client/mixc report
Report RPC returned Internal (unable to resolve config: target.service attribute not found)
```

as well as:

```
$ bazel-bin/cmd/server/mixs server --configStoreURL=fs://$(pwd)/testdata/configroot --logtostderr -v=6 --useAst=true 
```
```
$ bazel-bin/cmd/client/mixc report --bytes_attributes target.ip=0:0:0:0
Report RPC returned Internal (unable to resolve config: target.service attribute not found)

$ bazel-bin/cmd/client/mixc report --bytes_attributes target.ip=a:c:4:8
Report RPC returned OK

$ bazel-bin/cmd/client/mixc report
Report RPC returned Internal (unable to resolve config: target.service attribute not found)
```

Related to: #354

cc: @kyessenov 

**Release note**:

```release-note
Support target/source lookup via Pod IP.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1091)


adapter/kubernetes/cache.go
```diff
@@ -211,3 +298,76 @@ func (e eventType) String() string {
...
+	}
+	c.podSvcMapMutex.RUnlock()
+	c.podSvcMapMutex.Lock()
+	for _, deleteKey := range toDelete {
+		delete(c.podSvcMap, deleteKey)
+	}
+	c.podSvcMapMutex.Unlock()
+}
+
+func (c *controllerImpl) updateIPPodMap(obj interface{}) {
```
mandarjog: Say POD p1 belongs to service1 and service2.
service2 changed last, so in our map   p1 --> service2 is the mapping.

Now labels on POD p1 change, so it is no longer in service2.
We will get 2 calls I presume, 1 that the POD has changed. (this one)
And service membership has changed.

If we follow the service membership code, I am not sure it will be removed. 
Deletion only happens when service is deleted. either that or I am missing something.

mandarjog: If labels change, service membership should be re-evaluated.

Could you want endpoints which the endpoints controller already keeps in sync ?

pkg/il/compiler/compiler.go
```diff
@@ -206,6 +206,9 @@ func (g *generator) generateFunction(f *expr.Function, depth int, mode nilMode,
 		g.generateIndex(f, depth, mode, valueJmpLabel)
 	case "OR":
 		g.generateOr(f, depth, mode, valueJmpLabel)
+	case "ip":
```
ozevren: I think there is a great opportunity to generalize this code here by updating the default case instead.
default:
  for arg := range f.Args {
     g.generate(arg, ...)
  }
  g.builder.Call(f.Name)

pkg/il/evaluator/evaluator.go
```diff
@@ -146,7 +147,18 @@ func (e *IL) getOrCreateCacheEntry(expr string) (cacheEntry, error) {
 		glog.Infof("caching expression for '%s''", expr)
 	}
 
-	intr := interpreter.New(result.Program, make(map[string]interpreter.Extern))
+	ipExtern := interpreter.ExternFromFn("ip", func(in string) []byte {
+		if ip := net.ParseIP(in); ip != nil {
+			return []byte(ip)
+		}
+		return []byte{}
```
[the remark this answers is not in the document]
douglas-reid: `net.IP` will allow []byte with differing lens. It does however lead to weird behavior in other methods (like `String()`).

Error'ing may be the best path forward, as the zero value could lead to hard-to-debug errors later.  I'll update.

pkg/il/evaluator/evaluator.go
```diff
@@ -146,7 +147,18 @@ func (e *IL) getOrCreateCacheEntry(expr string) (cacheEntry, error) {
...
+	})
+
+	externMap := map[string]interpreter.Extern{
```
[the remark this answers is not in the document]
douglas-reid: Sure.

pkg/il/compiler/compiler.go
douglas-reid: I think you are probably right, but getting that all sorted out is a bit beyond what I want to tackle in this PR. I've added a TODO and intend on coming back to this ASAP.  OK?

ozevren: A TODO sounds good.

PR #1094, opened 2017-08-17
move mock mixer needed repositories into a bzl file.

qiwzhang:
So it will be easier for other repositories to load them.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1094)


repositories.bzl
```diff
@@ -1,22 +1,295 @@
...
-def new_git_or_local_repository(
-    name,
-    build_file,
-    path,
-    commit,
-    remote,
-    use_local = False):
+ISTIO_API_SHA = "a0ba5903ae2771eac13f80c3a936ed70fd3494d5"   # Aug 16, 2017 (no release)
+
+def go_istio_api_repositories(use_local=False):
```
douglas-reid: can we just make this a separate `api_repositories.bzl` that this script loads and calls?  that way it can be used elsewhere too.

repositories.bzl
```diff
@@ -1,22 +1,295 @@
...
-	    remote = remote
-        )
+      native.new_git_repository(
+          name = "com_github_istio_api",
+          build_file_content = ISTIO_API_BUILD_FILE,
+          commit = ISTIO_API_SHA,
+          remote = "https://github.com/istio/api.git",
+      )
+
+def go_googleapis_repositories():
```
douglas-reid: same question here: can we make this `googleapis.bzl` that gets loaded here (but can be used elsewhere)?

repositories.bzl
```diff
@@ -1,22 +1,295 @@
...
+)
+"""
+    native.new_git_repository(
+        name = "com_github_googleapis_googleapis",
+        build_file_content = GOOGLEAPIS_BUILD_FILE,
+        commit = "13ac2436c5e3d568bd0e938f6ed58b77a48aba15", # Oct 21, 2016 (only release pre-dates sha)
+        remote = "https://github.com/googleapis/googleapis.git",
+    )
+
+def go_x_tools_imports_repositories():
```
douglas-reid: same question as above: can we make this a separate file?

repositories.bzl
qiwzhang: Done

repositories.bzl
qiwzhang: Done

repositories.bzl
```diff
@@ -1,22 +1,295 @@
...
+    native.new_git_repository(
+	name = "org_golang_x_tools_imports",
+        build_file_content = BUILD_FILE,
+        commit = "e6cb469339aef5b7be0c89de730d5f3cc8e47e50",  # Jun 23, 2017 (no releases)
+        remote = "https://github.com/golang/tools.git",
+    )
+
+# This function should be used by others to use mock mixer.
+def mixer_test_repositories(use_local_api=False):
+    go_repositories()
```
qiwzhang: Dont't want to put too many things in WORKSPACE.  For example, istio/proxy WORKSAPCE are c++ related.

repositories.bzl
```diff
@@ -1,22 +1,295 @@
...
+    gogo_proto_repositories()
```
qiwzhang: It has some dependencies that mock test doesn't need.

test/repositories.bzl
```diff
@@ -0,0 +1,80 @@
...
+#     commit = "4b3296a43ce940ba47fab7ad35fdf5c0c18778cd",
+#     importpath = "github.com/istio/mixer",
+# )
+#
+# load("@com_github_istio_mixer//test:repositories.bzl", "mixer_test_repositories")
+# mixer_test_repositories(False)
+#
+def mixer_test_repositories(use_local_api=False):
+    go_repositories()
+    gogo_proto_repositories()
```
douglas-reid: my only concern with using `gogo_proto_repositories()` here is that we may have to keep it in sync with the stuff in `adapter_author_repositories()`. We explicitly chose there to override the ones in `gogo_proto_repositories()` to maintain complete control.

i'm not sure how big of an issue this is for a test setup, however.

qiwzhang: this function will be called by other repos, such as proxy, or pilot.  they may not call adapter_author_repositories().  mixer will not call this function.

PR #1097, opened 2017-08-17
Initial Circonus Metrics Adapter

redhotpenguin:
This PR is the initial Circonus metrics adapter work. I've tried to follow the patterns and styles present in the statsd, prometheus, and Stackdriver work.

I've verified that all tests for mixer (including the tests introduced with this change) are passing under the bazel build harness with the `make test` target.

Submitting this for initial review and feedback. After talking with @ZackButcher at the CloudFlare meetup a couple weeks ago, it sounds like some parts of the API will be changing, and I'm prepared to make the necessary changes on this PR to accommodate those. I was on the Istio community call last week, so I'm here to respond, and also provide user feedback on changes in the API.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1097)

### DIFF
--- a/pkg/expr/eval_test.go
+++ b/pkg/expr/eval_test.go
@@ -16,6 +16,8 @@ package expr
 
 import (
 	"fmt"
+	"net"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -105,6 +107,14 @@ func TestGoodEval(tt *testing.T) {
 				"request.user": "user2",
 			},
 			"user1", "",
+		},
+		{
+			`source.name| source.target`,
+			map[string]interface{}{
+				"source.name":   nil,
+				"source.target": nil,
+			},
+			nil, "",
 		},
 		{
 			`request.size| 200`,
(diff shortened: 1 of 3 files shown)


redhotpenguin:
Note - it looks like some changes were applied to master between the time I started this PR, got a hotdog, and submitted it, so I just rebased and pushed again. Thanks!

mandarjog:
@redhotpenguin Thanks for the PR. Like you mention in your comments, the input format is changing, along with how interfaces are defined. 

Here is how metric template is currently defined.
https://github.com/istio/mixer/blob/master/template/metric/template.proto
You should be able to see the file it generates in your workspace at the same location.

Take a look at structs `Type`, `InstanceParams`  and `Instance`.

/ok-to-test

adapter/circonus/circonus_test.go
```diff
@@ -0,0 +1,307 @@
...
+)
+
+var validTestGauge = int64(123)
+var validTestCounter = int64(1)
+var validTestHistogramString = "H[1.4e+08]=1"
+var validTestHistogramVal = 146 * time.Millisecond
+var validTestHistogramIntVal = 3459
+var validTestHistogramIntString = "H[3.4e+03]=1"
+
+func testServer(t *testing.T) *httptest.Server {
```
mandarjog: Does the test need to be this complex?
Can you define an interface with `Timing`, `Gauge`, and `Increment` functions and just test what you send to your fake rather than adding all the http details here?

mandarjog:
@redhotpenguin please ensure that you run
`bin/fmt.sh`  and `bin/linters.sh` before submitting the PR. Thanks.

ZackButcher:
FWIW I'm going to start in on new style metrics adapters for all of our existing metrics adapters tomorrow, so there should be some examples of the new style stuff shortly.

adapter/circonus/circonus_test.go
redhotpenguin: I've got that approach with the interface for `Timing` etc in progress. I ran into some package namespace errors that I haven't quite resolved yet, not sure if they are on the `bazel` side or `g`o. An introspection method was just added to `circonusgometrics` that I may be able to use instead. I agree that the existing test is overly complex and intend to simplify it.

redhotpenguin:
I've committed `96b80ed` with `fmt.sh` and `linter.sh` executed on the code. There's an existing linter complaint about checking the error from `defer r.Body.Close()`, but I see that idiom used elsewhere in mixer, so I'm assuming that's acceptable to move ahead with.

redhotpenguin:
Scratch my comment on `defer r.Body.Close()` - `mixer-presubmit.sh` seems to be erroring on those linter issues. I'll address those.

ZackButcher:
@redhotpenguin FYI here's the rewrite of Prometheus over to the new style for reference: https://github.com/istio/mixer/pull/1122 (here's the diff: https://gist.github.com/ZackButcher/efb457a3b605ad1b0cd3be09f2273207)

ZackButcher:
Marking do not merge for now to ensure the bot doesn't do it automatically while there are pending comments. We need to nail down lgtm vs approve behavior.

redhotpenguin:
@ZackButcher @mandarjog I've resolved the linter failures that were occurring with `prow/mixer-presubmit.sh`. Let me know if there is additional work you would like to see here. The tests are not a good as they could be, but I'm planning on a follow up PR for the new metrics style, so I could include that work there.

redhotpenguin:
I think I fat fingered this close (perhaps somewhere on Gogland). Will be reopening with the updated 0.2 API.

PR #1103, opened 2017-08-18
Introduce the stdio adapter.

geeknoid:
This adapter sends log entries and metrics to stdio streams.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1103)

### DIFF
--- a/adapter/stdio/stdio.go
+++ b/adapter/stdio/stdio.go
@@ -1,0 +1,246 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package stdio // import "istio.io/mixer/adapter/stdio"
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"time"
+
+	multierror "github.com/hashicorp/go-multierror"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"istio.io/mixer/adapter/stdio/config"
+	"istio.io/mixer/pkg/adapter"
+	"istio.io/mixer/template/logentry"
+	"istio.io/mixer/template/metric"
+)
+
+type (
+	zapBuilderFn func(outputPath string, encoding string) (*zap.Logger, error)
+	getTimeFn    func() time.Time
+
+	builder struct {
+		zapBuilder   zapBuilderFn // indirection to allow override in tests
+		logEntryVars map[string][]string
+		metricDims   map[string][]string
+	}
+
+	handler struct {
+		logger         *zap.Logger
+		severityLevels map[string]zapcore.Level
+		metricLevel    zapcore.Level
+		getTime        getTimeFn
+		logEntryVars   map[string][]string
+		metricDims     map[string][]string
+	}
+)
+
+// ensure our types implement the requisite interfaces
+var _ logentry.HandlerBuilder = &builder{}
+var _ logentry.Handler = &handler{}
+var _ metric.HandlerBuilder = &builder{}
+var _ metric.Handler = &handler{}
+
+///////////////// Configuration Methods ///////////////
+
+func (b *builder) Build(cfg adapter.Config, _ adapter.Env) (adapter.Handler, error) {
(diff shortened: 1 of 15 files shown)


adapter/stdio/stdio.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (builder) Build(cfg proto.Message, _ adapter.Env) (adapter.Handler, error) {
+	c := cfg.(*config.Params)
+
+	outputPath := stdErr
+	if c.LogStream == config.STDOUT {
+		outputPath = stdOut
+	}
+
+	buildZap := newZapLogger
```
ZackButcher: This var is only used once, just inline to

```go
zapLogger, err := newZapLogger(outputPath)
```

adapter/stdio/stdio.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (h *handler) HandleMetric(_ context.Context, instances []*metric.Instance) error {
+	var errors *multierror.Error
+
+	fields := make([]zapcore.Field, 0, 6)
+	for _, instance := range instances {
+		entry := zapcore.Entry{
+			LoggerName: instance.Name,
+			Level:      h.metricLevel,
+			Time:       time.Now(),
```
ZackButcher: Arguably we should use `request.time` for this, but it's not easy to plumb the attribute through. We should probably promote it up to a field on the metric Instance.

mandarjog:
@geeknoid 
```
I0820 02:53:13.837] adapter/stdio/stdio_test.go:309::warning: declaration of "err" shadows declaration at stdio_test.go:271 (vetshadow)
I0820 02:56:54.221] adapter/stdio/stdio_test.go:212:5:warning: ineffectual assignment to err (ineffassign)
I0820 02:56:54.222] adapter/stdio/stdio_test.go:271:5:warning: ineffectual assignment to err (ineffassign)
I0820 02:59:49.179] adapter/stdio/stdio_test.go:212:5:warning: this value of err is never used (SA4006) (staticcheck)
I0820 02:59:49.180] adapter/stdio/stdio_test.go:271:5:warning: this value of err is never used (SA4006) (staticcheck)
I0820 03:08:13.022] adapter/stdio/stdio_test.go:1::warning: file is not gofmted with -s (gofmt)
```

adapter/stdio/stdio.go
```diff
@@ -0,0 +1,199 @@
...
+func mapConfigLevel(l config.Params_Level) zapcore.Level {
+	if l == config.WARNING {
+		return zapcore.WarnLevel
+	} else if l == config.ERROR {
+		return zapcore.ErrorLevel
+	}
+	return zapcore.InfoLevel
+}
+
+func (builder) ConfigureLogEntryHandler(map[string]*logentry.Type) error {
```
mandarjog: A builder should implement this interface only if it does something meaningful.

geeknoid: I thought it was required to be implemented. Making this optional would be great.

@guptasu You on board with that idea? If you declare you support a template in the BuilderInfo struct, it just means you implement the handler, and optionally the handler builder ConfigureXXX method.

guptasu: Hmm.. I think that is too risky since it is prone to spelling mistake errors. Operator might miss spell the method and we will silently stop serving them type information. Too risky I think.
IMO, being strict is better here, writing an empty method is no big deal + adapter developer and his code reviewers will exactly know the reason for the method to be empty.

As I am writing the developer guide, I realize we are doing enough black magic and adding more of a black boxed concept will only add to confusion.

Having said that, if we really want this optional implementation of HandlerBuilder and still have a strict validation, we should ask the adapter developer to explicitly state in BuilderInfo (aka adapter.Info) that he wants just 'Handler' interface part of a particular mesh function. So, declaring just the mesh function name in the BuilderInfo will not be sufficient, they will have to give something in the lines of:
"metric.TemplateName,Handler" or "metric.TemplateName,Handler, HandlerBuilder". Which I think is also a complex concept to understand.

geeknoid:
PTAL.

Issues addressed, now featuring 100% test coverage.

adapter/stdio/stdio.go
guptasu: @mandarjog I think relying on validation tests can be risky, adapter developers might not be following the best practices of running the test suite. Also, I am not sure if we will be able to catch mis spelled method names via pure reflection without causing false positives.

guptasu:
I already have validation to ensure HandlerBuilder implement the interfaces for supported templates. Therefore, I would expect the mixs server to fail to start (in whatever tests we have that starts the server. Maybe e2e) since there is a faulty adapter attached to the mixer server.

ZackButcher:
Something might actually be broken in this PR. There's not a ton of runs, but it's only really failing consistently for this PR: https://prow.istio.io/?job=mixer-e2e-smoketest

yutongz:
BTW, we change something and now we do not need to say "/approve no-issue" any more. Just "/approve" is good enough for every pr even without related issue.

PR #1106, opened 2017-08-18
Log errors as error and don't log basic info unless verbosity is at least 1

ldemailly:
Log errors as error and don't log basic info unless verbosity is at
least 1

This is to avoid
```
0818 22:08:39.405042   21321 grpcServer.go:216] Report 6 returned with: OK 
I0818 22:08:39.405056   21321 grpcServer.go:82] Dispatching Preprocess
I0818 22:08:39.405251   21321 grpcServer.go:85] Preprocess returned with: OK 
I0818 22:08:39.405261   21321 grpcServer.go:82] Dispatching Preprocess
I0818 22:08:39.405280   21321 grpcServer.go:85] Preprocess returned with: OK 
I0818 22:08:39.405284   21321 grpcServer.go:103] Dispatching Check
I0818 22:08:39.405292   21321 grpcServer.go:105] Check returned with: OK 
I0818 22:08:39.405301   21321 grpcServer.go:134] Dispatching Quota
I0818 22:08:39.405263   21321 grpcServer.go:103] Dispatching Check
I0818 22:08:39.405357   21321 grpcServer.go:82] Dispatching Preprocess
I0818 22:08:39.405374   21321 grpcServer.go:105] Check returned with: OK 
```
and show errors when we get `CANCELLED`

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1106)

```release-note
improve logging (less verbose for normal case, more verbose for errors)
```

### DIFF
--- a/pkg/api/grpcServer.go
+++ b/pkg/api/grpcServer.go
@@ -79,9 +79,7 @@ func (s *grpcServer) Check(legacyCtx legacyContext.Context, req *mixerpb.CheckRe
 
 	globalWordCount := int(req.GlobalWordCount)
 
-	if glog.V(1) {
-		glog.Info("Dispatching Preprocess Check")
-	}
+	glog.V(1).Info("Dispatching Preprocess Check")
 	preprocResponseBag := attribute.GetMutableBag(requestBag)
 	out := s.aspectDispatcher.Preprocess(legacyCtx, requestBag, preprocResponseBag)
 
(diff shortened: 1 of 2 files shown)


pkg/api/grpcServer.go
```diff
@@ -79,16 +79,21 @@ func (s *grpcServer) Check(legacyCtx legacyContext.Context, req *mixerpb.CheckRe
 
 	globalWordCount := int(req.GlobalWordCount)
 
-	glog.Info("Dispatching Preprocess")
+	if glog.V(1) {
```
geeknoid: Here and elsewhere, you can just write:

glog.V(1).Info...

ldemailly: oh thanks, that'll be a lot less ugly/more concise, updating

ZackButcher: The trade-off is that the `if glog.V(foo) { ... }` style avoids evaluating arguments if logging is at a lower level than you're checking, while you'll always eval arguments in the `glog.V(foo).Info(bar)` style.

ldemailly: yes, in this case there was no expensive calculations/function calls
C++ has auto short circuit which is nice but this will do

PR #1109, opened 2017-08-20
Adapter 2 dispatch: Resolver

mandarjog:
1. Simplify Resolver interface
2. Introduce Actions interface to include lifecycle for ref counting.
3. Implement resolver that needs `rules map[string][]*Rule` as datastore.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1109)


pkg/runtime/dispatcher.go
```diff
@@ -208,29 +233,30 @@ func (m *dispatcher) Quota(ctx context.Context, requestBag attribute.Bag,
 	qma *aspect.QuotaMethodArgs) (*adapter.QuotaResult2, error) {
 	dispatched := false
 	qres, err := m.dispatch(ctx, requestBag, adptTmpl.TEMPLATE_VARIETY_QUOTA,
-		func(inst *cpb.Instance) bool {
-			return inst.Name == qma.Quota
-		},
```
mandarjog: Removed filter function as an arg, so the caller does its own filtering.

WORKSPACE
```diff
@@ -104,8 +104,11 @@ go_repository(
 load("//:x_tools_imports.bzl", "go_x_tools_imports_repositories")
 load("//:googleapis.bzl", "go_googleapis_repositories")
 load("//:istio_api.bzl", "go_istio_api_repositories")
+
 go_x_tools_imports_repositories()
+
```
mandarjog: automatic changes made by buildifier.

bin/fmt.sh
```diff
@@ -24,6 +24,6 @@ done
 gofmt -s -w ${GO_FILES}
 goimports -w -local istio.io ${GO_FILES}
 buildifier -mode=fix $(find . -name BUILD -type f)
-buildifier -mode=fix ./BUILD.api
+buildifier -mode=fix ./*.bzl
```
mandarjog: BUILD.api file was removed in the last commit and replaced with severl .bzl files.
This change ensures that fmt.sh runs to completion.

pkg/adapterManager/manager.go
```diff
@@ -56,15 +56,15 @@ var (
 	dispatchCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: "mixer",
-			Subsystem: "adapter",
+			Subsystem: "adapter_old",
```
mandarjog: `adapter` subsystem is now used by the "new adapter" code.
Therefore moving the older metric to "adapter_old" subsystem.
This will be removed once switch over is complete.

pkg/attribute/mutableBag.go
```diff
@@ -64,6 +64,14 @@ func GetMutableBag(parent Bag) *MutableBag {
 	return mb
 }
 
+// GetFakeMutableBagForTesting returns a Mutable bag based on the specified map
+// Use this function only for testing purposes.
+func GetFakeMutableBagForTesting(v map[string]interface{}) *MutableBag {
+	m := GetMutableBag(nil)
```
mandarjog: There is code to create a mutable bag / or bag replicated in several places in the code base.
The primary reason is that there is no bag constructor that takes the underlying map.

This method is marked for testing use only.

ZackButcher: Lets make this clearer: add a `*testing.T` parameter to the func.

pkg/runtime/dispatcher.go
```diff
@@ -103,43 +114,61 @@ type dispatcher struct {
...
-	// <resolver>
-	resolver atomic.Value
-
 	// gp is used to dispatch multiple adapters concurrently.
 	gp *pool.GoroutinePool
+
+	// protects resolver
+	sync.RWMutex
+
+	// <resolver>
```
ZackButcher: Just remove this comment, it doesn't add anything and lets you collapse the mutex beside the resolver.

```go
m sync.WRMutex
r Resolver
```

pkg/runtime/dispatcher.go
```diff
@@ -103,43 +114,61 @@ type dispatcher struct {
...
 	mapper expr.Evaluator
 
-	// <resolver>
-	resolver atomic.Value
-
 	// gp is used to dispatch multiple adapters concurrently.
 	gp *pool.GoroutinePool
+
+	// protects resolver
+	sync.RWMutex
```
ZackButcher: Make this a struct field, don't inline it. Making the struct itself a mutex is not what we need: you want a private mutex to guard a field in some methods, but instead you have a public mutex useable by anyone with a `dispatcher` in hand (i.e. the interface this struct presents does not match its intended use, and allows mis-use for no gain).

pkg/runtime/dispatcher.go
```diff
@@ -103,43 +114,61 @@ type dispatcher struct {
...
 // This function is called when configuration is updated by the user.
-func (m *dispatcher) SetResolver(rt Resolver) {
-	m.resolver.Store(rt)
+// oldResolver is returned so that it can be reclaimed.
+func (m *dispatcher) SetResolver(rt Resolver) (oldResolver Resolver) {
+	m.Lock()
+	oldResolver = m.resolver
+	m.resolver = rt
+	m.Unlock()
+	return oldResolver
```
ZackButcher: Just `return`, or remove the named return val.

pkg/runtime/resolver.go
```diff
@@ -0,0 +1,214 @@
...
+		resolveDuration.With(lbls).Observe(time.Since(start).Seconds())
+		resolveRules.With(lbls).Observe(float64(nselected))
+		raLen := 0
+		if ra != nil {
+			raLen = len(ra.Get())
+		}
+		resolveActions.With(lbls).Observe(float64(raLen))
+	}()
+
+	//
```
ZackButcher: nit: remove

pkg/runtime/resolver.go
```diff
@@ -0,0 +1,214 @@
...
+	rulesArr := make([][]*Rule, 0, 2)
+
+	// add default namespace if present
+	if dcf := r.rules[r.defaultConfigNamespace]; dcf != nil {
+		rulesArr = append(rulesArr, dcf)
+	} else if glog.V(3) {
+		glog.Infof("Resolve: no namespace config for %s", r.defaultConfigNamespace)
+	}
+
+	target = attr.(string)
```
ZackButcher: I don't think we check anywhere that the identity attribute is a `STRING` - we need to.

pkg/attribute/mutableBag.go
mandarjog: We don't want to compile in testing.T artifacts in production code. I think it is plenty clear from the method name, what it the intended use.

pkg/runtime/dispatcher.go
mandarjog: done.

pkg/runtime/dispatcher.go
mandarjog: done.

pkg/runtime/dispatcher.go
mandarjog: done.

pkg/runtime/resolver.go
mandarjog: good point. we should not check it here though. Identity attribute is part of the configuration, and we should fail validation if that attribute is ...`request.size`.

Practically that attribute is `destination.service` or (`target.service`) in the old nomenclature.

PR #1111, opened 2017-08-21
Add dns1035 compatible shortname to builderInfo

mandarjog:
All adapter configs (handlers) and template configs (instances) are CRD types in kubernetes.

In order to register a K8s type, the names must be a dns label (dns1035) compatible.
This PR introduces a derived BuilderInfo.ShortName that is computed from `.Name`.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1111)

### DIFF
--- a/pkg/adapter/info.go
+++ b/pkg/adapter/info.go
@@ -23,6 +23,9 @@ import (
 type BuilderInfo struct {
 	// Name returns the official name of the adapter.
 	Name string
+	// ShortName is used to register this type with the config system.
+	// It must be DNS-1035 compatible. If it is not specified, it will be derived from 'Name'.
+	ShortName string
 	// Description returns a user-friendly description of the adapter.
 	Description string
 	// CreateHandlerBuilder is a function that creates a HandlerBuilder which implements Builders associated
--- a/pkg/config/adapterInfoRegistry.go
+++ b/pkg/config/adapterInfoRegistry.go
@@ -16,6 +16,7 @@ package config
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/golang/glog"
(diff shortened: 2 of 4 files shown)


pkg/adapter/info.go
```diff
@@ -23,6 +23,9 @@ import (
 type BuilderInfo struct {
 	// Name returns the official name of the adapter.
 	Name string
+	// ShortName is used to register this type with the config system.
```
geeknoid: Why don't we just require the name to be in a compatible form to begin with, rather than complicate things by having N named?

mandarjog: we have been using the full go package path as the name, which is useful for a developer to find code. These names are already unique. All this information is recorded in the CRD. one can derive short name from the full name. 
If we only require shortname, this info is lost.

pkg/config/adapterInfoRegistry.go
```diff
@@ -27,47 +28,104 @@ type adapterInfoRegistry struct {
...
+	adapterName = strings.Replace(adapterName, ".", "-", -1)
+	const github = "github-com" // since . is replaced by - above.
+
+	// if name is already compatible just use it.
+	if DNS1035.MatchString(adapterName) {
+		return adapterName, nil
+	}
+
+	comps := strings.Split(adapterName, "/")
+	if len(comps) < 3 || (comps[0] == github && len(comps) < 4) {
```
ZackButcher: Why validate based on exact lengths, that feels wrong. What if I host my own private git repo with an adapter at at `myinternalurl.com/adapter`? Seems like overkill.

pkg/config/adapterInfoRegistry.go
```diff
@@ -27,47 +28,104 @@ type adapterInfoRegistry struct {
...
+
+// DNS1035 is the prebuilt matcher for DNS-1035 compatible names.
+var DNS1035 = regexp.MustCompile(DNS1035Str)
+
+func dns1035Err(name string) error {
+	return fmt.Errorf("%s cannot be converted to a DNS-1035 compatible name %s", name, DNS1035Str)
+}
+
+// NameToCfgKind returns Kind name given an adapter name.
+// istio.io/mixer/adapter/denier --> denier-mixer-istio-io
```
geeknoid: Should include "adapter" in there, right?

pkg/config/adapterInfoRegistry.go
```diff
@@ -27,47 +28,104 @@ type adapterInfoRegistry struct {
...
+// github.com/user1/mixerAdapters/denier --> denier-mixeradapters-user1
+// metricsco.io/mixerAdapters/metrics1 --> metrics1-mixeradapters-metricso-io
+// metricsco.io/mixerAdapters/adapters/metrics2 --> metrics2-mixeradapters-metricsco-io
+// kind = $pkg-$reponame-$org
+// Resulting name is DNS-1035 compatible. [a-z]([-a-z0-9]*[a-z0-9])?
+func NameToCfgKind(adapterName string) (string, error) {
```
geeknoid: I don't like the way this works.

I don't think we should treat github specially, we shouldn't have any special cases here. Just take the incoming slash-based name, flip it and dashify it.

pkg/adapter/info.go
geeknoid: Let's not add this field. We should just systematically derive the short name, no need to put the burden on the developer. Even if it is optional, the field would show up in docs. Since we can derive it, let's just do that all the time. We can always introduce this field later is there's a reason to allow the developer to control this name.

And in addition, I don't think calling this "ShortName" is correct. The name is actually just as long. ResourceTypeName or something may be more appropriate.

mandarjog:
abandoned in favor of #1145

PR #1120, opened 2017-08-22
Remove unnecessary methods from Store2, and moves handling of protobuf to `store`

jmuk:
**Release note**:

```release-note
none
```

This introduces `Store2Backend`, and keep `pkg/config/crd` untyped (spec is `map[string]interface{}`), and move the conversion part to `proto.Message` into `pkg/config/store`. With this, it is easier to add other types of store like FSstore, the same conversion logic can be reused.

Also, we've agreed that the previous `Store2` interface had too many methods which wouldn't be used actually. This simplifies the interface, as removing `RegisterKind` (integrated into `Init`), removing kinds parameter from `Watch`, and removing of `Put` / `Delete`.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1120)


pkg/config/store/queue.go
```diff
@@ -12,58 +12,74 @@
...
+	ev.Value, err = convertWithKind(spec, ev.Kind, q.kinds)
+	return err
+}
+
 func (q *eventQueue) run() {
 loop:
 	for {
 		select {
 		case <-q.ctx.Done():
+			fmt.Printf("done0\n")
```
mandarjog: fmt? use glog.Info ?

jmuk: oops, it was for my debugging. Removed.

pkg/config/store/store2.go
```diff
@@ -35,43 +36,96 @@ type Key struct {
...
+func (s *Store2) SetValidator(v Validator) {
+	// TODO: implement this
+}
 
-	// Get returns a resource's spec to the key.
-	Get(key Key, spec proto.Message) error
+// Init initializes the connection with the storage backend. This uses "kinds"
+// for the mapping from the kind's name and its structure in protobuf.
+// The connection will be closed after ctx is done.
+func (s *Store2) Init(ctx context.Context, kinds map[string]proto.Message) error {
```
mandarjog: Please make Store2 and interface.

pkg/config/store/convert.go
```diff
@@ -0,0 +1,46 @@
...
+	pbSpec := proto.Clone(base)
+	if err := convert(spec, pbSpec); err != nil {
+		return nil, err
+	}
+	return pbSpec, nil
+}
+
+func convert(spec map[string]interface{}, target proto.Message) error {
+	// This is inefficient; convert to a protobuf message through JSON.
+	// TODO: use reflect.
```
mandarjog: remove comment.
reflect will not work, it will not let us use ENUM symbolic constants.

jmuk: done

pkg/config/store/store2.go
```diff
@@ -35,43 +36,110 @@ type Key struct {
...
 
-	// ListKeys returns the list of all keys in the store.
-	ListKeys() []Key
+// store2 is the implementation of Store2 interface.
+type store2 struct {
+	kinds   map[string]proto.Message
+	backend Store2Backend
+}
+
+// NewStore2 creates a new Store2 instance with the specified backend.
```
mandarjog: We also need an abstract constructor that takes  `NewStore2(storeURL)` ?

jmuk: Done

pkg/config/store/store2.go
```diff
@@ -35,43 +38,143 @@ type Key struct {
 type Event struct {
 	Key
 	Type ChangeType
+
+	// Value refers the new value in the updated event. nil if the event type is delete.
```
mandarjog: The Value should be proto.Message, right ?

jmuk: Done

pkg/config/crd/store_test.go
```diff
@@ -16,243 +16,180 @@ package crd
...
+		watchers: map[string]*watch.FakeWatcher{},
 	}
-	store, err := NewStore(kubeconfig(t), []string{ns})
-	if err != nil {
-		deleteNamespace(k8sclient, ns)
-		t.Fatal(err.Error())
+	client := &Store{
+		conf:                 &rest.Config{},
+		discoveryBuilder:     createFakeDiscovery,
+		listerWatcherBuilder: lw,
```
mandarjog: aah ok, so listWatcher is for test injection.

pkg/config/store/queue.go
```diff
@@ -12,46 +12,65 @@
...
 
-func newQueue(ctx context.Context, cancel context.CancelFunc) *eventQueue {
+func newQueue(ctx context.Context, chin <-chan BackendEvent, kinds map[string]proto.Message) *eventQueue {
 	eq := &eventQueue{
-		ctx:    ctx,
-		cancel: cancel,
-		chout:  make(chan store.Event),
-		chin:   make(chan store.Event),
+		ctx:   ctx,
+		chout: make(chan Event),
```
mandarjog: why is the `out chan` blocking chan? no length specified.

pkg/config/store/store2.go
```diff
@@ -31,47 +34,154 @@ type Key struct {
...
+}
+
+// Get returns a resource's spec to the key.
+func (s *store2) Get(key Key, spec proto.Message) error {
+	obj, err := s.backend.Get(key)
+	if err != nil {
+		return err
+	}
+
+	return convert(obj, spec)
```
mandarjog: obj is `map[string]interface{}` right?

pkg/config/store/store2.go
```diff
@@ -31,47 +34,154 @@ type Key struct {
...
+}
+
+// List returns the whole mapping from key to resource specs in the store.
+func (s *store2) List() map[Key]proto.Message {
+	data := s.backend.List()
+	result := make(map[Key]proto.Message, len(data))
+	for k, spec := range data {
+		pbSpec, err := convertWithKind(spec, k.Kind, s.kinds)
```
mandarjog: since we have `s.kinds` having `convert` and `converWithKind` methods is unnecessary.
you could access the map in the call.

mandarjog:
Over all the PR looks good, 2 things though.

1. Please add one paragraph of design notes
2. Please increase the code coverage?  store2.go, store.go and resource.go have especially low code coverage.

pkg/config/crd/store_test.go
jmuk: yes.

pkg/config/store/store2.go
jmuk: yes

pkg/config/store/store2.go
jmuk: The main point here is to create a clone of the message from `s.kinds`, and then convert `spec` into the clone (otherwise `s.kinds` will be affected). Also, the same logic is used in `queue.go`.

Maybe this structure wasn't clear. removed `convertWithKind`, create `cloneMessage` utility instead.

jmuk:
Done. Also added tests. I think this improves the coverage (I'm afraid that `store2.go`'s coverage won't be very high, since it still has some code path used to connect with real k8s servers).

Updated the PR description to describe the overview of this change.

mandarjog:
@sebastienvas code coverage numbers are not being updated. how do we fix it ?

jmuk:
Per offline discussion, a new file `init.go` was created in `pkg/config/crd` to scope the code that requires actual kubernetes services for initialization.

Here's the local result of the code coverage:
```
pkg/config/crd/init.go: 0%
pkg/config/crd/resource.go: 100%
pkg/config/crd/store.go: 91.4%
pkg/config/store/convert.go: 100%
pkg/config/store/queue.go: 92.6%
pkgconfig/store/store2.go: 100%
```

pkg/config/store/queue.go
jmuk: added

sebastienvas:
@mandarjog Which numbers ? the one from codecov or the one in pkg_check. cc @yutongz

PR #1128, opened 2017-08-23
Introduce Mixer runtime Controller

mandarjog:
This is the 3rd PR to enable new Mixer adapter dispatch.
It introduces Mixer Controller.

Controller is responsible for watching configuration using the Store2 API.
Controller produces a resolver and installs it in the dispatcher.
Controller consumes potentially inconsistent configuration state from the config store and produces a consistent snapshot.
Controller must not panic on configuration problems, it should issues a warning and continue.

Note:
This PR does not have tests yet, I will be adding them.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1128)


pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+	// Handler is the initialized handler object.
+	Handler adapter.Handler
+
+	// HandlerCreateError records error while creating the handler
+	HandlerCreateError error
+
+	// Instances is the global list of instances associated with this handler
+	Instances map[string]bool
+
+	// sha is used to verify and update the handlerEntry.
```
guptasu: I think the code is already doing that.. might have to update the comment to clarify that sha is not just from handler param

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+	// For example prometheus adapter initialized objects based on this configuration.
+	// If new configInstances are associated with such a handler, it also necessitates
+	// recreation of the handler.
+	sha [sha1.Size]byte
+
+	// If this is set, the handler is closed.
+	notInUse bool
+}
+
+func newHandlerTable(instanceConfig map[string]*cpb.Instance, handlerConfig map[string]*cpb.Handler,
```
guptasu: are the args all instName -> cpb.Instances ? and handlerName -> HandlarConfig  ? So, if this HandlerTable is instantiate at init time, how does these values get changes upon config change ?

guptasu: I guess my confusion is: The above comment mentioned that HandlerTable created at startup,  But this initialization takes a specific configuraiton state (instances, handlerConfig). I was hoping this code would just instantiate "map[string]*HandlerEntry". May be I am missing the idea here.

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+		instanceConfig: instanceConfig,
+		handlerConfig:  handlerConfig,
+		buildHandler:   buildHandler,
+	}
+}
+
+// handlerTable initializes and maintains handlers.
+// It can efficiently initialize the next version of handlers based
+// on the previous handler table.
+type handlerTable struct {
```
guptasu: Does handlerTable need to be threadSafe ?

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+// Initialize the handler table based on configuration and the old handler table.
+// If handler config has not changed, connections from the old handler table are re-used.
+// This method does not return an error, it records errors in the handlerEntry.
+func (t *handlerTable) Initialize(oldTable map[string]*HandlerEntry) {
+	t.computeSha()
+	// run diff with the old handlerTable
+	for oh, ohe := range oldTable {
+		var he *HandlerEntry
+		if he = t.table[oh]; he == nil {
+			// old handler is not needed any more
```
guptasu: It is unclear what does it mean for the HandlerEntry for a handler in the handlerTable is nil ?

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+			Name:      handleName,
+			Instances: make(map[string]bool),
+		}
+		t.table[handleName] = h
+	}
+	h.Instances[instanceName] = true
+}
+
+// Initialize the handler table based on configuration and the old handler table.
+// If handler config has not changed, connections from the old handler table are re-used.
```
guptasu: If handler config has not changed or the associated instances has not changed.. I know we want to fix this in future.

pkg/runtime/handlerTable.go
guptasu: I think there will be two HandlerTables that you will maintain and just initialize the new one base done the old one. If that is correct, then I understood what you are doing. Considering adding comment to clarify lifecycle of the HandlerTable.

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// initialize handlers that were not previously covered.
+	for _, he := range t.table {
+		// this was already initialized.
+		if he.Handler != nil {
+			continue
+		}
+		// create a new handler
+		// handler error is marked inside the entry.
+		t.initHandler(he)
```
guptasu: nit : Should we just do this step in the above loop...when we know the sha has not matched .. just create handler object.

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+	// sha must use handler.Params since any change to Params
+	// requires creation of a new handler instance.
+	// Some handlers need to be told about data they are about to receive.
+	// For example prometheus adapter initialized objects based on this configuration.
+	// If new configInstances are associated with such a handler, it also necessitates
+	// recreation of the handler.
+	sha [sha1.Size]byte
+
+	// If this is set, the handler is closed.
+	notInUse bool
```
guptasu: Why do we need this .. resolver  was managing the cleanup right ?

pkg/runtime/handlerTable.go
```diff
@@ -0,0 +1,178 @@
...
+// handlerEntries can be replaced.
+type HandlerEntry struct {
+	// Name of the handler
+	Name string
+
+	// Handler is the initialized handler object.
+	Handler adapter.Handler
+
+	// HandlerCreateError records error while creating the handler
+	HandlerCreateError error
```
guptasu: What is this error used for ?

pkg/runtime/controller.go
```diff
@@ -0,0 +1,465 @@
...
+	// `Data` could be in an inconsistent state.
+	configState map[store.Key]proto.Message
+
+	// currently deployed resolver
+	resolver *resolver
+
+	// tableCache maps resolverId to handler table.
+	//tableCache map[int]map[string]*HandlerEntry
+	// table is the handler state currently in use.
+	table map[string]*HandlerEntry
```
guptasu: comment is not correct to the field name..

pkg/runtime/controller.go
```diff
@@ -0,0 +1,465 @@
...
+	ChangeResolver(rt Resolver)
+}
+
+// RulesKind defines the config kind name of mixer rules.
+const RulesKind = "mixer-rules"
+
+// New creates a new runtime Dispatcher
+// Create a new controller and a dispatcher.
+// Returns a ready to use dispatcher.
+func New(eval expr.Evaluator, gp *pool.GoroutinePool, handlerPool *pool.GoroutinePool,
```
guptasu: NewController

jmuk: NewDispatcher?  At least it returns a Dispatcher instance.

pkg/runtime/controller.go
```diff
@@ -0,0 +1,399 @@
...
+		case store.Delete:
+			delete(c.configState, ev.Key)
+		}
+	}
+	c.publishSnapShot()
+}
+
+// filterInstanceConfig filters the configState and produces instanceConfig
+// that points to valid templates.
+func (c *Controller) filterInstanceConfig() map[string]*cpb.Instance {
```
guptasu: IMO, this mtd can be called getInstances(). Filter seems more powerful than just ignoring bad reference instances. Same argument for similar other mtds

pkg/runtime/controller.go
```diff
@@ -0,0 +1,399 @@
...
+			Name:    k.Name,
+			Adapter: k.Kind,
+			Params:  cfg,
+		}
+	}
+	return handlerConfig
+}
+
+// convertToRuntimeRules converts internal rules to the format that resolver needs.
+func convertToRuntimeRules(ruleConfig map[string]map[string]*Rule) map[string][]*Rule {
```
guptasu: Consider type aliasing  and adding comment to what this map represents

pkg/runtime/controller.go
```diff
@@ -0,0 +1,399 @@
...
+		rule.actions = ruleActions
+		rn[k.Name] = rule
+	}
+
+	return ruleConfig
+}
+
+const cleanupSleepTime = 500 * time.Millisecond
+
+// processActions prunes actions that lack referential integrity and associate instances with
```
guptasu: Are we not doing ref identity check at all ? The classic vaidator should be running before this code right ?

pkg/runtime/handlerTable.go
mandarjog: While combining rules with initialized handlers, This error is logged along with the rules that were filtered out.

pkg/runtime/handlerTable.go
mandarjog: Updated comments.

pkg/runtime/handlerTable.go
mandarjog: No. All this code runs in config change control loop, which is single threaded.

pkg/runtime/handlerTable.go
mandarjog: updated comment.

pkg/runtime/handlerTable.go
mandarjog: It means the new configuration does not have a handler named `oh`.
Therefore that entry in the oldTable is marked for cleanup.

pkg/runtime/handlerTable.go
mandarjog: We cannot do it there. The purge loop runs on the  *oldTable* and copies over handler that can be re-used.  
The 2nd loop runs on the *newTable*

pkg/runtime/logger.go
```diff
@@ -0,0 +1,51 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runtime
+
+import (
+	"errors"
```
mandarjog: done

pkg/runtime/resolver_test.go
```diff
@@ -125,7 +125,7 @@ func TestResolver_Resolve(t *testing.T) {
 			rules := newRules(tc.variety, tc.rules)
 			bag := attribute.GetFakeMutableBagForTesting(tc.bag)
 			eval := fakePred(tc.selectReject, tc.selectError)
-			rv := NewResolver(eval, ia, ns, rules, nil)
+			var rv Resolver = newResolver(eval, ia, ns, rules, 1)
```
mandarjog: to force variable type to be an interface.

pkg/runtime/resolver.go
```diff
@@ -59,25 +59,23 @@ type resolver struct {
...
-	// After the resolver has no references, this state can be
-	// cleaned up correctly by the caller.
-	ruleState interface{}
-
 	// refCount tracks the number requests currently using this
 	// configuration. resolver state can be cleaned up when this count is 0.
 	refCount int32
+
+	// ID of the resolver for debugging.
+	ID int
```
mandarjog: Resolver ID defines the monotonically increasing ID for the config snapshot. 
It is used in logging now. I think it is useful to know it in the logs.

pkg/runtime/handlerTable.go
mandarjog: This where we record that **this** entry needs to be cleaned up by `close()` ing the handler.

pkg/runtime/handlerTable.go
mandarjog: Updated comments.

pkg/runtime/controller.go
ZackButcher: No, `New` is correct. `runtime.New` is not ambiguous.

pkg/runtime/tempstore.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2017 Istio Authors
```
jmuk: Please remove this file, my PR was merged now :wink:

pkg/api/grpcServer.go
```diff
@@ -52,14 +54,15 @@ type (
 )
 
 // NewGRPCServer creates a gRPC serving stack.
-func NewGRPCServer(aspectDispatcher adapterManager.AspectDispatcher, gp *pool.GoroutinePool) mixerpb.MixerServer {
+func NewGRPCServer(aspectDispatcher adapterManager.AspectDispatcher, dispatcher2 runtime.Dispatcher, gp *pool.GoroutinePool) mixerpb.MixerServer {
```
manlinl: Can we call dispatcher2 just dispatcher since there isn't naming conflict.

pkg/api/grpcServer.go
```diff
@@ -42,6 +43,7 @@ import (
 type (
 	// grpcServer holds the dispatchState for the gRPC API server.
 	grpcServer struct {
+		dispatcher2      runtime.Dispatcher
```
manlinl: Can we just call it dispatcher since no naming conflict

pkg/runtime/dispatcher.go
```diff
@@ -121,15 +121,13 @@ type dispatcher struct {
 	resolver     Resolver
 }
 
-// SetResolver installs a new resolver.
+// ChangeResolver installs a new resolver.
 // This function is called when configuration is updated by the user.
 // oldResolver is returned so that it can be reclaimed.
-func (m *dispatcher) SetResolver(rt Resolver) Resolver {
+func (m *dispatcher) ChangeResolver(rt Resolver) {
```
manlinl: SetResolver appears to be clearer about what the function is. ChangeResolver sounds like a single word.

pkg/runtime/handler.go
```diff
@@ -32,13 +32,17 @@ type (
...
+		GetTemplateInfo(template string) (template.Info, bool)
+	}
 	// HandlerFactory builds adapter.Handler object from adapter and instances configuration.
 	HandlerFactory interface {
 		Build(*pb.Handler, []*pb.Instance, adapter.Env) (adapter.Handler, error)
 	}
 
 	handlerFactory struct {
-		tmplRepo          template.Repository
+		tmplRepo          TemplateFinder
```
manlinl: Rename tempRepo to match the new type?

pkg/api/grpcServer.go
manlinl: Can we just call it dispatcher? so don't need to go back to clean up later. The same for other dispatcher2.

pkg/runtime/tempstore.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/golang/protobuf/proto"
+
+	"istio.io/mixer/pkg/config/store"
+)
+
+// Store abstraction used here.
+// This file is temporary.
+// It file will be removed when config.store.Store2 interface is added.
```
manlinl: typo.

pkg/runtime/resolver.go
manlinl: It probably should be called version or sequenceNumber.

pkg/runtime/resolver.go
```diff
@@ -59,25 +59,23 @@ type resolver struct {
...
 	rules map[string][]*Rule
 
-	// Opaque ruleState used in this configuration
-	// After the resolver has no references, this state can be
-	// cleaned up correctly by the caller.
-	ruleState interface{}
-
 	// refCount tracks the number requests currently using this
 	// configuration. resolver state can be cleaned up when this count is 0.
 	refCount int32
```
manlinl: Not related to this PR. Do you need make refCount atomic?

pkg/runtime/controller.go
```diff
@@ -0,0 +1,420 @@
...
+	configState map[store.Key]proto.Message
+
+	// currently deployed resolver
+	resolver *resolver
+
+	// table is the handler state currently in use.
+	table map[string]*HandlerEntry
+
+	// dispatcher is notified of changes.
+	dispatcher ResolverChangeListener
```
manlinl: Too many dispatcher, maybe just changeListener?

ZackButcher: `listener`

pkg/runtime/resolver.go
ZackButcher: It's only modified with `atomic.AddInt32` methods, so it's fine. `atomic.Value` doesn't help us here because a read-then-write of the value is not atomic (just the access; you're guaranteed to get some consistent version of the value), which is what we need here.

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+			name:     k.Name,
+		}
+		acts := c.processActions(rulec.Actions, handlerConfig, instanceConfig, ht)
+
+		rn := ruleConfig[k.Namespace]
+		if rn == nil {
+			rn = make(map[string]*Rule)
+			ruleConfig[k.Namespace] = rn
+		}
+		ruleActions := make(map[adptTmpl.TemplateVariety][]*Action)
```
guptasu: Considering moviing this actions stuff (next 5 lines) up right after the acts are computed..

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+	return rules, nrules
+}
+
+// processRules builds the current consistent view of the rules keyed by Namespace and then Name.
+// ht (handlerTable) keeps track of handler-instance association.
+func (c *Controller) processRules(handlerConfig map[string]*cpb.Handler,
+	instanceConfig map[string]*cpb.Instance, ht *handlerTable) rulesMapByNamespace {
+	// current consistent view of the rules
+	// keyed by Namespace and then Name.
+	ruleConfig := make(rulesMapByNamespace)
```
guptasu: Config to me seems more of a proto object... here this is a fully resolved Rule instance..
Suggestion : Renamed method to getResolvedRules and this variable to resolvedRules

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+		rn[k.Name] = rule
+	}
+
+	return ruleConfig
+}
+
+var cleanupSleepTime = 500 * time.Millisecond
+
+// processActions prunes actions that lack referential integrity and associate instances with
+// handlers that are later used to create new handlers.
```
guptasu: Consider a comment..: The reference to the built handler objects is not done in this method. it is done later in combineActionsToHandler done after handlers are built.

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+			}
+			act.instanceConfig = append(act.instanceConfig, inst)
+		}
+	}
+	return actions
+}
+
+// combineRulesHandlers sets handler references in rulesConfig.
+// Reject actions from rulesConfig whose handler could not be initialized.
+func combineRulesHandlers(ruleConfig rulesMapByNamespace, handlerTable map[string]*HandlerEntry) (rulesListByNamespace, int) {
```
guptasu: RefActionsToHandlers ?

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+				if glog.V(3) {
+					glog.Warningf("ConfigWarning unknown instance: %s", instName)
+				}
+				continue
+			}
+
+			ht.Associate(ic.Handler, instName)
+
+			ti := c.templateInfo[inst.Template]
+			// grouped by adapterName.
```
guptasu: Move this comment below

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+
+	return ruleConfig
+}
+
+var cleanupSleepTime = 500 * time.Millisecond
+
+// processActions prunes actions that lack referential integrity and associate instances with
+// handlers that are later used to create new handlers.
+func (c *Controller) processActions(acts []*cpb.Action, handlerConfig map[string]*cpb.Handler,
+	instanceConfig map[string]*cpb.Instance, ht *handlerTable) map[adptTmpl.TemplateVariety]map[string]*Action {
```
guptasu: Why do we need to group by adapterName ?

pkg/runtime/controller.go
```diff
@@ -0,0 +1,424 @@
...
+
+			ti := c.templateInfo[inst.Template]
+			// grouped by adapterName.
+			vAction := actions[ti.Variety]
+			if vAction == nil {
+				vAction = make(map[string]*Action)
+				actions[ti.Variety] = vAction
+			}
+
+			act := vAction[hc.Adapter]
```
guptasu: I am not sure I understand : Why are we merging instances across multiple actions that go to the same adapter.. It seems we are only creating one Action object for all the actions that have the same templateVariety and same AdapterName and merging their instances.. I might be missing something, but from the top level it seems incorrect.

pkg/runtime/init.go
```diff
@@ -0,0 +1,115 @@
...
+	"istio.io/mixer/pkg/template"
+)
+
+// This file contains code to create new objects that are
+// of package wide interest.
+
+// New creates a new runtime Dispatcher
+// Create a new controller and a dispatcher.
+// Returns a ready to use dispatcher.
+func New(eval expr.Evaluator, gp *pool.GoroutinePool, handlerPool *pool.GoroutinePool,
```
ZackButcher: No, this is not Java.

guptasu: Well this method is in runtime package but returns Dispatcher.. it could very well return Controller ore Resolver or anything.. New is just confusing for this package atleast because it exposes many other public Types

pkg/runtime/controller.go
```diff
@@ -0,0 +1,420 @@
...
+		}
+	}
+}
+
+// applyEvents applies given events to config state and then publishes a snapshot.
+func (c *Controller) applyEvents(events []*store.Event) {
+	for _, ev := range events {
+		switch ev.Type {
+		case store.Update:
+			c.configState[ev.Key] = ev.Value
```
jmuk: I don't think so. applyEvents is called from watchChanges, within a synchronous for-loop. No other applyEvents are called at the same time.
(and the use of configState is within publishSnapshot, which is only called from applyEvents and before watchChanges goroutine starts).

pkg/runtime/resolver.go
jmuk: If it's for debugging and logging, maybe it should be hidden (lower-cased)?

pkg/runtime/init.go
```diff
@@ -0,0 +1,115 @@
...
+	kindMap := kindMap(adapterInfo, templateInfo)
+	if err := s.Init(ctx, kindMap); err != nil {
+		return nil, nil, err
+	}
+	// create channel before listing.
+	watchChan, err := s.Watch(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	return s.List(), watchChan, nil
```
jmuk: s.List returns the all data mapping from key to the value in the store. The name matches with the existing store1. This means to return the initial data and a channel for further updates.

jmuk: BTW, strictly speaking there is a chance that an event arrives to the channel between line 58 and 62. In that case watchChan will report the data which is already in the initial data. This wouldn't be harmful though.

If that helps, maybe Store2's Watch should return the initial data along with the channel instead of using List explicitly -- which was similart to the design of Galley API.

pkg/runtime/controller.go
mandarjog: D

pkg/runtime/controller.go
mandarjog: This is just establishing association between handler and instances.
see `ht.Associate()` call.

pkg/runtime/controller.go
mandarjog: good catch. It should actually be (handler/template) per runtime.Action specs.

pkg/runtime/controller.go
mandarjog: done.

pkg/runtime/controller.go
mandarjog: good catch. see comments above.

pkg/runtime/controller.go
mandarjog: D

pkg/runtime/controller.go
```diff
@@ -0,0 +1,420 @@
...
+			}
+		}
+	}
+}
+
+// cleanupResolver cleans up handler table in the resolver
+// after the resolver is no longer in use.
+func cleanupResolver(r *resolver, table map[string]*HandlerEntry, timeout time.Duration) error {
+	start := time.Now()
+	for {
```
mandarjog: We are not really swapping the value here. 
We are polling the refcount in an atomic way until a timeout. This code should only read the value. CAS does not apply here.

pkg/runtime/handlerTable.go
mandarjog: not sure what you mean. ohe - old handler entry.

pkg/runtime/init.go
mandarjog: The `New` method does return a dispatcher, but it does more than that. 
It creates a controller, creates a dispatcher, associates one with the other, and fetches initial configuration. 

In effect this is creating new `runtime`.

pkg/runtime/init.go
mandarjog: Start watch returns currrent data and a watch channel that can be consumed later.
S.List() is the call that returns data. we cant return kindMap here.

pkg/runtime/resolver.go
mandarjog: D

pkg/runtime/controller.go
ZackButcher: Yeah, I wasn't thinking :)

mandarjog:
PTAL

pkg/runtime/controller.go
```diff
@@ -326,31 +330,34 @@ func (c *Controller) processActions(acts []*cpb.Action, handlerConfig map[string
 			ht.Associate(ic.Handler, instName)
 
 			ti := c.templateInfo[inst.Template]
-			// grouped by adapterName.
 			vAction := actions[ti.Variety]
```
guptasu: Clarification, for my understanding.. It seems along with splitting, we are also merging across actions... 
- instances across two actions with same handler, template and variety will be collapsed into one action... 
- And instances within one action with with different templatename or different variety will be split into separate actions.

Right ? Consider adding a comment about the merging logic

pkg/runtime/BUILD.bazel
```diff
@@ -3,10 +3,16 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "go_default_library",
     srcs = [
+        "controller.go",
```
guptasu: I agree, nothing for this PR. However, after 0.2, may be we can move the control loop related codebase into a separate pkg..

pkg/runtime/controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+// maxEvents is the likely maximum number of events
+// we can expect in a second. It is used to avoid slice reallocation.
+const maxEvents = 50
+
+// publishSnapShot converts the currently available configState into a resolver.
+// The config may be in an inconsistent state, however it *must* be converted into a consistent resolver.
+// The previous handler table enables handler cleanup and reuse.
+// This code is single threaded, it only runs on a config change control loop.
+func (c *Controller) publishSnapShot() {
```
guptasu: Consider adding a comment explaining this. Thanks

pkg/runtime/controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+	// Combine rules with the handler table.
+	// Actions referring to handlers in error, will be purged.
+	combineRulesHandlers(ruleConfig, ht.table)
+
+	// create rules that are in the format that resolver needs.
+	rules, nrules := convertToRuntimeRules(ruleConfig)
+
+	// Create new resolver and cleanup the old resolver.
+	c.resolverID++
```
guptasu: Sorry, I meant, I am not sure how will resolveId be used... It is mainly for testing in order to inspect the value ?

pkg/runtime/controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+var watchFlushDuration = time.Second
+
+// watchChanges watches for changes and publishes a new resolver.
+// watchChanges is started in a goroutine.
+func watchChanges(wch <-chan store.Event, applyEvents applyEventsFn) {
+	// consume changes and apply them to data indefinitely
+	var timeChan <-chan time.Time
+	var timer *time.Timer
+	events := make([]*store.Event, 0, maxEvents)
```
guptasu: append would panic right in case someone maliciously (or may be a bug in config server) causes the events to happen faster than we can handle.

pkg/runtime/controller.go
guptasu: I see now.

pkg/runtime/controller.go
guptasu: SG

PR #1130, opened 2017-08-24
fsstore backend for Store2

jmuk:
**Release note**:

```release-note
fsstore: "fs://" store backend is added to Store2 too. Helpful for e2e tests.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1130)


pkg/config/store/contextch.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// ContextChList manages the list of ContextCh.
+type ContextChList struct {
+	mu   sync.RWMutex
+	list []*contextCh
+}
+
+// NewContextChList creates a new list of ContextCh.
+func NewContextChList() *ContextChList {
```
ZackButcher: No need for a new method, delete and replace all occurrences with `&ContextChList{}`.

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,185 @@
...
+		delete(s.data, key)
+		evs = append(evs, BackendEvent{Key: key, Type: Delete})
+	}
+	for _, ev := range evs {
+		s.chs.Send(ev)
+	}
+}
+
+// NewFsStore2 creates a new FsStore2 instance.
+func NewFsStore2(root string) *FsStore2 {
```
ZackButcher: Shouldn't this return a `Store2Backend` instead? Do we really need the FsStore type to be public - it doesn't export any methods that aren't part of the `Store2Backend` interface.

pkg/config/store/contextch.go
```diff
@@ -0,0 +1,71 @@
...
+package store
+
+import (
+	"context"
+	"sync"
+)
+
+// contextCh is a utility type of outgoing channel with context. It can send messages
+// as long as the context is valid.
+type contextCh struct {
```
ZackButcher: I'm a bit confused about why we need this type: you can just close a channel, I don't think we need a context to handle closing it too.

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+	if len(updated) == 0 && len(removed) == 0 {
+		return
+	}
+	s.shas = newShas
+	evs := make([]BackendEvent, 0, len(updated)+len(removed))
+	for _, key := range updated {
+		parsed := map[string]interface{}{}
+		glog.Errorf("%+v", key)
```
mandarjog: why is this an error?

pkg/config/store/contextch.go
jmuk: ah, right... removed.

pkg/config/store/contextch.go
jmuk: No, closing channels won't help. Sending on closed channel causes a runtime panic (https://golang.org/ref/spec#Send_statements), that's not wanted here. Here want to skip sending to that channel (ignore it) in that case.

pkg/config/store/fsstore2.go
jmuk: my bad, not sure why this is here. Maybe for some printf-debugging?? removed. Thanks for catching.

pkg/config/store/fsstore2.go
jmuk: done

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,184 @@
...
+	result := map[Key][]byte{}
+
+	for _, kind := range s.kinds {
+		root := filepath.Join(s.root, kind)
+		err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if strings.HasSuffix(path, suffix) && (info.Mode()&os.ModeType) == 0 {
+				data, err := ioutil.ReadFile(path)
```
mandarjog: I see what you are doing here, however what we really want is to consume the same artifacts that we will send via kubectl create -f

1. each yaml file with `---` as a separator. 
2. Each element is full k8s doc.
```
apiVersion: config.istio.io/v1alpha2
kind: Metric
metadata:
  name: requestcount
  namespace: default
spec:
  description: request count by source, target, service, and code
  kind: COUNTER
  labels:
    method: 1
    response_code: 2
    service: 1
    source: 1
    target: 1
    version: 1
  value: INT64
``` 
This way we can test with the same files.

pkg/config/store/contextch.go
ZackButcher: I really think the ownership semantics are wrong here, and the watching for a channel closure is happening in the wrong spot. The worker needs to be signaled to stop listening, but instead you're trying to have the sender figure out when to stop pushing data.

----------

At a minimum use the `Done() <-chan struct{}` in the context during construction, not the context itself: you don't need or use it, and it's an odd pattern. Having two channels, one to receive data and one to signal finished, is totally normal in go. What's odd is having a channel and a context together that you use to accomplish the same thing.

```go
type (
    receiver struct {
        events chan<- BackendEvent
        done   <-chan Struct{}
    }

    receivers struct {
        m  sync.RWMutex
        ls []*listener 
    }
)

func (rs *receivers) Send(e BackendEvent) {
	rs.m.RLock()
 	for _, r := range rs.list {
 		select {
 		case <-r.done:
 		case r.events <- ev:
 		}
 	}
	rs.mu.RUnlock()
}
```

However, even then I'm not sure this is correct: if you have a select over two channels and both are valid (i.e. both the events channel is open to receive an event and the context is done) the channel used in the select is picked pseudo randomly, so it's possible that you still attempt the send and panic as a result.

jmuk:
PTAL

Updated the implementation to follow what @mandarjog is saying.

Also, due to offline discussion with @mandarjog, contextch.go file is removed. That's not necessary anymore with an assumption that a store supports a single watch channel at the same time.

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+const defaultDuration = time.Second / 2
+
+type resourceMeta struct {
+	Name      string
+	Namespace string
+}
+
+type resource struct {
```
mandarjog: please note that this is identical to crd/resource.

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,228 @@
...
+	checkDuration time.Duration
+	watchCtx      context.Context
+	watchCh       chan BackendEvent
+
+	mu   sync.RWMutex
+	data map[Key]*resource
+}
+
+var _ Store2Backend = &fsStore2{}
+
```
mandarjog: Add comment that path is only used for error reporting.

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		glog.Errorf("failure during filepath.Walk: %v", err)
+	}
+	return result
+}
+
+func (s *fsStore2) update() {
```
mandarjog: Rename:  `checkForUpdates()`

pkg/config/store/fsstore2.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	for _, ev := range evs {
+		select {
+		case <-s.watchCtx.Done():
+		case s.watchCh <- ev:
+		}
+	}
+}
+
+// NewFsStore2 creates a new Store2Backend backed by filesystem.
```
ZackButcher: nit: "backed by a filesystem", or "backed by the filesystem"

pkg/config/store/store2.go
```diff
@@ -27,6 +28,9 @@ import (
 // ErrNotFound is the error to be returned when the given key does not exist in the storage.
 var ErrNotFound = errors.New("not found")
 
+// ErrWatchAlreadyExist is the error to report that the watching channel already exists.
+var ErrWatchAlreadyExist = errors.New("watch already exist")
```
ZackButcher: nit: Exist->Exist**s**

mandarjog:
Apart from nits... would be good to fix during rebase

/lgtm

mandarjog:
/approve cancel

Apparently I need to hold of lgtm until the nits are resolved.

pkg/config/store/fsstore2.go
jmuk: This does update the content too. `checkAndUpdate` might be better?

pkg/config/store/fsstore2.go
jmuk: done

pkg/config/store/fsstore2.go
jmuk: done

pkg/config/store/fsstore2.go
jmuk: done

pkg/config/store/store2.go
jmuk: done

PR #1136, opened 2017-08-24
Update statsd to new adapter form

ZackButcher:
```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1136)

### DIFF
--- a/adapter/statsd2/config/config.proto
+++ b/adapter/statsd2/config/config.proto
@@ -42,14 +42,27 @@ message Params {
     // Chance that any particular metric is sampled when incremented; can take the range [0, 1], defaults to 1 if unspecified.
     float sampling_rate = 5;
 
-    // Map of metric name -> template string; the template will be filled with values from the metric's labels and the
-    // resulting string will be used as the statsd metric name. This allows easier creation of statsd metrics like
-    // `action_name-response_code`. The template strings must conform to go's text/template syntax. For the example of
-    // `action_name-response_code`, we use the template:
-    //    `{{.apiMethod}}-{{.responseCode}}`
-    //
-    // If a metric's name is not in the map then the exported statsd metric name will be exactly the metric's name.
-    //
-    // TODO: unify templates to use our expression language when its finalized
-    map<string, string> metric_name_template_strings = 6;
+    // Describes how to represent this metric in statsd
+    message MetricInfo {
+        // The type of metric.
+        enum Type {
+            UNKNOWN = 0;
+            COUNTER = 1;
+            GAUGE = 2;
+            DISTRIBUTION = 3;
+        }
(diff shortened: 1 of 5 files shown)


adapter/statsd2/statsd.go
```diff
@@ -0,0 +1,195 @@
...
+
+var (
+	_ metric.HandlerBuilder = &builder{}
+	_ metric.Handler        = &handler{}
+)
+
+// GetBuilderInfo returns the BuilderInfo associated with this adapter implementation.
+func GetBuilderInfo() adapter.BuilderInfo {
+	return adapter.BuilderInfo{
+		Name:        "istio.io/mixer/adapter/statsd2",
```
geeknoid: Can just put statds so we don't forget when doing the file-level rename.

adapter/statsd2/statsd.go
```diff
@@ -0,0 +1,195 @@
...
+		if !found {
+			env.Logger().Infof("template registered for nonexistent metric '%s'", metricName)
+			continue // we don't have a metric that corresponds to this template, skip processing it
+		}
+
+		var t *template.Template
+		if s.NameTemplate != "" {
+			t, _ = template.New(metricName).Parse(s.NameTemplate)
+			if err := t.Execute(ioutil.Discard, def.Dimensions); err != nil {
+				env.Logger().Warningf(
```
geeknoid: Having mistyped stuff in the template seems like a very common/likely problem. Can we catch this in validateConfig instead so the user gets the error at the right place?

adapter/statsd2/config/config.proto
```diff
@@ -0,0 +1,68 @@
...
+option (gogoproto.gostring_all) = false;
+
+message Params {
+    // Address of the statsd server, e.g. localhost:8125
+    string address = 1;
+
+    // Metric prefix, do not specify for no prefix
+    string prefix = 2;
+
+    // Flush Interval controls the maximum amount of time between sending metrics to the statsd collection server.
```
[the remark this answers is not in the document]
ZackButcher: Done

adapter/statsd2/config/config.proto
```diff
@@ -0,0 +1,68 @@
...
+    // Metrics are reported when either flush_bytes is full or flush_interval time has elapsed since the last report.
```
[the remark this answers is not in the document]
ZackButcher: Done

adapter/statsd2/statsd.go
ZackButcher: We do already - lines 90-94. We can remove the check here entirely, but that felt a little off.

ZackButcher:
PTAL

PR #1137, opened 2017-08-24
Retry fetching CRD definitions if it's not ready.

### LINKED ISSUE #1117: CRD Store.Init() should not return error for retryable conditions
CRD Store.Init() should not return error for retryable conditions.

If store is unable to watch for a "kind" it should continue to wait for a long time, and even after the timeout, it should wait to register the watch in the background.

Everything is eventually consistent, so Init should not throw an error.

jmuk:
fixes #1117

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1137)


pkg/config/crd/store.go
```diff
@@ -98,25 +104,43 @@ func (s *Store) Init(ctx context.Context, kinds []string) error {
...
+				informer := cache.NewSharedInformer(cl, nil, defaultResyncPeriod)
+				s.caches[res.Kind] = informer.GetStore()
+				informer.AddEventHandler(s)
+				go informer.Run(ctx.Done())
+			}
 		}
 	}
-	time.Sleep(initializationPeriod) // TODO
+	// TODO: remove this time.Sleep, add a better watcher.
+	time.Sleep(initializationPeriod)
```
ZackButcher: Do we need to block the goroutine processing config here? Can we remove the defer and shove it off to another goroutine?

```go
go func() {
    time.Sleep(initializationPeriod)
    cancel()
}()

File: pkg/config/crd/store.go [hunk 2]
```diff
@@ -51,6 +51,12 @@ const (
 	// initial data since it may arrive in any order, and it
 	// would result in temporary invalid status.
 	initializationPeriod = time.Second * 2
+
+	// crdRetryTimeout is the timeout duration to retry initialization
+	// of the caches when some CRDs are missing. The actual timeout
+	// can be customized through the timeout for the context passed to
+	// Init() method.
+	crdRetryTimeout = time.Minute * 5
```
ZackButcher: Do we have data on timing of CRDs? 5 minutes seems like a very very long time to wait. Keep in mind that we do this at Mixer startup before we start serving traffic.

pkg/config/crd/store.go
ZackButcher: The reasoning is that we do all of this stuff at startup, so we're effectively blocking the entire server starting waiting on this.

jmuk:
@guptasu This `pkg/config/crd` is for config store implementation -- the data is stored as custom resources (that's why it is placed under `pkg/config`).  I think this code is not very helpful for your usage -- anyways, this module maintains fetching of resources through k8s client library -- while your part is to build the _definitions_ of those resources, right?

The parts which can be reused for the definitions would be apiGroup / apiGroupVersion etc. If you want to export them, please make the changes.

pkg/config/crd/store.go
ZackButcher: Ah, gotcha. Reason for the sleep wasn't clear, I just assumed it was around the cancel.

I just want to avoid us piling on things that make Mixer startup slow. Once we have health checking implemented (hopefully in the next week or two) then we should skip blocking for this kind of thing at all. We can just report unhealthy until we get the initial configs. Till then this is fine.

mandarjog:
@jmuk what happens when a "kind" does not exist as a kind at all in config, we do not want to block in that case. 

This is possible if you have a brand new mixer build that understand about more "kinds" than are there. Cache initialization should be retried indefinitely with pauses until we can start watching the "kind".

For testing on an over taxed test cluster like ours, we can add large delays before Mixer starts serving traffic.

jmuk:
PTAL

Changes are to address @mandarjog's comment. We had a chat offline, and here's the summary:
- we should retry synchronously for certain time since mixer's deployment could happen at the same time CRDs are pushed. So we should keep the current structure of this PR, synchronously retrying
- we want to customize the duration for synchronous retrying, sometimes (like for e2e tests) the timeout should be longer. Added a new code to support this through query parameter of the config URL. Now params like `k8s://?retry-timeout=2m` specifies the timeout to be 2minutes, for example.
- for missing / non-existing crds; we should create goroutine(s) to wait for them asynchronously, but this is not in this PR. That will be addressed by another PR in the next week.

pkg/config/crd/store_test.go
```diff
@@ -33,6 +34,9 @@ import (
 	"istio.io/mixer/pkg/config/store"
 )
 
+// The "retryTimeout" used by the test.
+const testingRetryTimeout = 10 * time.Millisecond
```
ZackButcher: Can we make this as small as possible (1ms or less, ideally) - this kind of stuff adds up and makes tests take longer gradually over time.

jmuk: The value comes from my local environment, and it would be highly depending on the computing powers, amount of resources, goroutine scheduling, and so on. But I'm afraid that <1ms could be too short.

It's still hard (for me) to say what's the best value here 😢 added a TODO comment.

jmuk: Ah, wait -- I see another way, to use `HasSynced` property (I found `LastSyncResourceVersion` is racy, but `HasSynced` is not fortunately). Rewritten the code for the wait. PTAL.

PR #1147, opened 2017-08-25
Add extern support for IPADDRESS, DURATION and TIME types.

ozevren:
**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1147)

### DIFF
--- a/pkg/il/interpreter/interpreter_test.go
+++ b/pkg/il/interpreter/interpreter_test.go
@@ -15,6 +15,7 @@
 package interpreter
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/il/interpreter/interpreter_test.go
```diff
@@ -2256,8 +2352,33 @@ func runTestProgram(t *testing.T, p *il.Program, test test) {
...
+	if b1Ok != b2Ok {
+		return false
+	}
+	if b1Ok {
+		return bytesEqual(b1, b2)
+	}
+	return a1 == a2
+}
+
+func bytesEqual(b1 []byte, b2 []byte) bool {
```
douglas-reid: `bytes.Equal()` ?

ozevren: Thanks. That is much better.

PR #1151, opened 2017-08-28
Wire dispatcher

mandarjog:
This PR Wires new dispatcher into the grpcServer.
Mixer was tested locally to use a denier adapter and rule to correctly accept and deny requests.
integration tests will follow.

    1. Deliver adapter inventory as data to top level functions.
    2. Add configStore2URL and configDefaultNamespace command line args.
    3. Add debug to store implementation.
    4. Add attribute manifest processing to controller.
    5. Add VocabularyChangeListener to IL evaluator.
    6. Add tesdata/config directory with denier rules.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1151)

### DIFF
--- a/adapter/inventory.go
+++ b/adapter/inventory.go
@@ -47,3 +47,13 @@ func Inventory() []adapter.RegisterFn {
 		noop.Register,
 	}
 }
+
+// InventoryMap converts adapter inventory to a builder map.
+func InventoryMap(inv []adapter.InfoFn) map[string]*adapter.BuilderInfo {
+	m := make(map[string]*adapter.BuilderInfo, len(inv))
+	for _, ai := range inv {
+		bi := ai()
+		m[bi.Name] = &bi
+	}
+	return m
+}
--- a/cmd/server/cmd/root.go
+++ b/cmd/server/cmd/root.go
@@ -50,8 +50,9 @@ func GetRootCmd(args []string, info map[string]template.Info, adapters []adapter
 	_ = fs.Parse([]string{})
 	flag.CommandLine = fs
 
+	// template.NewRepository(info)
 	rootCmd.AddCommand(adapterCmd(printf))
-	rootCmd.AddCommand(serverCmd(template.NewRepository(info), adapters, printf, fatalf))
+	rootCmd.AddCommand(serverCmd(info, adapters, printf, fatalf))
 	rootCmd.AddCommand(crdCmd(info, adapters, printf))
 	rootCmd.AddCommand(shared.VersionCmd(printf))
 
--- a/cmd/server/cmd/server.go
+++ b/cmd/server/cmd/server.go
@@ -49,6 +49,7 @@ import (
 	"istio.io/mixer/pkg/expr"
 	"istio.io/mixer/pkg/il/evaluator"
 	"istio.io/mixer/pkg/pool"
+	mixerRuntime "istio.io/mixer/pkg/runtime"
 	"istio.io/mixer/pkg/template"
 	"istio.io/mixer/pkg/tracing/zipkin"
 	"istio.io/mixer/pkg/version"
@@ -75,6 +76,8 @@ type serverArgs struct {
 	serverKeyFile                 string
 	clientCertFiles               string
 	configStoreURL                string
+	configStore2URL               string
+	configDefaultNamespace        string
 	configFetchIntervalSec        uint
 	configIdentityAttribute       string
 	configIdentityAttributeDomain string
@@ -86,7 +89,7 @@ type serverArgs struct {
 	globalConfigFile string
 }
 
-func serverCmd(repo template.Repository, adapters []pkgAdapter.InfoFn, printf, fatalf shared.FormatFn) *cobra.Command {
+func serverCmd(info map[string]template.Info, adapters []pkgAdapter.InfoFn, printf, fatalf shared.FormatFn) *cobra.Command {
 	sa := &serverArgs{}
 	serverCmd := cobra.Command{
 		Use:   "server",
@@ -103,7 +106,7 @@ func serverCmd(repo template.Repository, adapters []pkgAdapter.InfoFn, printf, f
 			return nil
 		},
 		Run: func(cmd *cobra.Command, args []string) {
-			runServer(sa, repo, adapters, printf, fatalf)
+			runServer(sa, info, adapters, printf, fatalf)
 		},
 	}
 
@@ -139,6 +142,12 @@ func serverCmd(repo template.Repository, adapters []pkgAdapter.InfoFn, printf, f
 
 	serverCmd.PersistentFlags().StringVarP(&sa.configStoreURL, "configStoreURL", "", "",
 		"URL of the config store. May be fs:// for file system, or redis:// for redis url")
+
+	serverCmd.PersistentFlags().StringVarP(&sa.configStore2URL, "configStore2URL", "", "",
+		"URL of the config store. May be fs:// for file system, or redis:// for redis url")
+
+	serverCmd.PersistentFlags().StringVarP(&sa.configDefaultNamespace, "configDefaultNamespace", "", "default",
+		"Namespace used to store mesh wide configuration.")
 
 	// Hide configIdentityAttribute and configIdentityAttributeDomain until we have a need to expose it.
 	// These parameters ensure that rest of Mixer makes no assumptions about specific identity attribute.
@@ -187,7 +196,7 @@ func configStore(url, serviceConfigFile, globalConfigFile string, printf, fatalf
 	return s
 }
 
-func runServer(sa *serverArgs, repo template.Repository, adapters []pkgAdapter.InfoFn, printf, fatalf shared.FormatFn) {
+func runServer(sa *serverArgs, info map[string]template.Info, adapters []pkgAdapter.InfoFn, printf, fatalf shared.FormatFn) {
 	printf("Mixer started with args: %#v", sa)
 
 	var err error
@@ -218,6 +227,23 @@ func runServer(sa *serverArgs, repo template.Repository, adapters []pkgAdapter.I
 		}
 		eval = ilEval
 	}
+
+	repo := template.NewRepository(info)
+	adapter.Inventory2()
+	adapterMap := adapter.InventoryMap(adapters)
+
+	store2, err := store.NewRegistry2(config.Store2Inventory()...).NewStore2(sa.configStore2URL)
+	if err != nil {
+		fatalf("Failed to connect to the configuration2 server. %v", err)
+	}
+	dispatcher, err := mixerRuntime.New(eval, gp, adapterGP,
+		sa.configIdentityAttribute, sa.configDefaultNamespace,
+		store2, adapterMap, info,
+	)
+	if err != nil {
+		fatalf("Failed to create runtime dispatcher. %v", err)
+	}
+
 	store := configStore(sa.configStoreURL, sa.serviceConfigFile, sa.globalConfigFile, printf, fatalf)
 	adapterMgr := adapterManager.NewManager(adapter.Inventory(), aspect.Inventory(), eval, gp, adapterGP)
 	configManager := config.NewManager(eval, adapterMgr.AspectValidatorFinder, adapterMgr.BuilderValidatorFinder, adapters,
@@ -358,8 +384,9 @@ func runServer(sa *serverArgs, repo template.Repository, adapters []pkgAdapter.I
 
 	// get everything wired up
 	gs := grpc.NewServer(grpcOptions...)
+
 	// FIXME construct a runtime.New as dispatcher param
-	s := api.NewGRPCServer(adapterMgr, nil, gp)
+	s := api.NewGRPCServer(adapterMgr, dispatcher, gp)
 	mixerpb.RegisterMixerServer(gs, s)
 
 	printf("Istio Mixer: %s", version.Info)
--- a/cmd/server/cmd/server.go
+++ b/cmd/server/cmd/server.go
@@ -228,22 +228,26 @@ func runServer(sa *serverArgs, info map[string]template.Info, adapters []pkgAdap
 		eval = ilEval
 	}
 
+	var dispatcher mixerRuntime.Dispatcher
+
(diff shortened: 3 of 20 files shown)


cmd/server/cmd/server.go
```diff
@@ -49,6 +49,7 @@ import (
 	"istio.io/mixer/pkg/expr"
 	"istio.io/mixer/pkg/il/evaluator"
 	"istio.io/mixer/pkg/pool"
+	mixerRuntime "istio.io/mixer/pkg/runtime"
```
ZackButcher: Is this needed? It doesn't appear that any other package is called `runtime` in this file.

pkg/runtime/controller.go
```diff
@@ -75,13 +74,21 @@ type Controller struct {
 }
 
 // RulesKind defines the config kind name of mixer rules.
-const RulesKind = "mixer-rules"
+const RulesKind = "mixer-rule"
```
ZackButcher: Nit: there's quite a few type and const declarations in this file, we should consider consolidating into `type` and `const` blocks.

pkg/runtime/controller.go
```diff
@@ -233,12 +248,58 @@ func (c *Controller) validHandlerConfigs() map[string]*cpb.Handler {
...
+const maxAttributeManifests = 5
+
+// processAttributeManifests loads attribute manifests to produce an AttributeDescriptorFinder.
+// TODO in apply events mark if AttributeManifestKind changes, so this step is skipped.
+// attribute manifests are not expected to change often.
+func (c *Controller) processAttributeManifests() expr.AttributeDescriptorFinder {
+	attrs := make([]*cpb.AttributeManifest, 0, maxAttributeManifests)
+	// check rules and ensure only good handlers and instances are used.
+	// record handler - instance associations
+	nattributes := 0
```
ZackButcher: nit: s/nattributes/n. The variable isn't important enough to have a big name.

pkg/runtime/dispatcher.go
```diff
@@ -219,6 +219,13 @@ func (m *dispatcher) Check(ctx context.Context, requestBag attribute.Bag) (*adap
 		},
 	)
 	res, _ := cres.(*adapter.CheckResult)
+	if res == nil { // There was nothing to do.
+		res = &adapter.CheckResult{}
+	}
+
+	if glog.V(3) {
```
ZackButcher: `V(4)` here and elsewhere. I'm planning to go through and consolidate our logging levels, but for now lets keep things `V(2)` for normal and `V(4)` for verbose. I don't see many different levels providing us value.

pkg/runtime/init.go
```diff
@@ -69,23 +69,25 @@ func kindMap(adapterInfo map[string]*adapter.BuilderInfo,
...
 	}
 	// typed handlers
 	for kind, info := range adapterInfo {
 		kindMap[kind] = info.DefaultConfig
+		glog.Infof("adapter Kind: %s, %v", kind, info.DefaultConfig)
 	}
 	kindMap[RulesKind] = &cpb.Rule{}
+	glog.Infof("template Kind: %s, %v", RulesKind, &cpb.Rule{})
+	kindMap[AttributeManifestKind] = &cpb.AttributeManifest{}
+	glog.Infof("template Kind: %s, %v", AttributeManifestKind, &cpb.AttributeManifest{})
```
ZackButcher: Why are we logging empty structs here?

adapter/inventory.go
```diff
@@ -47,3 +47,13 @@ func Inventory() []adapter.RegisterFn {
 		noop.Register,
 	}
 }
+
+// InventoryMap converts adapter inventory to a builder map.
+func InventoryMap(inv []adapter.InfoFn) map[string]*adapter.BuilderInfo {
```
guptasu: Should this function be on BuilderInfoRegistry.go ?

cmd/server/cmd/server.go
```diff
@@ -218,6 +227,27 @@ func runServer(sa *serverArgs, repo template.Repository, adapters []pkgAdapter.I
...
+
+	var dispatcher mixerRuntime.Dispatcher
+
+	// TODO until the dispatcher 2 switch is complete,
+	// dispatcher 2 is only enabled when configStore2URL is specified.
+	if sa.configStore2URL != "" {
+		adapterMap := adapter.InventoryMap(adapters)
+		store2, err := store.NewRegistry2(config.Store2Inventory()...).NewStore2(sa.configStore2URL)
+		if err != nil {
+			fatalf("Failed to connect to the configuration2 server. %v", err)
```
guptasu: to the configuration server

cmd/server/cmd/server.go
```diff
@@ -218,6 +227,27 @@ func runServer(sa *serverArgs, repo template.Repository, adapters []pkgAdapter.I
...
+		}
+		dispatcher, err = mixerRuntime.New(eval, gp, adapterGP,
```
guptasu: should we just call this runtime ?

pkg/adapter/check.go
```diff
@@ -53,3 +54,7 @@ func (r *CheckResult) Combine(otherPtr interface{}) interface{} {
 	}
 	return r
 }
+
+func (r *CheckResult) String() string {
+	return fmt.Sprintf("CheckResult: status:%d, duration:%d, usecount:%d", r.Status, r.ValidDuration, r.ValidDuration)
```
guptasu: r.Status should have a non number version too. S_OK etc.

guptasu: Shall we also print the message inside r ?

pkg/api/grpcServer.go
```diff
@@ -119,10 +119,16 @@ func (s *grpcServer) Check(legacyCtx legacyContext.Context, req *mixerpb.CheckRe
 		if err != nil {
 			out2 = status.WithError(err)
 		} else {
+			out2 = cr.Status
 			resp.Precondition.ValidDuration = cr.ValidDuration
 			// FIXME cr.ValidUseCount should be int32
 			resp.Precondition.ValidUseCount = int32(cr.ValidUseCount)
 		}
+		if status.IsOK(out2) {
+			glog.V(1).Info("Check2 returned with ok : ", statusString(out2))
```
guptasu: Check returned with ok

pkg/il/evaluator/evaluator.go
```diff
@@ -101,6 +101,12 @@ func (e *IL) AssertType(expr string, finder expr.AttributeDescriptorFinder, expe
 	return nil
 }
 
+// ChangeVocabulary handles changing of the attribute vocabulary.
+func (e *IL) ChangeVocabulary(finder expr.AttributeDescriptorFinder) {
+	e.finder = finder
```
guptasu: This seems non thread safe, in case the same IL is used across multiple request ? Or that is not the case ?

pkg/runtime/controller.go
```diff
@@ -233,12 +248,58 @@ func (c *Controller) validHandlerConfigs() map[string]*cpb.Handler {
...
+// maxAttributeManifests is the likely maximum number of manifests in the config.
+// It is used to avoid reallocation on config change.
+const maxAttributeManifests = 5
+
+// processAttributeManifests loads attribute manifests to produce an AttributeDescriptorFinder.
+// TODO in apply events mark if AttributeManifestKind changes, so this step is skipped.
+// attribute manifests are not expected to change often.
+func (c *Controller) processAttributeManifests() expr.AttributeDescriptorFinder {
+	attrs := make([]*cpb.AttributeManifest, 0, maxAttributeManifests)
+	// check rules and ensure only good handlers and instances are used.
```
guptasu: comment copy-paste error

pkg/runtime/controller.go
```diff
@@ -233,12 +248,58 @@ func (c *Controller) validHandlerConfigs() map[string]*cpb.Handler {
...
+	// record handler - instance associations
+	nattributes := 0
+	for k, cfg := range c.configState {
+		if k.Kind != AttributeManifestKind {
+			continue
+		}
+		attrs = append(attrs, cfg.(*cpb.AttributeManifest))
```
guptasu: shall we just flatten all attributes in the AttributeManifest into a map ? GetAttribute might get faster

cmd/server/cmd/server.go
mandarjog: When you are reading code, `runtime` can be confusing, it feel like the builtin runtime.

cmd/server/cmd/server.go
```diff
@@ -140,6 +143,12 @@ func serverCmd(repo template.Repository, adapters []pkgAdapter.InfoFn, printf, f
 	serverCmd.PersistentFlags().StringVarP(&sa.configStoreURL, "configStoreURL", "", "",
 		"URL of the config store. May be fs:// for file system, or redis:// for redis url")
 
+	serverCmd.PersistentFlags().StringVarP(&sa.configStore2URL, "configStore2URL", "", "",
+		"URL of the config store. May be fs:// for file system, or redis:// for redis url")
+
+	serverCmd.PersistentFlags().StringVarP(&sa.configDefaultNamespace, "configDefaultNamespace", "", "default",
```
jmuk: I believe the default value shouldn't be `"default"` -- this is the namespace where many of k8s users are using by default.
Instead, use `"istio-config-default"` which is defined in https://github.com/istio/mixer/blob/master/pkg/runtime/resolver.go#L83.

adapter/inventory.go
mandarjog: I can't find this file. Are you saying we should create a new one ?

cmd/server/cmd/server.go
mandarjog: we will replace all "*2". Otherwise messages for 1 and 2 both look the same.

cmd/server/cmd/server.go
mandarjog: When you are reading code, `runtime` can be confusing, it feel like the builtin runtime.

pkg/adapter/check.go
mandarjog: D.

pkg/api/grpcServer.go
mandarjog: D

pkg/il/evaluator/evaluator.go
mandarjog: That is correct, my email earlier in the day was about this issue.'

The function just below this ( `ConfigChange`),  works in the exact same way, and it is in use today. I think it works because  `df` the attribute finders do not change often (or at all).

I will fix it in a another PR.

pkg/runtime/controller.go
mandarjog: I define at the top if they are generally useful.
If they are not generally useful, I define them close to the point of use.

I have moved some consts around to that end now.

pkg/runtime/controller.go
mandarjog: good catch. D.

pkg/runtime/controller.go
mandarjog: yes, was it was waiting on smarter  config change for attributes, did both.

pkg/runtime/dispatcher.go
mandarjog: I think there is value in more levels. 
Sometime you want super verbose logs, they could be v(6) even...
I changed these to v4.

pkg/il/evaluator/evaluator.go
guptasu: Yes..  let's not do in this PR.

adapter/inventory.go
mandarjog: we can keep that on the validation path.

cmd/server/cmd/server.go
mandarjog: aah yes, good catch. changed it.

cmd/server/cmd/root.go
```diff
@@ -50,8 +50,9 @@ func GetRootCmd(args []string, info map[string]template.Info, adapters []adapter
 	_ = fs.Parse([]string{})
 	flag.CommandLine = fs
 
+	// template.NewRepository(info)
```
jmuk: What is this line?  Remove this if this is just a comment-out.

pkg/config/store/convert.go
```diff
@@ -38,5 +39,9 @@ func convert(spec map[string]interface{}, target proto.Message) error {
 	if err != nil {
 		return err
 	}
-	return jsonpb.Unmarshal(bytes.NewReader(jsonData), target)
+	if err = jsonpb.Unmarshal(bytes.NewReader(jsonData), target); err != nil {
+		glog.Warningf("unable to unmarshal: %s, %s", err.Error(), string(jsonData))
```
jmuk: I'm not sure if this was added for your debugging, or this should be in production. Normally, simply returning an error would be sufficient (assuming that `err` contains information about name of property or types), unnecessary to dump the `jsonData`.

If you want to ensure `jsonData` in the error, maybe this line should return a new error wrapping `err`, as `fmt.Errorf("unable to unmarshal: %s, %s", err.Error(), string(jsonData))` rather than logging directly.

pkg/runtime/dispatcher.go
```diff
@@ -219,6 +219,13 @@ func (m *dispatcher) Check(ctx context.Context, requestBag attribute.Bag) (*adap
...
+		glog.Infof("Check %s", res)
```
jmuk: If using `Infof`, `glog.V(3).Infof(...)` should be better (shorter and clearer, I think)

pkg/runtime/dispatcher.go
```diff
@@ -346,18 +353,30 @@ func (m *dispatcher) run(ctx context.Context, runArgs []*runArg) (adapter.Result
 
 // runAsync runs the dispatchFn using a scheduler. It also adds a new span and records prometheus metrics.
 func (m *dispatcher) runAsync(ctx context.Context, callinfo *Action, results chan *result, do dispatchFn) {
+	if glog.V(4) {
+		glog.Infof("runAsync %v", *callinfo)
```
jmuk: ditto

ZackButcher:
/lgtm 

Bot won't recognize both in the same comment apparently.

PR #1153, opened 2017-08-28
wait asynchronously for CRDs which are not yet registered

### LINKED ISSUE #1117: CRD Store.Init() should not return error for retryable conditions

jmuk:
Fixes #1117.

Previous PR #1137 retries to check CRDs and creates the cache
stores, but the waiting has timeout. If such CRDs are registered
later through manual operations or additional scripts, mixer
fails to set up.

This PR allows to wait for such CRDs asynchronously.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1153)


pkg/config/crd/store.go
```diff
@@ -198,13 +237,16 @@ func toEvent(t store.ChangeType, obj interface{}) store.BackendEvent {
 }
 
 func (s *Store) dispatch(ev store.BackendEvent) {
+	s.watchMutex.RLock()
```
mandarjog: just `defer s.WatchMutext.RUnlock()` and remove the other 2

pkg/config/crd/store.go
```diff
@@ -84,40 +85,46 @@ func waitForSynced(ctx context.Context, informers map[string]cache.SharedInforme
...
-		if crdCtx.Err() != nil {
-			// TODO: runs goroutines for remaining kinds.
+	retryCtx := ctx
+	if timeout > 0 {
+		var cancel context.CancelFunc
+		retryCtx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+	for added := 0; added < len(kinds); {
+		if retryCtx.Err() != nil {
```
mandarjog: can you increase to .V(4) otherwise it is printing this message in a tight loop.

`Retrying ...`

pkg/config/crd/store.go
```diff
@@ -128,6 +135,7 @@ func (s *Store) Init(ctx context.Context, kinds []string) error {
 			glog.V(3).Infof("Failed to obtain resources for CRD: %v", err)
 			continue
 		}
+		s.cacheMutex.Lock()
```
mandarjog: use defer. It protects from panics.

pkg/config/crd/store.go
jmuk: done

pkg/config/crd/store.go
jmuk: I'm hesitant to use Lock here. This is in a loop, the use of defer here requires an extra func. Extra unnamed func just for defer doesn't look great to me.

pkg/config/crd/store.go
jmuk: done

pkg/config/store/fsstore2.go
```diff
@@ -157,7 +159,9 @@ func (s *fsStore2) checkAndUpdate() {
 		return
 	}
 	s.data = newData
+	s.watchMutex.RLock()
 	if s.watchCtx == nil || s.watchCtx.Err() != nil {
+		s.watchMutex.RUnlock()
```
mandarjog: defer unlock ?

pkg/config/crd/store.go
mandarjog: Do we need to re-check in a tight loop here? agreed that locking repeatedly in a tight loop is not very good.
Can we wait for 500ms or so before checking again ?

mandarjog:
@jmuk can you add a periodic sleep/poll instead of a tight loop checking for cache updates?
Otherwise this looks good.

pkg/config/crd/store.go
jmuk: added the delay.

PR #1156, opened 2017-08-29
Introduce the memquota2 adapter.

geeknoid:
This is a cut-and-paste of the original memory-based quota adapter. The core quota logic is unchanged, it's just the glue with the rest of the system that is new.

100% code coverage.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1156)

### DIFF
--- a/adapter/memquota2/config/config.proto
+++ b/adapter/memquota2/config/config.proto
@@ -1,0 +1,46 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package adapter.memquota.config;
+
+import "google/protobuf/duration.proto";
+import "gogoproto/gogo.proto";
+
+option go_package="config";
+option (gogoproto.goproto_getters_all) = false;
+option (gogoproto.equal_all) = false;
+option (gogoproto.gostring_all) = false;
+
+message Params {
+	message Quota {
+		// The name of the quota
+		string name = 1;
+
+		// The upper limit for this quota.
+		int64 max_amount = 2;
+
+		// The amount of time allocated quota remains valid before it is
+		// automatically released. This is only meaningful for rate limit
+		// quotas, otherwise the value must be zero.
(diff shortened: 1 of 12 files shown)


adapter/memquota2/config/config.proto
```diff
@@ -0,0 +1,43 @@
...
+import "google/protobuf/duration.proto";
+import "gogoproto/gogo.proto";
+
+option go_package="config";
+option (gogoproto.goproto_getters_all) = false;
+option (gogoproto.equal_all) = false;
+option (gogoproto.gostring_all) = false;
+
+message Params {
+	message Quota {
```
ZackButcher: I've been calling these sub-messages "FooInfo". I think it'd be better if these were named consistently across all of our adapters for this naming. Do we want to call these "Foo" or "FooInfo"? My concern is confusion between instances and these objects if we call them "Foo" (e.g. "Quota" is the name we use for the Istio-thing, and also the name you use in this adapter, but they're not the same thing).

adapter/memquota2/memquota.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (h *handler) Close() error {
+	h.common.ticker.Stop()
+	return nil
+}
+
+////////////////// Bootstrap //////////////////////////
+
+// GetBuilderInfo returns the BuilderInfo associated with this adapter implementation.
+func GetBuilderInfo() adapter.BuilderInfo {
```
ZackButcher: Nit: move this to the top of the file. It's effectively the summary of what's going on in this adapter, I think the whole file reads better with this up top.

adapter/memquota2/memquota.go
```diff
@@ -0,0 +1,246 @@
...
+
+func (b *builder) ConfigureQuotaHandler(types map[string]*quota.Type) error {
+	b.types = types
+	return nil
+}
+
+////////////////// Runtime Methods //////////////////////////
+
+func (h *handler) HandleQuota(context context.Context, instance *quota.Instance, args adapter.QuotaRequestArgs) (adapter.QuotaResult2, error) {
+	q := h.limits[instance.Name]
```
ZackButcher: What if the instance name isn't found at runtime?

adapter/memquota2/config/config.proto
geeknoid: Renamed to QuotaInfo

adapter/memquota2/memquota.go
geeknoid: There's a check in the Build method to ensure we have a limit for every instance we're receiving.

In general, I don't think we need adapters to double-check that Mixer isn't doing something invalid such as sending instances the adapters aren't prepared to handle.

adapter/memquota2/memquota.go
geeknoid: This follows the layout of all the other adapters I created. I was going to suggest changing your adapters to move GetBuilderInfo to the end of the file :-)

I find this bit of boilerplate not that interesting which is why I put it at the end, out of the way. I was actually thinking about moving this out into a standard info.go file to further isolate it...

ZackButcher: I'll move my stuff down to the bottom I guess :)

PR #1158, opened 2017-08-29
simplify crd gen and add route-rules and attribute-manifests

mandarjog:
1. Adds route-rule and attribute-manifests
2. Removes dep on k8s, (gogland is happy) 

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1158)

### DIFF
--- a/cmd/server/cmd/crd.go
+++ b/cmd/server/cmd/crd.go
@@ -104,7 +104,6 @@ type crdVar struct {
 	Version    string
 }
 
-// newCrdVar get a new crdVar
 func newCrdVar(shrtName, implName, pluralName, label string) *crdVar {
 	return &crdVar{
 		ShrtName:   shrtName,
--- a/cmd/server/cmd/crd.go
+++ b/cmd/server/cmd/crd.go
@@ -15,30 +15,46 @@
 package cmd
 
 import (
+	"bytes"
 	"sort"
+	gotemplate "text/template"
 
-	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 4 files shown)


cmd/server/cmd/crd.go
```diff
@@ -74,40 +90,56 @@ func listCrdsInstances(printf shared.FormatFn, infos map[string]template.Info) {
...
+	ShrtName   string
+	ImplName   string
+	PluralName string
+	Label      string
+	Name       string
+	Group      string
+	Version    string
+}
+
+// newCrdVar get a new crdVar
```
guptasu: consider removing this comment

cmd/server/cmd/crd.go
```diff
@@ -74,40 +90,56 @@ func listCrdsInstances(printf shared.FormatFn, infos map[string]template.Info) {
...
+		Group:      Group,
+		Version:    Version,
 	}
-	out, err := yaml.Marshal(crd)
-	if err != nil {
-		printf("%s", err)
-		return
+}
+
+func printCrd(printf, fatalf shared.FormatFn, shrtName, implName, pluralName, label string) {
```
guptasu: label is unclear. Any better name ? kind ?

guptasu: Curious why change from object model to text/template ? The original motivation such that in the future if we want to use the client library to publish CRD directly along with printing, the same code can be used.

cmd/server/cmd/crd.go
mandarjog: D

cmd/server/cmd/crd.go
mandarjog: It is the istio label, I think it is ok.

mandarjog: I think this is simpler code. If we decide to push it in, we can move back to using k8s.

PR #1161, opened 2017-08-29
Stackdriver Logging

ZackButcher:
Note the diff is larger than the actual change - that's because I moved the existing stackdriver.go from `//adapter/stackdriver/stackdriver.go` to `//adapter/stackdriver/metric/metric.go`, and repurposed `//adapter/stackdriver/stackdriver.go` into a facade over the various mesh function implementations.

A bit more testing needs to happen in this PR, but the main logic is fine to start reviewing. Note that part of the changes here are actually due to https://github.com/istio/mixer/pull/1146 which this PR is branched off of.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1161)


adapter/stackdriver/BUILD
```diff
@@ -1,52 +1,30 @@
...
-        "@com_google_cloud_go//monitoring/apiv3:go_default_library",
-        "@org_golang_google_api//option:go_default_library",
-        "@org_golang_google_genproto//googleapis/api/metric:go_default_library",
-        "@org_golang_google_genproto//googleapis/api/monitoredres:go_default_library",
-        "@org_golang_google_genproto//googleapis/monitoring/v3:go_default_library",
-        "@org_golang_x_net//context:go_default_library",
     ],
 )
+
+go_prefix("istio.io/mixer")
```
douglas-reid: curious: why add `go_prefix` here?

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+
+		logger adapter.Logger
+		client io.Closer
+		log    logFn
+		info   map[string]info
+	}
+)
+
+var (
+	_ logentry.HandlerBuilder = &builder{}
```
douglas-reid: is this strictly necessary?  won't the signature of `NewBuilder()` enforce what we are trying to enforce here? It seems like the `handler` related var is the only additive check.

also, I feel like this stuff could use an explanatory comment.

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (b *builder) Build(c adapter.Config, env adapter.Env) (adapter.Handler, error) {
+	logger := env.Logger()
+	cfg := c.(*config.Params)
+	infos := make(map[string]info)
+	for name, log := range cfg.LogInfo {
+		_, found := b.types[name]
+		if !found {
+			logger.Infof("configured with log info about %s which is not an Istio metric", name)
```
douglas-reid: is this message right? do we mean "which is not an Istio *log*" ?

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+			continue
+		}
+		infos[name] = info{
+			labels: log.LabelNames,
+			tmpl:   tmpl,
+			req:    log.HttpMapping,
+		}
+	}
+
+	client, err := b.makeClient(context.Background(), cfg.ProjectId, helper.ToOpts(cfg)...)
```
douglas-reid: does this mean that the client will always use the background context for sending updates to SD ?  It looks like this is a SD golang API convention.  Is their recommendation to use a single long-lived client (and forget about cancelation, deadlines, etc.)?

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+	info struct {
+		labels []string
+		tmpl   *template.Template
+		req    *config.Params_LogInfo_HttpRequestMapping
+	}
+
+	handler struct {
+		now func() time.Time // used for testing
+
+		logger adapter.Logger
```
douglas-reid: nit: having both `logger` and `log` is a bit confusing. I don't have any great suggestions here, unfortunately.

What about: `mixerLog` and `logEntry` (or similar)?

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		lip := variables[mapping.LocalIp]
+		localip, ok := lip.(string)
+		if ok {
+			req.LocalIP = localip
+		}
+
+		rip := variables[mapping.RemoteIp]
+		remoteip, ok := rip.(string)
```
douglas-reid: all of these lines could be collapsed:

```
if remoteip, ok := rip.(string); ok {
   req.RemoteIP = remoteip
}
```

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+		if ok {
+			req.RemoteIP = remoteip
+		}
+	}
+	return req
+}
+
+func toInt(v interface{}) (int64, bool) {
```
douglas-reid: isn't this really `toInt64` ?

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,204 @@
...
+	// case int, int8, int16, ...: return int64(i), true
```
douglas-reid: do we care about adding `int8`, `int16`, ... (and maybe `uint*`) ?

adapter/stackdriver/BUILD
ZackButcher: gazelle adds it

adapter/stackdriver/log/log.go
ZackButcher: Why? It's a pretty standard practice in go and I don't really think it's worth commenting (we'd say something like "ensure builder implements interface foo" but that's exactly what the code says too). Personally, I write it when I create the struct to ensure I have all the interfaces I want (before I write the actual funcs themselves). Leaving them around shows the intent. Plus, only logentry.HandlerBuilder is enforced, logentry.Handler is not enforced by return values.

adapter/stackdriver/log/log.go
ZackButcher: Good catch, thanks.

adapter/stackdriver/log/log.go
ZackButcher: This is the context used to create the client (it controls the lifetime of the transport), it's not used to send individual messages to the API. [Today you cannot control the context used to send bundles of `logging.Entry`s](https://github.com/GoogleCloudPlatform/google-cloud-go/blob/f0500eb3fbc9c327a4d4a05b40afbd77d0aee85a/logging/logging.go#L378-L382).

adapter/stackdriver/log/log.go
ZackButcher: Much better, thanks.

adapter/stackdriver/log/log.go
ZackButcher: Renamed

adapter/stackdriver/log/log.go
ZackButcher: I had them originally, but in practice these ints come from expression evaluation and are always int64 as a result.

adapter/stackdriver/log/log.go
ZackButcher: Changed `logger` to `l`

adapter/stackdriver/BUILD
jmuk: It's weird gazelle adds go_prefix here. Did you run `./bin/gazelle` script or run the gazelle command by yourself?

ZackButcher: Ran gazelle directly myself, though IIRC the script adds it too. Removed it either way.

jmuk: thanks. I will look into why gazelle runs weirdly after this PR is merged.

adapter/stackdriver/log/log.go
douglas-reid: That's what I thought from the signature. That's a bit unfortunate, but I guess it is something beyond our control.

adapter/stackdriver/log/log.go
douglas-reid: So, it seems like we agree that `logentry.HandlerBuilder` is enforced. If you want to leave this line for redundancy and consistency, I guess that's OK.  My original point, however, was that this line in particular is not necessary.

IMHO, anything that uses `_` for magic purposes should have a comment (here maybe: `ensure type safety`). This is similar to how I feel about imports with the blank identifier.

adapter/stackdriver/BUILD
```diff
@@ -1,52 +1,28 @@
-package(default_visibility = ["//visibility:public"])
-
-load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_prefix", "go_test")
```
douglas-reid: i think you can remove `go_prefix` from this load too.

adapter/stackdriver/log/log.go
```diff
@@ -0,0 +1,198 @@
...
+			tmpl:   tmpl,
+			req:    log.HttpMapping,
+		}
+	}
+
+	client, err := b.makeClient(context.Background(), cfg.ProjectId, helper.ToOpts(cfg)...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create stackdriver logging client: %v", err)
+	}
+	return &handler{log: client.Logger("istio-mixer").Log, client: client, now: time.Now, l: logger}, nil
```
douglas-reid: should the `logName` be configurable?  Or are we going to enforce always logging to `istio-mixer` ?

ZackButcher: My understanding is that this winds up being the log ID, which makes up the resource name. It identifies all the entries, so with this name all Stackdriver logs produced by Mixer belong to the same log collection. We should probably restructure this to create and hold on to a logger per Istio log name. We might even want to prefix the log name with "istio-" or similar.

testdata/config/stackdriver.yaml
```diff
@@ -0,0 +1,83 @@
...
+spec:
+  selector: "true"
+  actions:
+  - handler: handler.stackdriver.istio-config-default
+    instances:
+    - requestcount.metric.istio-config-default
+    - requestduration.metric.istio-config-default
+    - requestsize.metric.istio-config-default
+    - responsesize.metric.istio-config-default
+    - global.logentry.stackdriver.istio-config-default
```
douglas-reid: nit: add a newline here.

testdata/config/stackdriver.yaml
```diff
@@ -0,0 +1,83 @@
...
+    sourceUser: origin.user | ""
+    method: request.method | ""
+    url: request.path | ""
+    protocol: request.scheme | ""
+    responseCode: response.code | 0
+    responseSize: response.size | 0
+    latency: response.duration | "0s"
+  monitoredResourceType: global
+  monitoredResourceDimensions:
+    project_id:
```
douglas-reid: what happens if this is not set?  should we add a comment to this file here?

ZackButcher:
Tests pass on Doug's cluster and mine:
```shell
$ ./tests/e2e.sh --mixer_hub=gcr.io/google.com/zbutcher-test --mixer_tag=pr-1161
<snip>
Tests Summary
```

PR #1163, opened 2017-08-29
Update the logentry and metric templates with fields for monitored resources

ZackButcher:
Should monitored resource type be a string, or a `ValueType`? In Stackdriver it is always a string (as are all the dimensions).

**Release note**:

```release-note
Add fields for monitored resources to the metric and log entry templates.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1163)

### DIFF
--- a/template/logentry/template.proto
+++ b/template/logentry/template.proto
@@ -34,9 +34,13 @@ message Template {
     // Severity indicates the importance of the log entry.
     string severity = 3;
 
-    // An expression to compute the type of the monitored resource this log entry is being recorded on.
+    // Optional. An expression to compute the type of the monitored resource this log entry is being recorded on.
+    // If the logging backend supports monitored resources, these fields are used to populate that resource.
+    // Otherwise these fields will be ignored by the adapter.
     string monitored_resource_type = 4;
 
-    // A set of expressions that will form the dimensions of the monitored resource this log entry is being recorded on.
+    // Optional. A set of expressions that will form the dimensions of the monitored resource this log entry is being
+    // recorded on. If the logging backend supports monitored resources, these fields are used to populate that resource.
+    // Otherwise these fields will be ignored by the adapter.
     map<string, istio.mixer.v1.config.descriptor.ValueType> monitored_resource_dimensions = 5;
 }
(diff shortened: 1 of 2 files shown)


template/logentry/template.proto
```diff
@@ -33,4 +33,10 @@ message Template {
 
     // Severity indicates the importance of the log entry.
     string severity = 3;
+
+    // An expression to compute the type of the monitored resource this log entry is being recorded on.
+    string monitored_resource_type = 4;
```
manlinl: string is too limited, e.g. Google service control uses labels, map[string]string to identify resource. 
istio.mixer.v1.config.descriptor.ValueType is more flexible. How about calling the field monitored_resource, so it doesn't have to be a resource type?

manlinl: call the field monitored_resource to be generic?

template/metric/template.proto
```diff
@@ -29,4 +29,10 @@ message Template {
 
     // The unique identity of the particular metric to report.
     map<string, istio.mixer.v1.config.descriptor.ValueType> dimensions = 2;
+
+    // An expression to compute the type of the monitored resource this metric is being reported on.
+    string monitored_resource_type = 3;
```
manlinl: call the field monitored_resource to be generic?

template/logentry/template.proto
ZackButcher: I'm not sure what you mean by generic. Today we only have a single example of monitored resources, which are in Stackdriver. Stackdriver represents a MR as a type + labels. If we call the field only "monitored_resource" then I don't think it's clear what I should populate the field with.

ZackButcher:
PTAL

template/logentry/template.proto
manlinl: My understanding is the template is meant to be general purpose. So adapter can share with it. In stackdriver, it is a type + labels, in other monitor BE, could be just a url + key value pair attribute. My thought is name it resouce + dimension and let adapter developer to define its exact meaning. Not a big thing, so it is up to you.

ZackButcher:
(Trying to poke the bot to merge things)

PR #1167, opened 2017-08-30
pluralize kind names per k8s spec

mandarjog:
K8s expects kind names to be pluralized per k8s spec (i.e. correctly :-) )
This reuses their pluralize code.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1167)

### DIFF
--- a/cmd/server/cmd/crd.go
+++ b/cmd/server/cmd/crd.go
@@ -43,8 +43,8 @@ func crdCmd(tmplInfos map[string]template.Info, adapters []pkgAdapter.InfoFn, pr
 		Use:   "all",
 		Short: "List all CRDs",
 		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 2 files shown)


cmd/server/cmd/crd.go
```diff
@@ -104,19 +104,33 @@ type crdVar struct {
 	Version    string
 }
 
-func newCrdVar(shrtName, implName, pluralName, label string) *crdVar {
+// pluralize gives a plural the way k8s like it.
+// https://github.com/kubernetes/gengo/blob/master/namer/plural_namer.go
+func pluralize(singular string) string {
+	switch string(singular[len(singular)-1]) {
```
guptasu: Consider adding a test for these cases in crd_test.

mandarjog: Added test data for pluralization.

PR #1169, opened 2017-08-30
Fix the concurrency issue in evaluator.IL.

ozevren:
Mainly fixing two issues:
- The write to the finder pointer is not guaranteed to be atomic.
- Due to pointer capture in a local during the compilation, it was possible
  to add an interpreter to the cache that is based on the old version
  of the metadata.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1169)

### DIFF
--- a/pkg/il/evaluator/evaluator.go
+++ b/pkg/il/evaluator/evaluator.go
@@ -15,8 +15,10 @@
 package evaluator
 
 import (
+	"errors"
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/golang/glog"
 	lru "github.com/hashicorp/golang-lru"
@@ -33,9 +35,17 @@ import (
 // IL is an implementation of expr.Evaluator that also exposes specific methods.
 // Specifically, it can listen to config change events.
 type IL struct {
+	cacheSize   int
+	context     *attrContext
+	contextLock sync.RWMutex
+	fMap        map[string]expr.FuncBase
+}
+
+// attrContext captures the set of fields that needs to be kept & evicted together based on
+// particular attribute metadata that was supplied as part of a ChangeListener call.
+type attrContext struct {
 	cache  *lru.Cache
 	finder expr.AttributeDescriptorFinder
-	fMap   map[string]expr.FuncBase
 }
 
 var _ expr.Evaluator = &IL{}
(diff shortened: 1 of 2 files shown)


mandarjog:
Attribute manifest changes are exceedingly rare events, which why the current code has not been a problem. 
I would like to fix this by treating `df` as immutable and the config controller managing the lifecycle of expr, just like it manages lifecycle of handlers.  If attribute vocabulary changes, config controller will create a new expr instance. expr does not have to do any locking or even worry about any of this.
Given how unlikely this event is, we can trash the expression cache. 
If necessary, we can compile all expressions and fill in a new cache **before** making the expr active.

ozevren:
Hi Mandar,

I think we agree on the config/metadata lifecycle model that we should have eventually. We can add a pre-compilation step, triggered by the config controller. (The evaluator needs to have access to the full set of expressions though).

I assume the PR, as it currently stands, is acceptable to you? If so, can I get an LGTM?

ozevren:
Not sure how this PR affects codecov of adapter/stackdriver/bufferedClient.go

mandarjog:
What changed in the last commit ?

ozevren:
Just rebased to master.

pkg/il/evaluator/evaluator.go
```diff
@@ -32,9 +34,17 @@ import (
 // IL is an implementation of expr.Evaluator that also exposes specific methods.
 // Specifically, it can listen to config change events.
 type IL struct {
+	cacheSize       int
+	attrContext     *attrContext
+	attrContextLock sync.RWMutex
```
douglas-reid: not a blocker, but something to consider is just embedding `sync.RWMutex`  anonymously on `IL`:

```
type IL struct {
        sync.RWMutex
...
```

That would allow you to have:

```
e.RLock()
...
e.RUnlock()
```

instead of: `e.attrContextLock.RUnlock()`, etc.

Of course, that implies a slightly different semantic (is the fMap protected too?)

pkg/il/evaluator/evaluator.go
```diff
@@ -32,9 +34,17 @@ import (
 // IL is an implementation of expr.Evaluator that also exposes specific methods.
 // Specifically, it can listen to config change events.
 type IL struct {
+	cacheSize       int
+	attrContext     *attrContext
```
douglas-reid: up to you, but I might consider just calling the field `context`.  The `attr` is almost implied.

pkg/il/evaluator/evaluator.go
ozevren: The lock specifically protects the attrContext. I'd prefer to keep it that way to reduce to scope of locking (and contention). We can refactor as needed when/if we make a major change to the IL struct.

pkg/il/evaluator/evaluator.go
ozevren: Done

PR #1172, opened 2017-08-30
Move a bunch of adapters to the updated code gen model.

geeknoid:
I highlighted the code that needs to be deleted once the new code gen is done.

This looks like a pretty nice improvement in adapter authoring. WDYT?

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1172)

### DIFF
--- a/adapter/memquota2/memquota.go
+++ b/adapter/memquota2/memquota.go
@@ -24,7 +24,7 @@
 // - Since the data is all memory-resident and there isn't any cross-node
 // synchronization, this adapter can't be used in an Istio mixer where
 // a single service can be handled by different mixer instances.
-package memquota
+package memquota // import "istio.io/mixer/adapter/memquota"
 
 import (
 	"context"
@@ -37,10 +37,6 @@ import (
 	"istio.io/mixer/pkg/status"
 	"istio.io/mixer/template/quota"
 )
-
-type builder struct {
-	types map[string]*quota.Type
-}
 
 type handler struct {
 	// common info among different quota adapters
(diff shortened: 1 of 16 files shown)


mandarjog:
lgtm 

Changes are minimal. 

I wish existing functions were not moved around in the file, the changes appear more than what they should be their

geeknoid:
PTAL.

This now covers most of the adapters, except for prometheus and stackdriver. As Mandar pointed out, I did move some of the code vertically within the files in order to have a clean layout, consistent across all adapters. But the logic is mostly unaffected by these changes, as can be seen by the fact the tests all received only marginal changes.

adapter/memquota2/memquota.go
```diff
@@ -24,7 +24,7 @@
 // - Since the data is all memory-resident and there isn't any cross-node
 // synchronization, this adapter can't be used in an Istio mixer where
 // a single service can be handled by different mixer instances.
-package memquota
+package memquota // import "istio.io/mixer/adapter/memquota"
```
guptasu: is this comment intentional ?

adapter/statsd2/statsd.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package statsd
+package statsd // import "istio.io/mixer/adapter/statsd"
```
guptasu: extra comment ?

geeknoid: These comments are actually not comments, but directives to the compiler. See https://stackoverflow.com/questions/37330303/forcing-a-specific-import-path-in-go. Doug recommended we start using these.

It's ridiculous that this looks like a normal comment, but that's what the language designers chose.

PR #1174, opened 2017-08-31
Add manifest and metrics in config files.

mandarjog:
This PR adds metrics and fixes manifest config files in the new format.

This is just so the we have access to a minimal set of configs.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1174)

### DIFF
--- a/testdata/config/metrics.yaml
+++ b/testdata/config/metrics.yaml
@@ -1,0 +1,56 @@
+apiVersion: "config.istio.io/v1alpha2"
+kind: prometheus
+metadata:
+  name: prometheus
+  namespace: istio-config-default
+spec:
+  metrics:
+  - name: requestcount
+    kind: COUNTER
+    label_names: [ source, target, service, method, version,  response_code ]
+  - name: requestduration
+    kind: COUNTER
+    label_names: [ source, target, service, method, version,  response_code ]
+---
+apiVersion: "config.istio.io/v1alpha2"
+kind: metric
+metadata:
+  name: requestcount
+  namespace: istio-config-default
+spec:
+  value: "1"
+  dimensions:
+    source: source.labels["app"] | "unknown"
+    target: target.service | "unknown"
+    service: target.labels["app"] | "unknown"
+    method: request.path | "unknown"
+    version: target.labels["version"] | "unknown"
+    response_code: response.code | 200
+---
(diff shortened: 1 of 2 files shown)


testdata/config/metrics.yaml
```diff
@@ -0,0 +1,56 @@
...
+    target: target.service | "unknown"
+    service: target.labels["app"] | "unknown"
+    method: request.path | "unknown"
+    version: target.labels["version"] | "unknown"
+    response_code: response.code | 200
+---
+apiVersion: "config.istio.io/v1alpha2"
+kind: mixer-rule
+metadata:
+  name: prommetrics
```
guptasu: Should these names be more verbose...like prommetrics_rule. This name prommetrics by itself is confusing.. other names are fine

testdata/config/metrics.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: "config.istio.io/v1alpha2"
+kind: prometheus
+metadata:
+  name: prometheus
+  namespace: istio-config-default
+spec:
+  metrics:
+  - name: requestcount
```
douglas-reid: pardon my config ignorance: are the names jammed together because certain characters are invalid (or not supported)?

`request_count` is much more readable than `requestcount`. i'd almost prefer `requests` to `requestcount`, but we need a general approach to handling. It also looks a bit... off... to have `requestcount` and `response_code` in the same config stanza.

It seems weird to have the config store dictate the naming used for other backends.

mandarjog: k8s is touchy feely about names. The only names that **everyone** likes are all lowercase letters. `request_count` is not allowed, neither is `requestCount`.
All names have to be valid subdomains.  so we **can** use `request-count`, but i don't know where else it will bite us.

testdata/config/metrics.yaml
mandarjog: its `kind` is already mixer-rule

testdata/config/metrics.yaml
douglas-reid: so, can we add support for aliasing the metrics names (within adapters, I guess) to prometheus and elsewhere?

this is a usability issue, imho.

PR #1180, opened 2017-08-31
Suppress gogo* warnings in bazel

douglas-reid:
This PR adds an output filter to the default bazel build configuration to suppress WARNINGs related to codegen stuff.

This is meant to be a short-term workaround until a patch that addresses the need for the `output_to_workspace` option can be developed and fixed.

This PR also moves from the custom `mixer_gen` rule definition to a `native.genrule` for simplicity.

**Release note**:

```release-note-none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1180)

### DIFF
--- a/bazel.rc
+++ b/bazel.rc
@@ -1,5 +1,2 @@
-test --test_output=errors
-test --test_size_filters=-large,-enormous
-
 # Always include the version information in the build
 build --workspace_status_command=./bin/build_stamp.sh --action_env=ISTIO_VERSION --output_filter=^gogo
--- a/tools/codegen/generate.bzl
+++ b/tools/codegen/generate.bzl
@@ -2,36 +2,6 @@ load("@org_pubref_rules_protobuf//protobuf:rules.bzl", "proto_compile")
 load("@org_pubref_rules_protobuf//gogo:rules.bzl", "gogoslick_proto_library", "gogo_proto_library")
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
-def _impl(ctx):
-  m = []
-  for k, v in ctx.attr.importmap.items():
-    m += ["-m %s:%s" % (k, v)]
-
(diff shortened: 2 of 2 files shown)


tools/codegen/generate.bzl
```diff
@@ -107,13 +90,7 @@ def mixer_proto_library(
...
       "importmap": dict(dict(MIXER_IMPORT_MAP, **GOGO_IMPORT_MAP), **importmap),
       "inputs": inputs + MIXER_INPUTS + PROTO_INPUTS,
-      # TODO HELP PLS. Need this to make build work. Is this bad ?
-      # Was getting error: output 'template/ .. .. *_tmpl.pb.go' was not created
-      # But setting this to True, gives a scary warning on the commandline
-      # * - Generating files into the workspace...  This is potentially           *
-      # *   dangerous (may overwrite existing files) and violates bazel's         *
-      # *   sandbox policy.                                                       *
-      "output_to_workspace": True,
+      "output_to_workspace": True, # TODO: Remove when an appropriate fix/workaround exists.
```
guptasu: consider posting the issue id here

bazel.rc
```diff
@@ -0,0 +1,2 @@
+# Always include the version information in the build
+build --workspace_status_command=./bin/build_stamp.sh --action_env=ISTIO_VERSION --output_filter=^gogo
```
guptasu: Curious does this mean all warnings from gogo rules will be suppressed.. Is there any mechanism to suppress warnings just in a particular dir or the specific location where the target is called.

douglas-reid: It does mean that all warnings from gogo* rules will be suppressed. I believe that the filter is just for rule names (and not for directories). Hopefully, this will be just temporary.

I've sent a PR to fix the root cause (pubref/rules_protobuf#118).

PR #1186, opened 2017-09-01
allow CRD store to limit namespaces.

### LINKED ISSUE #1182: K8s store should support a list of namespaces to listen on
In a shared cluster testing, we may be running running multiple mixers. If Mixer always watches for all namespaces in the cluster, this leads to problems.

K8s store must take a list of namespaces 

k8s:// hdjkdkkd/?namespaces=n1,n2...

jmuk:
fixes #1182

**Release note**:

```release-note
query parameter ?namespaces=default,istio-config-default is available for k8s:// config store url.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1186)


pkg/config/crd/store.go
```diff
@@ -138,17 +151,20 @@ func (s *Store) checkAndCreateCaches(
 		}
 		s.cacheMutex.Lock()
 		for _, res := range resources.APIResources {
-			if _, ok := s.caches[res.Kind]; ok {
+			if _, ok := s.caches[cacheKey{res.Kind, checkNs}]; ok {
```
mandarjog: `checkNs` is not a loop variable, why is the 1st element of the namespaces special? why is it checked here in a loop?

jmuk: removed, reformed the loop structure a bit.

jmuk:
I created a benchmark test https://github.com/istio/mixer/pull/1324 to check the performance characteristics of this PR. Some results;

base (without this)
----

```
BenchmarkDummy/Init-32              1000           1402330 ns/op          127878 B/op       1199 allocs/op
BenchmarkDummy/InitWithValues-32                     500           3167422 ns/op         1492697 B/op      19844 allocs/op
BenchmarkDummy/List-32                              2000           1071960 ns/op          144309 B/op        451 allocs/op
BenchmarkDummy/Watch-32                            50000             31320 ns/op            2758 B/op         64 allocs/op
BenchmarkActual/Init-32                                1        2046302756 ns/op         1144920 B/op      14247 allocs/op
BenchmarkActual/InitWithValues-32                      1        2064746008 ns/op        26500400 B/op     300553 allocs/op
BenchmarkActual/List-32                             1000           1443901 ns/op          144280 B/op        451 allocs/op
BenchmarkActual/Watch-32                              30         139732466 ns/op           41810 B/op        564 allocs/op
```

with this
----
```
BenchmarkDummy/Init-32              1000           1577135 ns/op          258748 B/op       2371 allocs/op
BenchmarkDummy/InitWithValues-32                     300           4443140 ns/op          550099 B/op       7266 allocs/op
BenchmarkDummy/List-32                              3000            613945 ns/op          114876 B/op        471 allocs/op
BenchmarkDummy/Watch-32                            50000             30372 ns/op            2752 B/op         64 allocs/op
BenchmarkActual/Init-32                                1        6041307782 ns/op         1916616 B/op      23912 allocs/op
BenchmarkActual/InitWithValues-32                      1        6045596992 ns/op         7247784 B/op      82533 allocs/op
BenchmarkActual/List-32                             2000            796373 ns/op          114940 B/op        471 allocs/op
BenchmarkActual/Watch-32                              30         138374282 ns/op           41658 B/op        564 allocs/op
```

This actually reduces the memory usage of `Init`, but it gets much slower (because it needs to issue a fetch request per namespace). `List` performance is improved, because previously it needs to list the full data list and then filter unrelated data. `Watch` performance is unchanged.

Not appearing in the benchmark, but k8s client library will use http2 whenever possible, so the TCP connection will be reused among watchers.

By seeing these results, I think this PR is not very worthwhile to be merged.

PR #1187, opened 2017-09-01
 Enable dual quota check and add quota config

mandarjog:
1. Add Quota (and stdio) adapters to inventory.
    2. Update grpcServer to handle dual quota dispatch.
    3. Only apply quota when pre-condition succeeds and update tests.
    4. Add testdata/config/quota.yaml quota configuration.
    5. Do not evaluate empty selector predicates.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1187)


adapter/BUILD
```diff
@@ -16,6 +16,7 @@ go_library(
         "//adapter/ipListChecker:go_default_library",
         "//adapter/kubernetes:go_default_library",
         "//adapter/memQuota:go_default_library",
+        "//adapter/memquota2:go_default_library",
```
douglas-reid: do we need this here?  no, right?

testdata/config/quota.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: quota
+  namespace: istio-config-default
+spec:
+#  selector: request.headers["user"] != "admin"
+  actions:
+  - handler: memquota.memquota.istio-config-default
+    instances:
+    - requestcount.quota.istio-config-default
+---
```
douglas-reid: nit: add a newline.

mandarjog:
@douglas-reid PTAL

PR #1190, opened 2017-09-01
Refactor mixer server to expose a common setup method for test env

manlinl:
**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1190)


geeknoid:
This is fine for short term, but longer term I don't like having any dependencies on command-line tools from outside. If we want to reuse the server startup glue code, we should create a separate package for that purpose.

manlinl:
Agree, we should move the server startup code /mixer/pkg/somewhere

PR #1191, opened 2017-09-01
Rename //pkg/handler to //pkg/handlers

ZackButcher:
**Release note**:

```release-note
None
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1191)


ZackButcher:
We're gonna handle this in a different way.

PR #1195, opened 2017-09-01
Fixing post-presubmit failures due to test timeouts. See https://gith…

gnirodi:
…ub.com/istio/istio/issues/633

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1195)

### DIFF
--- a/pkg/config/crd/BUILD
+++ b/pkg/config/crd/BUILD
@@ -27,7 +27,7 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    size = "small",
+    size = "large",
     srcs = [
         "store_test.go",
     ],


jmuk:
It shouldn't take that much, and so I'm afraid that `size="large"` wouldn't solve the problem, anyways it would fail. Maybe it's some test flake. I'll check on my side.

ldemailly:
is there a medium ? why large ?

gnirodi:
This PR is no longer required, given that the root cause for timeouts were due to a race condition that was fixed via https://github.com/istio/mixer/pull/1210

PR #1197, opened 2017-09-01
Don't use encoding/gob to compute SHAs of config

ZackButcher:
Gob was barfing attempting to encode some nested protos due to requiring registration of types up front. For this usecase we just need SHAs of strings and proto fields, so we implement this directly for strings and use proto.Marshal to get a []byte from the protos

**Release note**:

```release-note
None
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1197)

### DIFF
--- a/pkg/runtime/handlerTable.go
+++ b/pkg/runtime/handlerTable.go
@@ -153,14 +153,14 @@ func encode(buf *bytes.Buffer, v interface{}) {
 			glog.Warningf("Failed to write %v to a buffer: %v", t, err)
 		}
 	case proto.Message:
-		if b, err := proto.Marshal(proto.Message(t)); err != nil {
+		if b, err := proto.Marshal(t); err != nil {
 			glog.Warningf("Failed to marshall %v into a proto: %v", t, err)
 		} else if _, err := buf.Write(b); err != nil {
 			glog.Warningf("Failed to write %v to buffer: %v", b, err)
 		}
 	default:
 		glog.Warningf("Fell into default case for v.(type): %#v; falling back to fmt.Fprintf()", t)
-		if _, err := fmt.Fprintf(buf, "%v", t); err != nil {
+		if _, err := fmt.Fprintf(buf, "%+v", t); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/runtime/handlerTable.go
```diff
@@ -145,20 +146,33 @@ func (t *handlerTable) initHandler(he *HandlerEntry) {
...
+			glog.Warningf("Failed to write %v to buffer: %v", t, err)
+		}
 	}
 }
 
 // computeSha for individual handler entries
 func (t *handlerTable) computeSha() {
-	var buff bytes.Buffer
-	enc := gob.NewEncoder(&buff)
+	buf := new(bytes.Buffer)
```
mandarjog: `var buf bytes.Buffer`  is preferred.

ZackButcher: We need a pointer to a buffer, not the buffer itself (`*bytes.Buffer` implements `io.Writer`, `bytes.Buffer` does not). `new(bytes.Buffer)` and `&bytes.Buffer{}` are identical.

pkg/runtime/handlerTable.go
```diff
@@ -145,20 +146,33 @@ func (t *handlerTable) initHandler(he *HandlerEntry) {
...
+		}
+	case proto.Message:
+		if b, err := proto.Marshal(proto.Message(t)); err != nil {
+			glog.Warningf("Failed to marshall %v into a proto: %v", t, err)
+		} else if _, err := buf.Write(b); err != nil {
+			glog.Warningf("Failed to write %v to buffer: %v", b, err)
+		}
+	default:
+		glog.Warningf("Fell into default case for v.(type): %#v; falling back to fmt.Fprintf()", t)
+		if _, err := fmt.Fprintf(buf, "%v", t); err != nil {
```
mandarjog: Use `%+v` so it also includes field names, so delimiting is better.

pkg/runtime/handlerTable.go
mandarjog: ok.

PR #1200, opened 2017-09-01
Add TCP attributes to appropriate manifests

douglas-reid:
This PR adds the attributes recently added to the global dictionary list to the manifests (old and new) within mixer.

**Release note**:

```release-note-none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1200)

### DIFF
--- a/testdata/config/attributes.yaml
+++ b/testdata/config/attributes.yaml
@@ -49,6 +49,24 @@ spec:
         valueType: STRING
       target.uid:
         valueType: STRING
+      connection.id:
+        valueType: STRING
+      connection.received.bytes:
+        valueType: INT64
+      connection.received.bytes_total:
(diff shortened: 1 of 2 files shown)


wora:
We should use prefix names everywhere. For example, creation time instead
of time created, total_bytes instead if bytes total, max length instead of
length max.

This is covered by Google API Design Guide. I wonder if we should open
source the design guide and invite Istio community to collaborate on it. If
we design the system using proto3, there are many benefits to have a high
quality design guide.

douglas-reid:
@wora this is just updating mixer's encoding of the attributes to match the canonical list in the istio/api repo. we should update that global list and then update everything else to match (including proxy/pilot).

PR #1203, opened 2017-09-02
remove dashes from crd names and update configs

mandarjog:
Remove dashes from all crd names.

We should add validation that rejects template and adapter names with dashes in a subsequent PR.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1203)

### DIFF
--- a/pkg/runtime/controller.go
+++ b/pkg/runtime/controller.go
@@ -84,10 +84,10 @@ type Controller struct {
 }
 
 // RulesKind defines the config kind name of mixer rules.
-const RulesKind = "mixer-rule"
+const RulesKind = "rule"
 
 // AttributeManifestKind define the config kind name of attribute manifests.
-const AttributeManifestKind = "attribute-manifest"
+const AttributeManifestKind = "attribute"
 
 // ResolverChangeListener is notified when a new resolver is created due to config change.
 type ResolverChangeListener interface {
--- a/pkg/runtime/controller.go
+++ b/pkg/runtime/controller.go
@@ -87,7 +87,7 @@ type Controller struct {
 const RulesKind = "rule"
 
 // AttributeManifestKind define the config kind name of attribute manifests.
-const AttributeManifestKind = "attribute"
+const AttributeManifestKind = "attributemanifest"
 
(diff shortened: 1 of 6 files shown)


pkg/runtime/controller.go
```diff
@@ -84,10 +84,10 @@ type Controller struct {
 }
 
 // RulesKind defines the config kind name of mixer rules.
-const RulesKind = "mixer-rule"
+const RulesKind = "rule"
 
 // AttributeManifestKind define the config kind name of attribute manifests.
-const AttributeManifestKind = "attribute-manifest"
+const AttributeManifestKind = "attribute"
```
douglas-reid: is `attribute` preferred to `manifest` ?

pkg/runtime/controller.go
```diff
@@ -84,10 +84,10 @@ type Controller struct {
 }
 
 // RulesKind defines the config kind name of mixer rules.
-const RulesKind = "mixer-rule"
+const RulesKind = "rule"
```
douglas-reid: `rule` seems fine. does it collide with anything else in istio?

mandarjog: There is `routeule`, so I think it is ok to claim `rule`

pkg/runtime/controller.go
mandarjog: I think so. Manifest is just a collection, the attribute is the important part.

ldemailly:
is there a description of the original problem with dash and pluralization somewhere ?

mandarjog:
@ldemailly [pluralization](https://github.com/kubernetes/kubernetes/issues/51639) is a minor problem. This PR addresses it for https://github.com/istio/mixer/pull/1167

kubectl does auto pluralization, so we do the same now.

Dashes has *not* been a problem for me in my testing, but @kyessenov and @andraxylia have mentioned that they had problems. At this point I want to be extra cautious, K8s is hypersensitive to naming.

*All* systems are agreeable to identifiers with *all lowercase letters*, they are the chicken of naming, no one has a problem eating it.

ldemailly:
I think you meant tofu :)

Issue #1209, opened 2017-09-05
pkg/config/crd test is flaky


jmuk:
I got some reports saying the test may fail with errors like:
```
W0904 15:52:31.360] ____From Testing //pkg/config/crd:go_default_test:
I0904 15:52:31.461] ==================== Test output for //pkg/config/crd:go_default_test:
I0904 15:52:31.461] --- FAIL: TestStore (1.00s)
I0904 15:52:31.462] 	store_test.go:181: Got context deadline exceeded, Want nil
I0904 15:52:31.462] 	store_test.go:185: Got not found, Want nil
I0904 15:52:31.462] 	store_test.go:188: Got map[], Want map[adapter:noop name:default]
I0904 15:52:31.463] 	store_test.go:192: Got map[], Want map[default.Handler.istio-mixer-testing:map[]]
I0904 15:52:31.463] 	store_test.go:200: Got not found, Want nil
I0904 15:52:31.463] 	store_test.go:203: Got map[], Want map[adapter:noop2 name:default]
```

It seems that `waitFor` fails to receive the expected events.

https://github.com/istio/istio/issues/633 indicates that it may still fail to receive the events even though we increase the timeout. Needs to investigate why.

jmuk:
Thanks, closing...
cc @gnirodi

PR #1210, opened 2017-09-05
create fake watchers earlier for CRD tests

jmuk:
Maybe there's a race between the invocation of `WatchFunc` and
the `put` method which injects the test data. If `put` comes
earlier, that event never comes to the watching channel (because
the watcher isn't there yet), therefore `waitFor` will fail to
receive the expected event.

For #1209

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1210)


jmuk:
seems not fixing the issue... rather it increases the chance of the timeout. 😢 
will look into the details more, but closing this PR for now.

PR #1212, opened 2017-09-05
Fix crd test flakeness

jmuk:
This should fix the test flakeness and potential timeout for `pkg/config/crd`. At least this should remove
the timeout failure which hides the actual problem if there are.

- similar to #1210, creates the watcher instance earlier, to resolve the race condition between invocation of `WatchFunc` (of creating watcher) and `put` (use of watcher)
- replace `FakeWatcher` by `RaceFreeFakeWatcher`. Otherwise, `put` may be blocked due to write of channel with no listeners.
- restructure `TestCrdsRetryAsynchronously` a bit; make the put/watch/get testing after verifying the store creates the cache. Otherwise, the cache isn't ready yet, and the test can fail.
- minor cleanup to cancel the context for `Init` after the test case finishes.

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1212)

### DIFF
--- a/pkg/config/crd/store_test.go
+++ b/pkg/config/crd/store_test.go
@@ -67,19 +67,19 @@ type dummyListerWatcherBuilder struct {
 func (d *dummyListerWatcherBuilder) build(res metav1.APIResource) cache.ListerWatcher {
 	w := watch.NewRaceFreeFake()
 	d.mu.Lock()
-	defer d.mu.Unlock()
 	d.watchers[res.Kind] = w
+	d.mu.Unlock()
 
 	return &cache.ListWatch{
 		ListFunc: func(metav1.ListOptions) (runtime.Object, error) {
-			d.mu.RLock()
-			defer d.mu.RUnlock()
 			list := &unstructured.UnstructuredList{}
+			d.mu.RLock()
 			for k, v := range d.data {
 				if k.Kind == res.Kind {
 					list.Items = append(list.Items, *v)
 				}
 			}
+			d.mu.RUnlock()
 			return list, nil
 		},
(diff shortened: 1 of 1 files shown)


pkg/config/crd/store_test.go
```diff
@@ -61,10 +61,15 @@ func createFakeDiscovery(*rest.Config) (discovery.DiscoveryInterface, error) {
 type dummyListerWatcherBuilder struct {
 	mu       sync.RWMutex
 	data     map[store.Key]*unstructured.Unstructured
-	watchers map[string]*watch.FakeWatcher
+	watchers map[string]*watch.RaceFreeFakeWatcher
 }
 
 func (d *dummyListerWatcherBuilder) build(res metav1.APIResource) cache.ListerWatcher {
+	w := watch.NewRaceFreeFake()
+	d.mu.Lock()
```
guptasu: Any reason to not do simple lock-> edit -> unlock..
defer will keep the lock for longer that needed (tll the entire mtd ends)

d.mu.Lock()
d.watchers[res.Kind] = w
d.mu.Unlock()
?

Same for the defer Unlocks below

jmuk: Changed here and below `RLock`.  I think defer is better for `put` and `delete`, since it can early-exit and use of defer simplifies the code.

jmuk:
@mandarjog 

the smoketest succeeds with my private GKE cluster, as follows:
```
% ./tests/e2e.sh --mixer_hub=gcr.io/istio-testing --mixer_tag=e0450e5f473a31281b13ed4bb2d3f084a51213e2
(snip)
ests Summary
```

Please merge this, as we discussed offline.

PR #1217, opened 2017-09-06
Add adpater test framework and sample tests

manlinl:
- Add testenv package which allows adapter developers to instantiate Mixer server and client in test proc for easy test and debug.
- Add a sample test based on denier

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1217)

### DIFF
--- a/cmd/server/cmd/BUILD
+++ b/cmd/server/cmd/BUILD
@@ -13,6 +13,7 @@ go_library(
     ],
     visibility = [
         "//cmd:__subpackages__",
+        "//test/testenv:__subpackages__",
     ],
     deps = [
         "//adapter:go_default_library",
--- a/cmd/server/cmd/test_server.go
+++ b/cmd/server/cmd/test_server.go
@@ -46,11 +46,12 @@ var defaultSeverArgs = serverArgs{
 
 // SetupTestServer sets up a test server environment
 func SetupTestServer(info map[string]template.Info, adapters []adapter.InfoFn, configStoreURL string, configStore2URL string,
-	configDefaultNamespace string, configIdentityAttributeDomain string) *ServerContext {
+	configDefaultNamespace string, configIdentityAttributeDomain string, useAst bool) *ServerContext {
 	sa := defaultSeverArgs
 	sa.configStoreURL = configStoreURL
 	sa.configStore2URL = configStore2URL
 	sa.configDefaultNamespace = configDefaultNamespace
 	sa.configIdentityAttributeDomain = configIdentityAttributeDomain
+	sa.useAst = useAst
 	return setupServer(&sa, info, adapters, shared.Printf, shared.Fatalf)
 }
--- a/test/testenv/BUILD
+++ b/test/testenv/BUILD
@@ -1,0 +1,38 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test", "go_binary")
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "testenv.go",
+    ],
+    deps = [
+        "//cmd/server/cmd:go_default_library",
+        "//cmd/shared:go_default_library",
+        "//pkg/adapter:go_default_library",
+        "//pkg/template:go_default_library",
+        "@com_github_istio_api//:mixer/v1",  # keep
+        "@org_golang_google_grpc//:go_default_library",
+    ],
+)
+
+go_test(
+    name = "go_default_test",
+    size = "small",
+    srcs = ["testenv_test.go"],
+    data = [
+        "//testdata",
+    ],
+    library = ":go_default_library",
+    deps = [
+        "//adapter/denier:go_default_library",
+        "//pkg/adapter:go_default_library",
+        "//pkg/attribute:go_default_library",
+        "//template:go_default_library",
+        "//test/testenv:go_default_library",
+        "@com_github_istio_api//:mixer/v1",  # keep
+        "@com_github_pborman_uuid//:go_default_library",
+        "@org_golang_x_net//context:go_default_library",
+    ],
+)
--- a/test/testenv/testenv.go
+++ b/test/testenv/testenv.go
@@ -1,0 +1,84 @@
+package testenv
+
+import (
+	"io"
+	"net"
+
+	"google.golang.org/grpc"
+	mixerpb "istio.io/api/mixer/v1"
+	"istio.io/mixer/cmd/server/cmd"
+	"istio.io/mixer/cmd/shared"
+	"istio.io/mixer/pkg/adapter"
+	"istio.io/mixer/pkg/template"
+)
+
+// TestEnv interface
+type TestEnv interface {
+	io.Closer
+	CreateMixerClient() (mixerpb.MixerClient, *grpc.ClientConn, error)
+}
+
+type testEnv struct {
+	addr         string
+	mixerContext *cmd.ServerContext
+	shutdown     chan struct{}
+}
+
+// Args includes the required args to initialize Mixer server.
+type Args struct {
+	MixerServerAddr               string
+	ConfigStoreURL                string
+	ConfigStore2URL               string
+	ConfigDefaultNamespace        string
+	ConfigIdentityAttributeDomain string
+	UseAstEvaluator               bool
+}
+
+// NewEnv creates a TestEnv instance.
+func NewEnv(args *Args, info map[string]template.Info, adapters []adapter.InfoFn) (TestEnv, error) {
+	lis, err := net.Listen("tcp", args.MixerServerAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	context := cmd.SetupTestServer(info, adapters, args.ConfigStoreURL, args.ConfigStore2URL,
+		args.ConfigDefaultNamespace, args.ConfigIdentityAttributeDomain, args.UseAstEvaluator)
+	shutdown := make(chan struct{})
+
+	go func() {
+		shared.Printf("Start test Mixer on:%v\n", lis.Addr())
+		{
+			defer context.GP.Close()
+			defer context.AdapterGP.Close()
+			if err := context.Server.Serve(lis); err != nil {
+				shared.Printf("Mixer Shutdown: %v\n", err)
+			}
+		}
+		shutdown <- struct{}{}
+	}()
+
+	return &testEnv{
+		addr:         lis.Addr().String(),
+		mixerContext: context,
+		shutdown:     shutdown,
+	}, nil
+}
+
+// CreateMixerClient returns a Mixer Grpc client.
+func (env *testEnv) CreateMixerClient() (mixerpb.MixerClient, *grpc.ClientConn, error) {
+	conn, err := grpc.Dial(env.addr, grpc.WithInsecure())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return mixerpb.NewMixerClient(conn), conn, nil
+}
+
+// Close cleans up TestEnv.
+func (env *testEnv) Close() error {
+	env.mixerContext.Server.GracefulStop()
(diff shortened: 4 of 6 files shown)


test/testenv/BUILD
```diff
@@ -0,0 +1,38 @@
...
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "testenv.go",
+    ],
+    deps = [
+        "//cmd/server/cmd:go_default_library",
+        "//cmd/shared:go_default_library",
+        "//pkg/handler:go_default_library",
+        "//template:go_default_library",
```
guptasu: Does this mean adapters can only test againt the built in templates ? It seems like the supported templates should be passed in from the test and not be part of the test environment

test/testenv/testenv.go
```diff
@@ -0,0 +1,82 @@
+package testenv
```
guptasu: Should the directory structure be called test/framework/env.go

test/testenv/testenv.go
```diff
@@ -0,0 +1,82 @@
...
+// Args includes the required args to initialize Mixer server.
+type Args struct {
+	ConfigStoreURL                string
+	ConfigStore2URL               string
+	ConfigDefaultNamespace        string
+	ConfigIdentityAttributeDomain string
+}
+
+// NewTestEnv creates a TestEnv instance.
+func NewTestEnv(args *Args, adapters []handler.InfoFn) (TestEnv, error) {
```
guptasu: test.NewEnv ?

test/testenv/testenv.go
```diff
@@ -0,0 +1,82 @@
...
+	lis, err := net.Listen("tcp", "localhost:0")
```
guptasu: Should we make '"localhost:0"' a param ?

test/testenv/testenv.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	context := cmd.SetupTestServer(template.SupportedTmplInfo, adapters, args.ConfigStoreURL, args.ConfigStore2URL,
```
guptasu: I think this should be a parameter 'template.SupportedTmplInfo', Will allow more flexibility for testing.

test/testenv/testenv.go
```diff
@@ -0,0 +1,82 @@
...
+		args.ConfigDefaultNamespace, args.ConfigIdentityAttributeDomain)
+	shutdown := make(chan struct{})
+
+	go func() {
+		shared.Printf("Start test mixer on:%v", lis.Addr())
```
guptasu: should we print mixer closing as well.. Consider pasting the output of the console (same that adapter developer will see) for the sample test you have.

test/testenv/testenv_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/pborman/uuid"
+	"golang.org/x/net/context"
+	mixerapi "istio.io/api/mixer/v1"
+	"istio.io/mixer/adapter/denier"
+	"istio.io/mixer/pkg/attribute"
+	"istio.io/mixer/pkg/handler"
+)
+
+func closeHelper(c io.Closer) {
```
guptasu: I think all the helper methods in this file should be in the testenv.go.. because every adapter test will be using them. It will be super slick if all they have to type is the content of your sample 'TestDenierAdapter'

test/testenv/testenv_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	code := m.Run()
+	os.Exit(code)
+}
+
+func TestDenierAdapter(t *testing.T) {
+	configStore, err := buildTestConfigStore([]string{
+		"../../testdata/config/attributes.yaml",
```
guptasu: This is tricky.. I would recomment asking for text, creating file on the fly , putting them in temp dir and running from it.. Ideally adapter developer should just be writing TestDenierAdapter. For reference https://github.com/istio/mixer/pull/1176/files#diff-8349006db2c242fd7424e1dfb3295840R32

manlinl: This just serves as an example, so I use a checked-in config instead of have the config inline in test. But the test framework itself doesn't prevent adapter developer from doing what you describe. Once mem based config store is in, that should be preferred way.

test/testenv/testenv_test.go
manlinl: agree

test/testenv/testenv.go
manlinl: agree.

test/testenv/testenv.go
manlinl: This means choosing a open port on loop back interface. Test developer then doesn't need to scratch their heads to figure out what port they should use.

test/testenv/testenv.go
manlinl: My thought is make 'testenv' the framework name. 'framework' is too general for a package name following guide here https://golang.org/doc/effective_go.html#package-names

test/testenv/BUILD
manlinl: We can make that possible.

test/testenv/testenv.go
manlinl: Done

test/testenv/testenv.go
manlinl: Shutdown message is printed in the if-block below.

test/testenv/testenv.go
manlinl: localhost:0 chooses a free port on loop back interface. So it has some flexibility. But it doesn't hurt to make it an argument.

test/testenv/testenv_test.go
manlinl: I am hoping the help methods of copying files to creating config store aren't not needed in the future by using memory based config store. closerHelper could be generally helpful, so it could be move to some common package. For now, I would prefer keep the test framework compact.

test/testenv/BUILD
manlinl: Done

cmd/server/cmd/BUILD
```diff
@@ -13,6 +13,7 @@ go_library(
     ],
     visibility = [
         "//cmd:__subpackages__",
+        "//test/testenv:__subpackages__",
```
guptasu: Sorry didn't mentioned this before. I am not excited with the fact that this library is inside test dir, since it is something that adapter developers will use. @ZackButcher @douglas-reid @mandarjog thoughts ? What should be the right house for the mixer testing framework

cmd/server/cmd/test_server.go
```diff
@@ -41,7 +41,7 @@ var defaultSeverArgs = serverArgs{
 	configFetchIntervalSec:        3,
 	configIdentityAttribute:       "target.service",
 	configIdentityAttributeDomain: "",
-	useAst: false,
+	useAst: true,
```
guptasu: I think for now this should be a paramter.. mainly because we want more test that will be written in next 1-2 using your framework to by default exercise the IL code.. and find bugs similar to what you discovered.. so, let's make this false by default and let each test override it explicitly.

guptasu: Once IL is the default, the flag goes away and we will clear the test then

test/testenv/testenv.go
guptasu: @douglas-reid since he is good with conventions.

test/testenv/testenv_test.go
guptasu: SG. Let's use the checked in config for this test. Later as we find more common patterns, we might find code that can belong to the test env framework

cmd/server/cmd/BUILD
manlinl: As we discussed, for now, let's leave it in test dir since it has a dependency on cmd/server/cmd package. Post 0.2, we can refactor the code and move to pkg.

ZackButcher: I think this is reasonable: `//test` is probably the right place for the test framework to live. Long term we should refactor the test framework to depend on `//pkg/api` and do the wiring up itself, rather than rely on `//cmd/server`'s wiring. I think this will lead to a nicer, more general interface to write tests against; I'd like to see us use essentially this framework to execute our own integration Mixer integration tests down the line.

test/testenv/testenv.go
```diff
@@ -0,0 +1,83 @@
...
+type Args struct {
+	MixerServerAddr               string
+	ConfigStoreURL                string
+	ConfigStore2URL               string
+	ConfigDefaultNamespace        string
+	ConfigIdentityAttributeDomain string
+}
+
+// NewEnv creates a TestEnv instance.
+func NewEnv(args *Args, info map[string]template.Info, adapters []handler.InfoFn) (TestEnv, error) {
```
ZackButcher: Just `New`, we're already in package `testenv` so at usage sites this will read `testenv.New` rather than `testenv.NewEnv`, which stutters.

test/testenv/testenv_test.go
ZackButcher: This is fine for now, but when we go back to refactor we should pull these files in via bazel rather than relying on directory structure. (I know you pull in a file group, but there's some magic we can use to move the files to the same dir as the test at build/execution time.)

Longer term I'd like to see us move to an interface where the entire test body looks something like:
```go
client, _ := testenv.New(MyAdapter.GetInfo, "myfirstconfig.yaml", "mysecondconfig.yaml", ...)

attr := map[string]interface{}{"request.headers": map[string]string{"clnt": "abc"}}
bag := getAttrBag(attr)
request := mixerapi.CheckRequest{Attributes: bag}
resq, err := client.Check(context.Background(), &request)
if err != nil {
	t.Errorf("fail to send check to Mixer %v", err)
}
if resq.Precondition.Status.Code != 7 {
	t.Error(`resq.Precondition.Status.Code != 7`)
}
```

And all the cleanup is handled by the framework.

test/testenv/testenv_test.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return err
+	}
+	defer closeHelper(out)
+	_, err = io.Copy(out, in)
+	return err
+
+}
+
+func buildTestConfigStore(relativePaths []string) (string, error) {
```
ZackButcher: `buildConfigStore`, we're already in a _test.go file so we know it has to be for a test.

test/testenv/testenv_test.go
manlinl: Ideally, test framework should use mem based store so no clean up on FS is needed. But this would be post v0.2

test/testenv/testenv_test.go
manlinl: done

test/testenv/testenv.go
manlinl: That would prevent us from adding other New

test/testenv/testenv.go
```diff
@@ -0,0 +1,84 @@
...
+	MixerServerAddr               string
+	ConfigStoreURL                string
+	ConfigStore2URL               string
+	ConfigDefaultNamespace        string
+	ConfigIdentityAttributeDomain string
+	UseAstEvaluator               bool
+}
+
+// NewEnv creates a TestEnv instance.
+func NewEnv(args *Args, info map[string]template.Info, adapters []adapter.InfoFn) (TestEnv, error) {
```
guptasu: Rename to New ?

test/testenv/testenv.go
ZackButcher: Sure - when we add another thing to this package that needs a `New*` method, rename it. For now you have a single public _thing_ that this package does, so call this method New so it reads naturally.

test/testenv/testenv.go
manlinl: What if we add something later in the package that need a new method?

PR #1221, opened 2017-09-06
Update MixerIL to return "" when lookup from a stringmap fails.

ozevren:
- Introduce new opcodes, NLookup, ANLookup that returns default value
when the lookup fails.
- Update the compiler to use these new opcodes when generating code.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1221)

### DIFF
--- a/pkg/il/builder.go
+++ b/pkg/il/builder.go
@@ -202,6 +202,11 @@ func (f *Builder) Lookup() {
 	f.op0(Lookup)
 }
 
+// NLookup appends the "nlookup" instruction to the byte code.
+func (f *Builder) NLookup() {
+	f.op0(NLookup)
+}
+
 // TLookup appends the "tlookup" instruction to the byte code.
 func (f *Builder) TLookup() {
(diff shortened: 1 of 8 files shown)


douglas-reid:
@ozevren just as an fyi: i think you may have missed the announcement. if you can run the e2e successfully on your own cluster and have one of us verify, we will merge around the failures on prow.

ozevren:
Thanks. Good to know. My local setup is currently broken, I'll try
rerunning the e2e tests on GitHub (at least for today) and update my local
setup tomorrow if I cannot jam it through that way. :)

ZackButcher:
FYI the tests are in a bad state atm, the chances that you get a passing run are essentially zero. Get e2e.sh invocation from one of the failed e2e smoketest runs and execute it on your own cluster, and get someone else to execute it on theirs and we can force merge in.

ZackButcher:
@mandarjog can you force merge, it's good to go.

PR #1228, opened 2017-09-07
Add labels processing per spec

mandarjog:
1. Metadata.labels now works.
2. `istio-protocol: tcp` is the only label that is processed. A rule marked as `tcp` will only be evaluated if `context.protocol == tcp`
3. if namespace is omitted from fully qualified name of a handler or an instance, the rule's namespace is used.
    This shortens the configs. 

```
apiVersion: "config.istio.io/v1alpha2"
kind: metric
metadata:
  name: bytessent
spec:
  value: connection.sent.bytes
  dimensions:
    source: source.labels["app"] | "unknown"
    target: target.service | "unknown"
    service: target.labels["app"] | "unknown"
    method: request.path | "unknown"
    version: target.labels["version"] | "unknown"
    response_code: response.code | 200
  monitored_resource_type: '"UNSPECIFIED"'

---
apiVersion: "config.istio.io/v1alpha2"
kind: rule
metadata:
  name: tcpmetrics
  labels:
    istio-protocol: tcp
spec:
  actions:
  - handler: handler.prometheus
    instances:
    - bytessent.metric
```

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1228)

### DIFF
--- a/adapter/memquota2/dedup.go
+++ b/adapter/memquota2/dedup.go
@@ -105,8 +105,8 @@ func (du *dedupUtil) reapDedup() {
 	du.oldDedup = du.recentDedup
 	du.recentDedup = t
 
-	if len(t) > 0 && du.logger.VerbosityLevel(6) {
-		du.logger.Infof("Running reaper to reclaim %d old deduplication entries", len(t))
+	if len(t) > 0 && du.logger.VerbosityLevel(4) {
+		du.logger.Infof("Running repear to reclaim %d old deduplication entries", len(t))
 	}
 
 	// TODO: why isn't there a O(1) way to clear a map to the empty state?!
--- a/adapter/memquota2/dedup.go
+++ b/adapter/memquota2/dedup.go
@@ -105,7 +105,7 @@ func (du *dedupUtil) reapDedup() {
 	du.oldDedup = du.recentDedup
 	du.recentDedup = t
 
-	if len(t) > 0 && du.logger.VerbosityLevel(4) {
+	if du.logger.VerbosityLevel(4) {
 		du.logger.Infof("Running repear to reclaim %d old deduplication entries", len(t))
 	}
 
--- a/pkg/config/store/queue.go
+++ b/pkg/config/store/queue.go
@@ -48,7 +48,7 @@ func (q *eventQueue) convertValue(ev BackendEvent) (Event, error) {
 		return Event{}, err
 	}
 	if ev.Value == nil {
-		return Event{}, err
+		return Event{Key: ev.Key, Type: ev.Type}, nil
 	}
 	if err = convert(ev.Value.Spec, pbSpec); err != nil {
 		return Event{}, err
--- a/pkg/config/store/queue.go
+++ b/pkg/config/store/queue.go
@@ -47,10 +47,14 @@ func (q *eventQueue) convertValue(ev BackendEvent) (Event, error) {
 	if err != nil {
 		return Event{}, err
 	}
-	if err = convert(ev.Value, pbSpec); err != nil {
+	if err = convert(ev.Value.Spec, pbSpec); err != nil {
 		return Event{}, err
 	}
-	return Event{Key: ev.Key, Type: ev.Type, Value: pbSpec}, nil
+	return Event{Key: ev.Key, Type: ev.Type, Value: &Resource{
+		Key:      ev.Key,
+		Metadata: ev.Value.Metadata,
+		Spec:     pbSpec,
+	}}, nil
 }
 
 func (q *eventQueue) run() {
--- a/pkg/config/store/queue.go
+++ b/pkg/config/store/queue.go
@@ -51,7 +51,6 @@ func (q *eventQueue) convertValue(ev BackendEvent) (Event, error) {
 		return Event{}, err
 	}
 	return Event{Key: ev.Key, Type: ev.Type, Value: &Resource{
-		Key:      ev.Key,
 		Metadata: ev.Value.Metadata,
 		Spec:     pbSpec,
 	}}, nil
--- a/pkg/config/store/queue.go
+++ b/pkg/config/store/queue.go
@@ -47,6 +47,9 @@ func (q *eventQueue) convertValue(ev BackendEvent) (Event, error) {
 	if err != nil {
 		return Event{}, err
 	}
+	if ev.Value == nil {
+		return Event{}, err
+	}
 	if err = convert(ev.Value.Spec, pbSpec); err != nil {
 		return Event{}, err
 	}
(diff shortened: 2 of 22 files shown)


testdata/config/metrics.yaml
```diff
@@ -12,6 +12,7 @@ spec:
     method: request.path | "unknown"
     version: target.labels["version"] | "unknown"
     response_code: response.code | 200
+  monitored_resource_type: '"UNSPECIFIED"'
```
guptasu: @mandarjog is it possible to submit change to this file separately first (force push if you want). This change is blocking manual testing of mixer using the testdata configs (for the report templates only).

@douglas-reid FYI, since he ran into this issue last night.

adapter/memquota2/dedup.go
```diff
@@ -105,7 +105,7 @@ func (du *dedupUtil) reapDedup() {
 	du.oldDedup = du.recentDedup
 	du.recentDedup = t
 
-	if du.logger.VerbosityLevel(4) {
+	if len(t) > 0 && du.logger.VerbosityLevel(4) {
```
douglas-reid: we have at least 3 PRs with 3 different fixes for this.  let's settle on one approach.

testdata/config/prometheus.yaml
```diff
@@ -34,16 +34,24 @@ spec:
         numFiniteBuckets: 8
         scale: 1
         growthFactor: 10
+  - name: tcp_bytes_sent
+    instance_name: bytessent.metric.istio-config-default
+    kind: DISTRIBUTION
+    label_names: [ source, target, service, method, version, response_code ]
```
douglas-reid: if possible, i'd prefer to use the metrics definitions in my outstanding PR.

In particular, things like `method` and `response_code` are not applicable, and I'd like to clarify the names around `source`, `target`, and which one `version` applies to.  Also, I think this should be a counter (and not a distribution), as this will like accumulate through the course of a connection (and counter would enable easy bytes/second charts, etc.)

testdata/config/tcpMetrics.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: "config.istio.io/v1alpha2"
+kind: metric
+metadata:
+  name: bytessent
+  namespace: istio-config-default
+spec:
+  value: connection.sent.bytes
+  dimensions:
```
douglas-reid: please see my above comment re: metrics shape.

adapter/memquota2/dedup.go
mandarjog: I will remove this file from my PR, should not have checked it in. It just locally annoying me.

testdata/config/prometheus.yaml
mandarjog: I am happy to exclude these files from the PR, if your PR already does it in a different way.

testdata/config/tcpMetrics.yaml
mandarjog: same comment as before.

adapter/memquota2/dedup.go
douglas-reid: yeah, it is super annoying. i'm fine with your fix... or zack's making it level 6... or which ever... it just needs to go in.

testdata/config/prometheus.yaml
douglas-reid: cool. even if they go in, i can rebase and reshape them in my TCP PR.

pkg/config/store/store2_test.go
```diff
@@ -100,12 +109,19 @@ func TestStore2(t *testing.T) {
...
+			Spec: map[string]interface{}{"name": "default", "adapter": "noop"},
+		},
+	}
+	wantEv := Event{Key: k, Type: Update,
+		Value: &Resource{
+			Spec: want,
+		},
+	}
+	ev := <-ch
+	if !reflect.DeepEqual(ev, wantEv) {
```
jmuk: I prefer the previous structure -- `if ev := <-ch; !reflect.`... would be fine (and more concise), I think)

pkg/runtime/controller.go
```diff
@@ -368,15 +386,25 @@ func (c *Controller) processRules(handlerConfig map[string]*cpb.Handler,
...
-	instanceConfig map[string]*cpb.Instance, ht *handlerTable) map[adptTmpl.TemplateVariety]map[string]*Action {
+	instanceConfig map[string]*cpb.Instance, ht *handlerTable, namespace string) map[adptTmpl.TemplateVariety]map[string]*Action {
 
 	actions := make(map[adptTmpl.TemplateVariety]map[string]*Action)
 
 	for _, ic := range acts {
 		var hc *cpb.Handler
+		if !isFQN(ic.Handler) {
+			ic.Handler = ic.Handler + "." + namespace
+		}
```
jmuk: This section is not tested in the controller_test.go, right?  It's better to add tests for them.

pkg/runtime/controller.go
```diff
@@ -336,16 +351,19 @@ func (c *Controller) processRules(handlerConfig map[string]*cpb.Handler,
...
 		if k.Kind != RulesKind {
 			continue
 		}
+
+		cfg := obj.Spec
 		rulec := cfg.(*cpb.Rule)
 		rule := &Rule{
 			selector: rulec.Selector,
 			name:     k.Name,
+			rtype:    resourceType(obj.Metadata.Labels),
```
jmuk: can you write test patterns for this?

testdata/config/tcpMetrics.yaml
mandarjog: actually I will update this, it has an example of `istio-protocol: tcp` I don't want to remove that file.

pkg/config/store/queue.go
```diff
@@ -47,10 +47,16 @@ func (q *eventQueue) convertValue(ev BackendEvent) (Event, error) {
 	if err != nil {
 		return Event{}, err
 	}
-	if err = convert(ev.Value, pbSpec); err != nil {
+	if ev.Value == nil {
 		return Event{}, err
```
jmuk: This line will return an empty event with no error. Please update as:
```go
return Event{Key: ev.Key, Type: ev.Type}, nil
```

guptasu:
e2e passed on my cluster
```
./e2e.sh --mixer_hub=gcr.io/istio-testing --mixer_tag=a764e1b3cd766b896fb0fd4d3d4382e3f14b4b59
Tests Summary
```

mandarjog:
@jmuk @douglas-reid addressed review comments. Removed several yaml files so other PRs can check them in instead.

testdata/config/tcpMetrics.yaml
douglas-reid: cool. i'll adjust the other PR as necessary when I get to it. thanks!

PR #1232, opened 2017-09-07
JWT Token adapter

RoeyMaor:
**Release note**:

```release-note
JWT Token attribute generating adapter.
used for token validation and policy definition based on token claims and metadata.

The adapter is configured with a list of token issuers, from which it accepts tokens and fetches public keys, according to the JWT,JWK standards.

See descriptors.yml for the list of attributes generated by the adapter

Examples:

example adapter params (in adapters.yml):
issuers:
-  name: "local" (needs to match the value of the token's "iss" claim)
   pubKeyUrl: "https://jwktestserver:21356/keys"
   claimNames: ['name','locations.us.ca','servers.haifa'] (requested claims to output, with hierarchical . notation)
   claimRenames:
     haifa: servers.haifa (ability to rename requested claims in the output)

example policy based on adapter's attributes:
rules:
- aspects:
  - kind: denials
  selector: request.token.valid==false || request.token.claims["name"]!="johnDoe"

```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1232)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -615,3 +615,15 @@ git_repository(
     commit = "b0822890273f91d5aa8c40ea1a89ba01e0f0ee9d",  # Aug 22, 2017
     remote = "https://github.com/istio/test-infra.git",
 )
+
+go_repository(
+    name = "com_github_asaskevich_govalidator",
+    commit = "15028e809df8c71964e8efa6c11e81d5c0262302", # Jul 30, 2017
+    importpath = "github.com/asaskevich/govalidator",
+)
+
+go_repository(
+    name = "com_github_dgrijalva_jwt-go",
+    commit = "a539ee1a749a2b895533f979515ac7e6e0f5b650", # June 8, 2017
+    importpath = "github.com/dgrijalva/jwt-go",
+)
--- a/adapter/BUILD
+++ b/adapter/BUILD
@@ -24,6 +24,7 @@ go_library(
         "//adapter/statsd2:go_default_library",
         "//adapter/stdio:go_default_library",
         "//adapter/stdioLogger:go_default_library",
+        "//adapter/token:go_default_library",
         "//pkg/adapter:go_default_library",
     ],
 )
--- a/adapter/inventory.go
+++ b/adapter/inventory.go
@@ -26,6 +26,7 @@ import (
 	"istio.io/mixer/adapter/serviceControl"
 	"istio.io/mixer/adapter/statsd"
 	"istio.io/mixer/adapter/stdioLogger"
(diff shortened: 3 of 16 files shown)


mandarjog:
Hey @RoeyMaor, thanks for your PR.
We are nearing 0.2 release.
Mixer adapter model has completely changed from 0.1 to 0.2.
So unfortunately we will not be able to accept an adapter based on the 0.1 model.

Here is our a [draft of our adapter developer's guide](https://docs.google.com/document/d/1rKPt2Z2acy4pRwcvScPa-Na-EnU-6poGR-Rj_fZtaIc/edit?ts=59a708e6#heading=h.7j0k8yg4uz4m).
Take a look at the walkthrough section.
You may also look at https://github.com/istio/mixer/blob/master/adapter/list/list.go and other new adapter.

qiwzhang:
Hey @RoeyMaor,

There is another plan to implement JWT token authentication in Istio/proxy. The prototype code is here: https://github.com/istio/proxy/tree/master/src/envoy/auth

Compare to implementing in the Mixer side.  Proxy side has following benefits
1) not need to pass Token to Mixer,  some users may consider Token to be secret
2) Easier to pass the token payload to backend service.

RoeyMaor:
Hi,
@mandarjog It looks like this design change did not include "attributes" kind adapters like kubernetes which stayed in their 0.1 design form in 0.2 as well.
@qiwzhang The sensitive information issue in mixer is not a token issue but an issue in general, api keys, sensitive cookies etc are already being passed to mixer, this is a general problem that needs to be solved.
I think the two solutions don't necessarily contradict each other. Mixer is intended to be istio's policy decision point, and deciding that policies that are related to X,Y,Z are to be decided outside of mixer, by components A,B,C doesn't fit with that philosophy, and makes it harder for the operator. When deciding on policies in mixer, one could use mixer's already available arsenal of policy definition adapters, like in the example I gave. In addition, Implementing in mixer means that a change requires to redeploy mixer only, and not all relevant sidecars.
Both solutions have their advantages and disadvantages.

geeknoid:
@guptasu is currently designing a revamped model for attribute generating adapters. Docs are available on the team drive. I don't know if we want to get this change in first, or wait till enough of the new model has been deployed.

@apicl is leading an effort to introduce API management functionality to Istio. One of the features will be JWT decoding in the proxy. We originally wanted to do this kind of processing all in Mixer, but it turns out to get decent caching in the proxy, we need to decode JWT over there. Given this, do you think it is still valuable to provide an adapter that also decodes JWTs?

geeknoid:

Thanks.

PR #1235, opened 2017-09-07
Fix bugs in Stackdriver implementation and correct configs.

ZackButcher:
The version of the Stackdriver Logging PR was missing some commits (turns out I had split them off to their own branch but I thought I'd made them in the PR branch).

We correct the configuration for Stackdriver so it actually works, and introduce new Stackdriver specific metrics filling in the monitored resource fields.

I've verified both metrics and logs are correctly produced and ingested by the Stackdriver APIs using these configs.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1235)

### DIFF
--- a/adapter/stackdriver/log/log.go
+++ b/adapter/stackdriver/log/log.go
@@ -26,6 +26,7 @@ import (
 	"cloud.google.com/go/logging"
 	xctx "golang.org/x/net/context"
 	"google.golang.org/api/option"
+	"google.golang.org/genproto/googleapis/api/monitoredres"
 
 	"istio.io/mixer/adapter/stackdriver/config"
 	"istio.io/mixer/adapter/stackdriver/helper"
@@ -112,7 +113,7 @@ func (b *builder) Build(ctx context.Context, env adapter.Env) (adapter.Handler, 
 			log:    client.Logger(name).Log,
 		}
 	}
-	return &handler{client: client, now: time.Now, l: logger}, nil
+	return &handler{client: client, now: time.Now, l: logger, info: infos}, nil
 }
 
 func (h *handler) HandleLogEntry(_ context.Context, values []*logentry.Instance) error {
@@ -131,13 +132,22 @@ func (h *handler) HandleLogEntry(_ context.Context, values []*logentry.Instance)
 		payload := buf.String()
 		pool.PutBuffer(buf)
 
-		linfo.log(logging.Entry{
+		e := logging.Entry{
 			Timestamp:   h.now(), // TODO: use timestamp on Instance when timestamps work
 			Severity:    logging.ParseSeverity(v.Severity),
 			Labels:      toLabelMap(linfo.labels, v.Variables),
 			Payload:     payload,
 			HTTPRequest: toReq(linfo.req, v.Variables),
-		})
+		}
+
+		// If we don't set a resource the SDK will populate a global resource for us.
+		if v.MonitoredResourceType != "" {
+			e.Resource = &monitoredres.MonitoredResource{
+				Type:   v.MonitoredResourceType,
+				Labels: helper.ToStringMap(v.MonitoredResourceDimensions),
+			}
+		}
+		linfo.log(e)
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


adapter/stackdriver/stackdriver.go
```diff
@@ -17,7 +17,7 @@ package stackdriver
 import (
 	"context"
 
-	multierror "github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-multierror"
```
douglas-reid: small comment style-wise for your consideration: since the package path is `go-multierror` but it is actually the package `multierror`, using an alias to make that explicit in the client code can help improve readability. 

up to you.

testdata/config/stackdriver.yaml
```diff
@@ -55,34 +54,116 @@ spec:
...
+  monitoredResourceType: '"global"'
+  monitoredResourceDimensions:
+    # When using the global MR, this must be filled in, preferably matching the
+    # project_id given above in stackdriver config.
+    project_id: '"<your project ID>"'
+---
+apiVersion: "config.istio.io/v1alpha2"
+kind: metric
+metadata:
+  name: stackdriver-requestcount
```
douglas-reid: do we need / want a comment somewhere in our jumble of configs to indicate that we are trying to keep these definitions in sync with the non `stackdriver-` metrics?

also, i think we are avoiding `-`s entirely in names now.  @mandarjog is more knowledgeable on the details there, however.

adapter/stackdriver/stackdriver.go
ZackButcher: goimports removed it (or maybe gogland did), IDC either way.

testdata/config/stackdriver.yaml
ZackButcher: Added comment. Pinged mjog about naming.

adapter/stackdriver/stackdriver.go
ZackButcher: Heh, I added it back in my commit but it was removed when I formatted pre-commit. So definitely goimports removing it.

douglas-reid: interesting. i don't feel strongly at all. proceed.

adapter/stackdriver/log/log.go
```diff
@@ -131,13 +132,22 @@ func (h *handler) HandleLogEntry(_ context.Context, values []*logentry.Instance)
...
 			Payload:     payload,
 			HTTPRequest: toReq(linfo.req, v.Variables),
-		})
+		}
+
+		// If we don't set a resource the SDK will populate a global resource for us.
+		if v.MonitoredResourceType != "" {
+			e.Resource = &monitoredres.MonitoredResource{
+				Type:   v.MonitoredResourceType,
+				Labels: helper.ToStringMap(v.MonitoredResourceDimensions),
```
guptasu: I think if we make the template for MonitoredResourceDimenstions as map<string, string> you don't need the ToStringMap

ZackButcher: I intentionally didn't so that the operator could use any expressions they wanted, and we handle making them into strings since that's what the API requires. For example there's no IP address to string function in our expression language, so they could not use IP_ADDRESS expressions as labels, which is a pain.

guptasu: Got it.. so if it was map<string, string>, operator won't be able to pass IP_ADDRESS expression.

ZackButcher: Exactly.

ZackButcher:
PTAL

adapter/stackdriver/metric/metric_test.go
```diff
@@ -22,7 +22,7 @@ import (
 	"testing"
 	"time"
 
-	monitoring "cloud.google.com/go/monitoring/apiv3"
```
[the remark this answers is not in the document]
jmuk: It's working on my machine. I'm afraid that your goimports is a bit old.  Maybe good to run `go get -u golang.org/x/tools/cmd/goimports` ?
(by the way, should we do this through install_linters.sh or something?)

douglas-reid:
we can fix the naming thing later if we needed. let's get this in today.

ZackButcher:
Booo submit queue robot, merge it!

PR #1237, opened 2017-09-08
Grand rename and switch off legacy check/report

mandarjog:
1. Rename rule.selector to rule.match
   Accept "selector" in yaml with a  deprecation warning.
2. Rename target.* to destination.*
    Look for target.* if destination.* is not present. 
3. Remove legacy check and report calls. 
4. Remove chck/report legacy adapters. Keep K8s and memquota
5. Purge legacy config files of unnecessary sections.
**Release note**:
```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1237)

### DIFF
--- a/adapter/inventory.go
+++ b/adapter/inventory.go
@@ -15,32 +15,16 @@
 package adapter
 
 import (
-	"istio.io/mixer/adapter/denyChecker"
-	"istio.io/mixer/adapter/genericListChecker"
-	"istio.io/mixer/adapter/ipListChecker"
 	"istio.io/mixer/adapter/kubernetes"
 	"istio.io/mixer/adapter/memQuota"
 	"istio.io/mixer/adapter/noopLegacy"
-	"istio.io/mixer/adapter/prometheus"
-	"istio.io/mixer/adapter/redisquota"
-	"istio.io/mixer/adapter/serviceControl"
-	"istio.io/mixer/adapter/statsd"
-	"istio.io/mixer/adapter/stdioLogger"
 	"istio.io/mixer/pkg/adapter"
 )
 
 // InventoryLegacy returns the inventory of all available adapters.
 func InventoryLegacy() []adapter.RegisterFn {
 	return []adapter.RegisterFn{
-		denyChecker.Register,
-		genericListChecker.Register,
-		ipListChecker.Register,
 		memQuota.Register,
-		prometheus.Register,
-		redisquota.Register,
-		serviceControl.Register,
-		statsd.Register,
-		stdioLogger.Register,
 		kubernetes.Register,
 		noopLegacy.Register,
 	}
--- a/adapter/memquota2/dedup.go
+++ b/adapter/memquota2/dedup.go
@@ -105,7 +105,7 @@ func (du *dedupUtil) reapDedup() {
 	du.oldDedup = du.recentDedup
 	du.recentDedup = t
 
-	if du.logger.VerbosityLevel(4) {
+	if len(t) > 0 && du.logger.VerbosityLevel(4) {
 		du.logger.Infof("Running repear to reclaim %d old deduplication entries", len(t))
 	}
 
--- a/pkg/api/grpcServer.go
+++ b/pkg/api/grpcServer.go
@@ -98,7 +98,12 @@ func (c *compatBag) Get(name string) (v interface{}, found bool) {
 	if !strings.HasPrefix(name, "destination.") {
 		return
 	}
-	return c.parent.Get(strings.Replace(name, "destination.", "target.", 1))
+	compatAttr := strings.Replace(name, "destination.", "target.", 1)
+	v, found = c.parent.Get(compatAttr)
+	if found {
+		glog.Warningf("Deprecated attribute %s found", compatAttr)
+	}
+	return
 }
 
 func (c *compatBag) Names() []string {
--- a/pkg/api/grpcServer.go
+++ b/pkg/api/grpcServer.go
@@ -84,34 +84,27 @@ func NewGRPCServer(aspectDispatcher adapterManager.AspectDispatcher, dispatcher 
 	}
 }
 
+// compatBag implements compatibility between destination.* and target.* attributes.
 type compatBag struct {
 	parent attribute.Bag
 }
 
-// if a destination.* att is missing check the corresponding
-// target.*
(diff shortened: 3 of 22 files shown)


mandarjog:
yey!
https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_mixer/1237/mixer-e2e-smoketest/492/
```
I0908 09:33:36.265] Tests Summary
I0908 09:33:36.265] PASSED: //tests/e2e/tests/mixer:go_default_test 
I0908 09:33:36.266] PASSED: //tests/e2e/tests/bookinfo:go_default_test 
I0908 09:33:36.266] process 517 exited with code 0 after 15.4m
```

adapter/inventory.go
```diff
@@ -15,32 +15,16 @@
 package adapter
 
 import (
-	"istio.io/mixer/adapter/denyChecker"
-	"istio.io/mixer/adapter/genericListChecker"
-	"istio.io/mixer/adapter/ipListChecker"
 	"istio.io/mixer/adapter/kubernetes"
 	"istio.io/mixer/adapter/memQuota"
```
douglas-reid: why leave legacy `memQuota` ?

adapter/memquota2/dedup.go
```diff
@@ -105,7 +105,7 @@ func (du *dedupUtil) reapDedup() {
 	du.oldDedup = du.recentDedup
 	du.recentDedup = t
 
-	if du.logger.VerbosityLevel(4) {
+	if len(t) > 0 && du.logger.VerbosityLevel(4) {
```
douglas-reid: have we fixed this elsewhere yet?

pkg/api/grpcServer.go
```diff
@@ -83,6 +84,36 @@ func NewGRPCServer(aspectDispatcher adapterManager.AspectDispatcher, dispatcher
 	}
 }
 
+// compatBag implements compatibility between destination.* and target.* attributes.
+type compatBag struct {
+	parent attribute.Bag
+}
+
+// if a destination.* attribute is missing, check the corresponding target.* attribute.
```
douglas-reid: will this impact the "used attributes" calculation?  and do we care?

testdata/config/prometheus.yaml
```diff
@@ -41,7 +41,7 @@ metadata:
   name: prommetrics
   namespace: istio-config-default
 spec:
-  selector: "true"
+  #match: request.method == "/login"
```
douglas-reid: why add the commented out `request.method` bit here?  I almost feel that is adding more confusion than just dropping `match` altogether`.

testdata/configroot/scopes/global/subjects/global/rules.yml
```diff
@@ -11,70 +11,25 @@ rules:
       input_expressions:
         sourceUID: source.uid | ""
         sourceIP: source.ip | ip("0.0.0.0") # default to unspecified ip addr
-        targetUID: target.uid | ""
-        targetIP: target.ip | ip("0.0.0.0") # default to unspecified ip addr
+        targetUID: destination.uid | ""
```
douglas-reid: we should file an issue to update the k8s adapter and config proto

adapter/memquota2/dedup.go
ZackButcher: Looks like a bad merge somewhere along the way stomped the fix: https://github.com/istio/mixer/pull/1224 was merged but https://github.com/istio/mixer/pull/1228 undid it by mistake.

testdata/config/attributes.yaml
```diff
@@ -51,6 +51,8 @@ spec:
         valueType: STRING
       target.uid:
         valueType: STRING
+      destination.uid:
```
ZackButcher: Nit: this list was in alphabetical order before, can we maintain that? The connection attributes should be moved too.

testdata/config/quota.yaml
```diff
@@ -25,7 +25,7 @@ metadata:
   name: quota
   namespace: istio-config-default
 spec:
-#  selector: request.headers["user"] != "admin"
+#  match: source.user  != "admin"
```
ZackButcher: Same as above

pkg/api/grpcServer.go
mandarjog: It will not alter that calculation. It is checking for exactly one deprecated attribute in case the main one is not found.

adapter/inventory.go
mandarjog: legacy memQuota is enforcing quota at present.
That switch will happen as a bug later.

testdata/config/prometheus.yaml
mandarjog: dropped.

testdata/config/quota.yaml
mandarjog: I think it is useful to have an example in the checked in file. 
This says impose quota if not admin user.

pkg/config/store/store2.go
```diff
@@ -165,13 +165,36 @@ func (s *store2) Watch(ctx context.Context) (<-chan Event, error) {
...
 
+const (
+	ruleKind      = "rule"
+	selectorField = "selector"
+	matchField    = "match"
+)
+
+// warnDeprecationAndFix warns users about deprecated fields.
+// It maps the field into new name.
+func warnDeprecationAndFix(key Key, spec map[string]interface{}) map[string]interface{} {
```
guptasu: Why warn...  why not just error out, as bad field ? Else if people ignore warning, they we might end up with configs with selector and match going forward..

testdata/config/attributes.yaml
mandarjog: is it ok, not reorder it right now ?
We will do a reorder PR later :-)

pkg/config/store/store2.go
mandarjog: This code is why integration tests already pass.

testdata/config/attributes.yaml
ZackButcher: Yeah, not blocking.

mandarjog:
@douglas-reid @ZackButcher PTAL. I only really changed a comment so far.

ZackButcher:
I'll let Doug give the final LG.

PR #1238, opened 2017-09-08
restrict namespaces in CRD store through ?ns=...

jmuk:
I think it was implemented, but then the feature was dropped
at some point. This revives that implementation; filtering
configs in unrelated namespaces.

**Release note**:

```release-note
"ns" query parameter is available in the ConfigStore2URL; like k8s://?ns=mynamespace,istio-config-default
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1238)


jmuk:
/assign @mandarjog 
This is for #1182

PR #1243, opened 2017-09-08
Added authz template and opa adapter

mangchiandjjoe:
**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1243)


adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (b *builder) Build(context context.Context, env adapter.Env) (adapter.Handler, error) {
+	h := handler{
+		policy:      b.adapterConfig.Policy,
+		checkMethod: b.adapterConfig.CheckMethod,
+		env:         env,
+	}
+
+	parsed, err := ast.ParseModule("", string(h.policy))
+	if err != nil {
```
ZackButcher: It's fine to ignore this error and the compilation error below: you've already verified in Validate() that things work, it's not strictly necessary to check again here. We guarantee that `Validate` will always be called before `Build`. (It's also fine to leave the checks.)

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,183 @@
...
+	h.compiler = compiler
+	h.context = context
+
+	return h, nil
+}
+
+////////////////// Runtime Methods //////////////////////////
+
+func (h handler) HandleAuthz(context context.Context, instance *authz.Instance) (adapter.CheckResult, error) {
+	variables := make(map[string]interface{})
```
ZackButcher: You should probably assert in your `Validate` func that no key names for subjects, resources, or verbs overlap. Otherwise you'll lose data at runtime. The names for all of these are available in the `Types` handed to you `SetAuthzTypes`.

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,183 @@
...
+
+	rego := rego.New(
+		rego.Compiler(h.compiler),
+		rego.Query(h.checkMethod),
+		rego.Input(variables),
+	)
+
+	// Run evaluation.
+	rs, err := rego.Eval(h.context)
+
```
ZackButcher: nit: remove newline here.

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,183 @@
...
+	if len(rs) != 1 {
+		return adapter.CheckResult{
+			Status: rpc.Status{Code: int32(rpc.PERMISSION_DENIED),
+				Message: "authzOpa: request was rejected"},
+		}, nil
+	}
+
+	result, ok := rs[0].Expressions[0].Value.(bool)
+	if !ok {
```
ZackButcher: Change this to `if !ok || result == false { ... }` and combine the two `ifs`.

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,183 @@
...
+		rego.Query(h.checkMethod),
+		rego.Input(variables),
+	)
+
+	// Run evaluation.
+	rs, err := rego.Eval(h.context)
+
+	if err != nil {
+		return adapter.CheckResult{
+			Status: rpc.Status{Code: int32(rpc.PERMISSION_DENIED),
```
ZackButcher: Here and elsewhere you can use [pkg/status](https://github.com/istio/mixer/blob/master/pkg/status/status.go#L61) to reduce the boiler plate Status proto construction you're doing here.

adapter/authzOpa/authzOpa_test.go
```diff
@@ -0,0 +1,147 @@
...
+	info := GetInfo()
+
+	if !contains(info.SupportedTemplates, authz.TemplateName) {
+		t.Error("Didn't find all expected supported templates")
+	}
+
+	b := info.NewBuilder().(*builder)
+	b.SetAdapterConfig(cfg)
+	if err := b.Validate(); err != nil {
+		t.Errorf("Got error %v, expecting success", err)
```
ZackButcher: Here and in the error check below you should probably use `t.Fatal` since we can't proceed with the test given the broken state.

adapter/authzOpa/authzOpa_test.go
```diff
@@ -0,0 +1,147 @@
...
+	      input.verb = rule.verbs[_]
+	    }`
+	checkMethod = "data.mixerauthz.allow"
+	cfg         = &config.Params{
+		Policy:      policy,
+		CheckMethod: checkMethod,
+	}
+)
+
+func TestAccepted(t *testing.T) {
```
ZackButcher: Can you combine these two test methods into a single table driven test? You can keep the setup the same, make the input to the table the `user`, `verb`, and expected output `Code`.

ZackButcher:
Looks pretty good, just some minor comments.

template/authz/template.proto
```diff
@@ -0,0 +1,29 @@
...
+
+syntax = "proto3";
+
+package authz;
+
+import "mixer/v1/config/descriptor/value_type.proto";
+import "pkg/adapter/template/TemplateExtensions.proto";
+
+option (istio.mixer.v1.config.template.template_variety) = TEMPLATE_VARIETY_CHECK;
+option (istio.mixer.v1.config.template.template_name) = "Authz";
```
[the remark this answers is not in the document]
mangchiandjjoe: Removed

testdata/config/attributes.yaml
```diff
@@ -82,6 +82,10 @@ metadata:
   namespace: istio-config-default
 spec:
    attributes:
+      source.user:
```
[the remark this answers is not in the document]
mangchiandjjoe: Removed

testdata/config/attributes.yaml
```diff
@@ -82,6 +82,10 @@ metadata:
...
+        valueType: STRING
+      source.groups:
```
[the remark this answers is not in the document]
mangchiandjjoe: Authz doesn't use this attribute yet. Removed.

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Removed error handlers in the Build function

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Done

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Applied

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Applied

adapter/authzOpa/authzOpa_test.go
mangchiandjjoe: Merged test cases

adapter/authzOpa/authzOpa_test.go
mangchiandjjoe: Applied

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Where I can validate configurations for authz template? authzOpa adapter gets configuration it's own from the SetAdapterConfig method.

ZackButcher: The flow to initialize a new adapter, validate config is this:

1. `Set_Types` is called for each template you support (so in this case, just `SetAuthzTypes`). The types you're provided describe the shape (including the label types and names) of every object your adapter will ever see at runtime (when those types change we close the current instance and create a new one).
2. `SetAdapterConfig` is called
3. `Validate` is called; if it returns a non-nil error the config update is rejected and reported as bad.
4. `Build` is called; if if returns a non-nil error then we continue serving with the previous instance of the adapter (effectively rejecting the new config).

So if you just tuck the `authz.Type`s you get during `SetAuthzTypes` into a field on the builder like you do for your adapter config, you'll have both in hand during `Validate` to assert that no keys overlap.

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,164 @@
...
+	compiler.Compile(map[string]*ast.Module{"": parsed})
+	if compiler.Failed() {
+		ce = ce.Appendf(GetInfo().Name,
+			"Failed to compile the OPA policy: %v", compiler.Errors)
+	}
+
+	return
+}
+
+func (b *builder) Build(context context.Context, env adapter.Env) (adapter.Handler, error) {
```
ZackButcher: nit: style wise, I think this body would be more idiomatic:
```go
func (b *builder) Build(context context.Context, env adapter.Env) (adapter.Handler, error) {
    parsed, _ := ast.ParseModule("", string(h.policy))
    compiler := ast.NewCompiler()
    compiler.Compile(map[string]*ast.Module{"": parsed})
    return &handler{
      policy:      b.adapterConfig.Policy,
      checkMethod: b.adapterConfig.CheckMethod,
      env:         env,
      compiler:    compiler,
      context:     context,
    }, nil
}
```

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,164 @@
...
+type (
+	builder struct {
+		adapterConfig *config.Params
+	}
+
+	handler struct {
+		policy      string
+		checkMethod string
+
+		context  context.Context
```
ZackButcher: Sorry, missed this on the first pass: why do you hold on to the context here? I don't think you need it (you don't use it AFAIK).

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,164 @@
...
+		env:         env,
+	}
+
+	parsed, _ := ast.ParseModule("", string(h.policy))
+
+	compiler := ast.NewCompiler()
+	compiler.Compile(map[string]*ast.Module{"": parsed})
+
+	h.compiler = compiler
+	h.context = context
```
ZackButcher: I don't think you need to hold on to this context: to intent is for it to be used in any RPCs made during Build as part of setup, to enable cancelation of the build and that kind of thing (no existing adapter uses it for this today, but it's a brand new part of the interface).

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,164 @@
...
+	// Run evaluation.
+	rs, err := rego.Eval(h.context)
+	if err != nil {
+		return adapter.CheckResult{
+			Status: status.WithPermissionDenied(fmt.Sprintf("authzOpa: request was rejected: %v", err)),
+		}, nil
+	}
+
+	if len(rs) != 1 {
+		return adapter.CheckResult{
```
ZackButcher: Not necessary in this PR, but long term we need to think about cacheability here: the proxy uses this information to form cache keys. By supplying no caching information we force the proxy to call Mixer on every single request (massive degredation in system performance). You'll likely want the operator to provide caching information per OPA policy they're evaluating, and pass it back to the proxy here.

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,164 @@
...
+
+	h.compiler = compiler
+	h.context = context
+
+	return h, nil
+}
+
+////////////////// Runtime Methods //////////////////////////
+
+func (h handler) HandleAuthz(context context.Context, instance *authz.Instance) (adapter.CheckResult, error) {
```
ZackButcher: All of the funcs on handler should take a pointer receiver: right now you're forcing a copy of the entire handler state (i.e. compiled OPA VM) on every usage of the adapter. Unless it's intentional that you're forcing the copy (reading through the OPA code it doesn't appear that the compiler needs to be copied per use).

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Removed

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Applied

adapter/authzOpa/authzOpa.go
mangchiandjjoe: Removed

adapter/authzOpa/authzOpa.go
```diff
@@ -0,0 +1,164 @@
...
+	return nil
+}
+
+////////////////// Bootstrap //////////////////////////
+
+// GetBuilderInfo returns the BuilderInfo associated with
+// this adapter implementation.
+func GetInfo() adapter.Info {
+	return adapter.Info{
+		Name:        "authzOpa",
```
ZackButcher: Name must be all lowercase, alphanumeric, starting with a letter. That's why the presubmit is failing for you.

In general though I think we should just call this adapter "opa", or even "openpolicyagent"; we don't typically include the template name in the adapter name. @mandarjog @douglas-reid thoughts w.r.t. naming?

mangchiandjjoe: Changed name to opa

adapter/authzOpa/authzOpa.go
mangchiandjjoe: I will create another issue for the cacheability

adapter/authzOpa/authzOpa.go
ZackButcher: Talked out of band - bad reading on my part, the entire compiler is not being copied, just a pointer to it. Either way, there's no need for a copy at all, we can just use a pointer to the handler instance.

ZackButcher:
I haven't had the chance to review newer changes here, someone else should while I'm in a weird timezone. I think all the logic is fine, I haven't had a chance to look at the tests.

template/authz/template.proto
```diff
@@ -0,0 +1,27 @@
...
+syntax = "proto3";
+
+package authz;
+
+import "mixer/v1/config/descriptor/value_type.proto";
+import "mixer/v1/template/extensions.proto";
+
+option (istio.mixer.v1.template.template_variety) = TEMPLATE_VARIETY_CHECK;
+
+message Template {
```
geeknoid: Need comments on the struct and its fields.

template/authz/template.proto
```diff
@@ -0,0 +1,27 @@
...
+  map<string, istio.mixer.v1.config.descriptor.ValueType> subject = 1;
+  map<string, istio.mixer.v1.config.descriptor.ValueType> action = 2;
```
geeknoid: Does action really need to be this generic? Wouldn't a single string be sufficient here?

mangchiandjjoe:
I will split the PR

template/authz/template.proto
mangchiandjjoe: According to the Istio Native RBAC Design V2 document, action has multiple fields, namespace, service, method, path and attributes.

template/authz/template.proto
mangchiandjjoe: Added in the new PR

mangchiandjjoe:
Closing the PR to split. Separated PRs for the authz template and opa adapter will be created.

mangchiandjjoe:
PR for authorization was created
https://github.com/istio/mixer/pull/1458

PR #1248, opened 2017-09-08
Update linters, pre-commit, and pre-push hooks

ZackButcher:
Update:
- linter script to always skip //bin and //testdata
- the linter script's invocation in the pre-push script to set the branch to diff against to upstream/master, vastly reducing the amount linted from the entire repo to only the packages your PR touched relative to upstream/master.
- always run ./bin/bazel_to_go.py before running ./bin/fmt.sh to avoid fmt changing a ton of stuff due to the golang representation of our repo not matching the latest state.

I was able to actually use `git push` to send this PR rather than the `git push --no-verify` I've been doing for weeks now. Big improvement.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1248)

### DIFF
--- a/bin/linters.sh
+++ b/bin/linters.sh
@@ -25,6 +25,7 @@ prep_linters() {
 
 go_metalinter() {
     local parent_branch="${PARENT_BRANCH}"
+    echo parent_branch = "${parent_branch}"
     if [[ ! -z ${TRAVIS_PULL_REQUEST} ]];then
         # if travis pull request only lint changed code.
         if [[ ${TRAVIS_PULL_REQUEST} != "false" ]]; then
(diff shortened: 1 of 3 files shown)


ZackButcher:
Note that this assumes that in your git repo you've named the remote pointing at `github.com/istio/mixer` exactly "upstream", which is what our developer guide recommends.

bin/linters.sh
```diff
@@ -37,7 +38,7 @@ go_metalinter() {
 
     if [[ -z ${LAST_GOOD_GITSHA} ]] && [[ -n "${parent_branch}" ]]; then
         LAST_GOOD_GITSHA="$(git log ${parent_branch}.. --pretty="%H"|tail -1)"
-        [[ ! -z ${LAST_GOOD_GITSHA} ]] && LAST_GOOD_GITSHA="${LAST_GOOD_GITSHA}^"
+        [[ ! -z ${LAST_GOOD_GITSHA} ]] && LAST_GOOD_GITSHA="${LAST_GOOD_GITSHA}"
```
jmuk: I'm not sure what this line means now. This line does not change `LAST_GOOD_GITSHA` anyways. Just removing this line would be better.

ZackButcher: Done

ZackButcher:
Added the line back since I don't know why the caret was added in the first place.

ZackButcher:
Weird.

/retest

PR #1250, opened 2017-09-09
Warn and accept deprecated selector

mandarjog:
**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1250)

### DIFF
--- a/pkg/config/store/convert.go
+++ b/pkg/config/store/convert.go
@@ -24,6 +24,28 @@ import (
 	"github.com/golang/glog"
 )
 
+const (
+	ruleKind      = "rule"
+	selectorField = "selector"
+	matchField    = "match"
+)
+
+// warnDeprecationAndFix warns users about deprecated fields.
+// It maps the field into new name.
+func warnDeprecationAndFix(key Key, spec map[string]interface{}) map[string]interface{} {
+	if key.Kind != ruleKind {
+		return spec
+	}
+	sel := spec[selectorField]
+	if sel == nil {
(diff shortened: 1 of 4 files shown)


pkg/config/store/convert.go
```diff
@@ -24,6 +24,28 @@ import (
...
+func warnDeprecationAndFix(key Key, spec map[string]interface{}) map[string]interface{} {
+	if key.Kind != ruleKind {
+		return spec
+	}
+	sel := spec[selectorField]
+	if sel == nil {
+		return spec
+	}
+	glog.Warningf("Deprecated field 'selector' used in %s. Use 'match' instead.", key)
+	spec[matchField] = sel
```
jmuk: should we check if `matchField` already exists in the spec?  It's very unlikely to happen, but it overwrites the match field if both exist.

pkg/config/store/convert.go
```diff
@@ -24,6 +24,28 @@ import (
...
 
+const (
+	ruleKind      = "rule"
+	selectorField = "selector"
+	matchField    = "match"
+)
+
+// warnDeprecationAndFix warns users about deprecated fields.
+// It maps the field into new name.
+func warnDeprecationAndFix(key Key, spec map[string]interface{}) map[string]interface{} {
```
jmuk: It would be great to add test cases of this function itself.

jmuk:
/lgtm
(i'm not an approver)

PR #1256, opened 2017-09-10
Add overrides to memquota2

mandarjog:
This PR adds override support to memquota2 as `back end configuration` with the intention of migrating then to individual config resources in 0.3

It gains feature parity with the old memquota setup + old config model. 
This lets Mixer switch off old quota dispatch after creating replacement test configurations.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1256)

### DIFF
--- a/adapter/memQuota/util/util.go
+++ b/adapter/memQuota/util/util.go
@@ -120,7 +120,7 @@ func (qu *QuotaUtil) ReapDedup() {
 	qu.OldDedup = qu.RecentDedup
 	qu.RecentDedup = t
 
-	if qu.Logger.VerbosityLevel(4) {
+	if len(t) > 0 && qu.Logger.VerbosityLevel(4) {
 		qu.Logger.Infof("Running repear to reclaim %d old deduplication entries", len(t))
 	}
 
--- a/adapter/memquota2/config/config.proto
+++ b/adapter/memquota2/config/config.proto
@@ -26,8 +26,28 @@ option (gogoproto.gostring_all) = false;
 
 message Params {
 	message Quota {
+		option (gogoproto.goproto_getters) = true;
 		// The name of the quota
 		string name = 1;
+
+		// The upper limit for this quota.
+		int64 max_amount = 2;
+
+		// The amount of time allocated quota remains valid before it is
+		// automatically released. This is only meaningful for rate limit
+		// quotas, otherwise the value must be zero.
+		google.protobuf.Duration valid_duration = 3 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+
+		// Overrides associated with this quota.
+		// The first matching override is applied.
+		repeated Override overrides = 4 [(gogoproto.nullable) = false];
+	}
+	message Override {
+		option (gogoproto.goproto_getters) = true;
+
+		// The specific dimensions for which this override applies.
+		// String representation of instance dimensions is used to check against configured dimensions.
+		map <string, string> dimensions = 1;
 
 		// The upper limit for this quota.
 		int64 max_amount = 2;
--- a/adapter/memquota2/memquota.go
+++ b/adapter/memquota2/memquota.go
@@ -49,6 +49,9 @@ type handler struct {
 
 	// the limits we know about
 	limits map[string]*config.Params_Quota
+
+	// logger provided by the framework
+	logger adapter.Logger
 }
 
 // Limit is implemented by Quota and Override messages.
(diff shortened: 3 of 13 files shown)


mandarjog:
Added istio/istio [PR](https://github.com/istio/istio/pull/725)

adapter/memquota2/config/config.proto
```diff
@@ -26,6 +26,7 @@ option (gogoproto.gostring_all) = false;
 
 message Params {
 	message Quota {
+		option (gogoproto.goproto_getters) = true;
```
mandarjog: Need getters so we can abstract over `Quota` and `Override` message.
They both implement 
```
interface {
 GetValidDuration()
  GetMaxAmount()
}
```

adapter/memQuota/util/util.go
```diff
@@ -120,7 +120,7 @@ func (qu *QuotaUtil) ReapDedup() {
 	qu.OldDedup = qu.RecentDedup
 	qu.RecentDedup = t
 
-	if qu.Logger.VerbosityLevel(4) {
+	if len(t) > 0 && qu.Logger.VerbosityLevel(4) {
 		qu.Logger.Infof("Running repear to reclaim %d old deduplication entries", len(t))
```
douglas-reid: let's also fix: "repear" -> "reaper"

adapter/memquota2/memquota.go
```diff
@@ -48,11 +48,53 @@ type handler struct {
...
+// Limit is implemented by Quota and Override messages.
+type Limit interface {
+	GetMaxAmount() int64
+	GetValidDuration() time.Duration
+}
+
+// matchDimensions matches configured dimensions with dimensions of the instance.
+func matchDimensions(cfg map[string]string, inst map[string]interface{}) bool {
+	for k, val := range cfg {
+		rval := inst[k]
```
douglas-reid: so, this requires the dimensions to be in the same order as well, right?

adapter/memquota2/memquota.go
```diff
@@ -48,11 +48,53 @@ type handler struct {
...
+	}
+	return true
+}
+
+// limit returns the limit associated with this particular request.
+// Check if the instance matches an override, else return the default limit.
+func limit(cfg *config.Params_Quota, instance *quota.Instance) Limit {
+	for idx := range cfg.Overrides {
+		o := cfg.Overrides[idx]
+		if matchDimensions(o.Dimensions, instance.Dimensions) {
```
douglas-reid: don't the `MakeKey()` methods provide a set of functionality we could use instead of `matchDimensions`?  Is there a reason not to use the existing Key funcs?

pkg/api/grpcServer.go
```diff
@@ -218,6 +218,17 @@ func (s *grpcServer) Check(legacyCtx legacyContext.Context, req *mixerpb.CheckRe
 					GrantedAmount: qma.Amount,
 				}
 			}
+
+			if glog.V(2) {
```
douglas-reid: do we want this section of code?  Doesn't this imply dimensions that might not be on the quota descriptor?  For instance, I might not choose to dimension quota with `source.name`.

pkg/runtime/dispatcher.go
```diff
@@ -236,15 +236,21 @@ func (m *dispatcher) Quota(ctx context.Context, requestBag attribute.Bag,
 	qres, err := m.dispatch(ctx, requestBag, adptTmpl.TEMPLATE_VARIETY_QUOTA,
 		func(call *Action) []dispatchFn {
 			for _, inst := range call.instanceConfig {
-				if inst.Name != qma.Quota {
-					continue
-				}
+				// if inst.Name != qma.Quota {
```
douglas-reid: what is the harm in leaving this check enabled?

testdata/config/quota.yaml
```diff
@@ -17,7 +23,7 @@ metadata:
 spec:
   dimensions:
     source: source.labels["app"] | "unknown"
```
douglas-reid: with the raw vm work, i'm wondering if we should switch from labels to `source.service` (at least for 0.2).

pkg/runtime/dispatcher.go
mandarjog: It will not work. The qma.Name == `RequestCount`,  however instance names in config are fully qualified. So mixer client **will** have to send the specific fully qualified name over.

adapter/memquota2/memquota.go
mandarjog: `MakeKey()` creates a unique key using __all__ dimensions. 
`matchDimensions` needs to match on partial dimensions.

adapter/memQuota/util/util.go
ZackButcher: How do we keep ending up with this change being removed? This is at least the third time.

pkg/api/grpcServer.go
mandarjog: regardless of quota dimensioning, I want accessLog to convey just some reasonable notion of source and target, how much quota was requested and how much was granted.

adapter/memquota2/memquota.go
ZackButcher: This is a map lookup - it shouldn't impose ordering constraints, right?

testdata/config/quota.yaml
mandarjog: how about `source.labels["app"] | source.service`

pkg/api/grpcServer.go
ZackButcher: +1, I don't like this chunk. _If_ we want this kind of logging, just log "allocated %d tokens along dimensions: %v". Don't hard code attributes anywhere.

adapter/memquota2/memquota.go
mandarjog: Not really. we are checking *all* dimensions specified in adapterConfig. They can appear in any order.

adapter/memQuota/util/util.go
mandarjog: This PR can now remove the entire adapter, but I will send a separate PR for that.

testdata/config/quota.yaml
douglas-reid: how about `source.labels["app"] | source.service | "unknown"` to cover all bases?

pkg/api/grpcServer.go
mandarjog: I just log `destination.service` along with quota response now.

I also log full dimensions at log level 2.

testdata/config/quota.yaml
mandarjog: done.

adapter/memquota2/memquota.go
douglas-reid: ah, ok. i was looking too quickly while trying to participate in that meeting.

PR #1265, opened 2017-09-11
fix extra newline causing loop of warning

ldemailly:
```release-note
NONE
```

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1265)


douglas-reid:
is this the fix we want?  or do we want to fix the logic that is confused by newlines?

PR #1270, opened 2017-09-12
Assume k8s:// inCluster kubernetes if  configStore2URL is empty

mandarjog:
**Release note**:

```release-note
configStore2URL is the URL to connect to the config store. Use k8s://path_to_kubeconfig or fs:// for file system. If path_to_kubeconfig is empty or the parameter is unspecified, in-cluster kubeconfig is used.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1270)

### DIFF
--- a/cmd/server/cmd/server.go
+++ b/cmd/server/cmd/server.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
 	_ "expvar" // For /debug/vars registration. Note: temporary, NOT for general use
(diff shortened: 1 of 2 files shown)


cmd/server/cmd/server.go
```diff
@@ -89,6 +90,33 @@ type serverArgs struct {
 	globalConfigFile string
 }
 
+func (sa *serverArgs) String() string {
+	var b bytes.Buffer
```
guptasu: Can we just reflect on the fields and do : b.WriteString(fmt.Sprint("'%s': '%v'\n", field.Name, field.Value)) ? There is a change this list might get out of sync.. (code review is the only way to prevent that from happening. Since the code review differ only shows the changed lines and it's immediate surrounding, chances are this will get missed in PR review).

Or just use [spew.dump]("github.com/davecgh/go-spew/spew"
) or %v on serverArgs.

mandarjog: Thats what I tried first, but all fields are private at present, so , they will all have to be turned into public fields.

guptasu: Consider using spew.dump. I think that works on private fields too. We can just delete tall this code.

PR #1276, opened 2017-09-12
fix test flakiness in queue_test.go

### LINKED ISSUE #1273: queue_test.go may be flaky
I've heard that sometimes pkg/config/store/queue_test.go can fail with the following log:

```
==================== Test output for //pkg/config/store:go_default_test:
(snip)
--- FAIL: TestQueueCancelSync (0.00s)
    queue_test.go:149: Got name  Want 8
    queue_test.go:149: Got name  Want 9
(snip)
FAIL
```

This is unexpected. Will investigate why.

jmuk:
fixes #1273

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1276)

### DIFF
--- a/pkg/config/store/queue_test.go
+++ b/pkg/config/store/queue_test.go
@@ -17,6 +17,7 @@ package store
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
@@ -139,19 +140,27 @@ func TestQueueCancelSync(t *testing.T) {
 			Value: &BackEndResource{},
 		}
 	}
+	// Wait for the queue's run loop to process chin.
+	time.Sleep(time.Millisecond)
 	cancel()
 	// Wait for the queue's run loop to end.
 	time.Sleep(time.Millisecond)
 	// Read the bufferred events.
+	emptyEv := Event{}
 	for i := 0; i < choutBufSize; i++ {
 		ev := <-q.chout
+		if reflect.DeepEqual(ev, emptyEv) {
+			// ev can be empty; this happens when run() goroutine doesn't run enough
+			// before cancel() is called.
+			break
+		}
 		if ev.Name != fmt.Sprintf("%d", i) {
 			t.Errorf("Got name %s Want %d", ev.Name, i)
 		}
 	}
 	// After the buffer runs out, it should return an empty since it's closed due to cancel.
 	ev := <-q.chout
-	if ev.Name != "" {
-		t.Errorf("Got %+v, Want empty", ev)
+	if !reflect.DeepEqual(ev, emptyEv) {
+		t.Errorf("Got %+v, Want empty %+v", ev, emptyEv)
 	}
 }


pkg/config/store/queue_test.go
```diff
@@ -139,19 +140,27 @@ func TestQueueCancelSync(t *testing.T) {
 			Value: &BackEndResource{},
 		}
 	}
+	// Wait for the queue's run loop to process chin.
+	time.Sleep(time.Millisecond)
```
guptasu: how do we know wait for 1 millisecond enough ?

pkg/config/store/queue_test.go
```diff
@@ -139,19 +140,27 @@ func TestQueueCancelSync(t *testing.T) {
 			Value: &BackEndResource{},
 		}
 	}
+	// Wait for the queue's run loop to process chin.
```
guptasu: Is `chin` a thing ? Sorry, I am not clear what it means.

pkg/config/store/queue_test.go
guptasu: Anything else that we can do to not sleep but rather be more event driven ?

pkg/config/store/queue_test.go
```diff
@@ -139,19 +140,27 @@ func TestQueueCancelSync(t *testing.T) {
...
 	// Wait for the queue's run loop to end.
 	time.Sleep(time.Millisecond)
 	// Read the bufferred events.
+	emptyEv := Event{}
 	for i := 0; i < choutBufSize; i++ {
 		ev := <-q.chout
+		if reflect.DeepEqual(ev, emptyEv) {
+			// ev can be empty; this happens when run() goroutine doesn't run enough
+			// before cancel() is called.
+			break
```
guptasu: If we break here, does it mean we didn't test the thing at all ?

pkg/config/store/queue_test.go
jmuk: Yes it's the input channel created in line 134.

pkg/config/store/queue_test.go
jmuk: There's no guarantee 1 msec is sufficient. There's no hook here.  That's why I added the break at line 152-156 below.
I think however, theoretically it's nicer to wait some short time before `cancel()` is called.

pkg/config/store/queue_test.go
jmuk: It should break here because `q.chout` is considered to be closed already. This is the core part to address the error you've met. Here's the thing; `q.chout` is a channel with some size of buffer, and `q` maintains a goroutine to push events into `chout`. This test pushes sufficient events, and then cancel the context, and check the content in the `chout`.
The existing code assumes `chout`'s buffer is filled with the created events, but sometimes goroutine can be stopped before putting the events fully. In that case it should break the loop.

geeknoid:
Is this PR still useful?

If so, could you rebase and see it through?
If not, let's delete the PR.

Thanks.

jmuk:
deleting, I haven't heard similar reports recently, assuming obsolete

PR #1280, opened 2017-09-12
fix perftests

jmuk:
- pkg/attributes and pkg/adapterManager do not have benchmarks anymore
- pkg/api and pkg/il/interpreter are missing in preftest
- fix pkg/api/perf_test.go's panic by adding the new dispatcher
- fix interpreterBenchmark_test.go by fixing the IL opcodes

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1280)

### DIFF
--- a/bin/perftest.sh
+++ b/bin/perftest.sh
@@ -8,7 +8,7 @@ cd $ROOT
 
 
 echo "Perf test"
-DIRS="pkg/expr pkg/attribute pkg/adapterManager"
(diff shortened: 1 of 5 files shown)


mandarjog:
Thanks for fixing this.
We need to write a perf test for `pkg/runtime` as well, which is equivalent to the prior `pkg/adapterManager` perf test.

/lgtm

PR #1283, opened 2017-09-12
rename deprecated files property in bazel

theganyo:
**Release note**: rename deprecated 'files' property in bazel to 'srcs'

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1283)

### DIFF
--- a/docker/BUILD
+++ b/docker/BUILD
@@ -6,10 +6,10 @@ load(":cacerts.bzl", "cacerts")
 
 pkg_tar(
     name = "mixer_tar",
-    extension = "tar.gz",
-    files = [
+    srcs = [
         "//cmd/server:mixs",
     ],
+    extension = "tar.gz",
     mode = "0755",
     package_dir = "/usr/local/bin/",
     tags = ["manual"],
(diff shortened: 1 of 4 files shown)


theganyo:
Hmm. The CI tests are saying `no such attribute 'srcs' in 'pkg_tar' rule`. I assume because the CI is using an old version of bazel...

ZackButcher:
Actually, in this case it looks like the warning was a lie: all of the [bazel_docker repo](https://github.com/bazelbuild/rules_docker) uses `files`, not `srcs`. Likely what happened is that Bazel deprecated `files` upstream and bazel_docker hasn't updated yet.

ZackButcher:
Yeah, that's a bummer. Thank you for spending the time to put together a PR though! We can either leave this PR up and we can merge it in with an updated SHA for bazel_docker after they fix things upstream, or we can close this.

theganyo:
Call it as you see it. Either way, I suppose an issue should be raised with the bazelbuild folks?

ZackButcher:
Closing now and we'll create a new PR to update things.

Issue #1282, opened 2017-09-12
IL Runtime fatal error: concurrent map read and map write


mandarjog:
IL runtime throws a fatal error during testing.
`Istio Mixer: version: unknown (build: 2017-09-11-543c458, status: Clean)`

```
[19:56:49] fatal error: concurrent map read and map write
[19:56:49] 
[19:56:49] goroutine 75 [running]:
[19:56:49] runtime.throw(0x18c88aa, 0x21)
[19:56:49]      /home/bootstrap/.cache/bazel/_bazel_bootstrap/2ba7a783f3c05b4605aea2aadf54c5cd/external/go1_8_3_linux_amd64/src/runtime/panic.go:596 +0x95 fp=0xc420d0fec8 sp=0xc420d0fea8
[19:56:49] runtime.mapaccess2_faststr(0x16a80c0, 0xc420e86cf0, 0xc4203a1d63, 0x7, 0xc420d8e628, 0x1)
[19:56:49]      /home/bootstrap/.cache/bazel/_bazel_bootstrap/2ba7a783f3c05b4605aea2aadf54c5cd/external/go1_8_3_linux_amd64/src/runtime/hashmap_fast.go:326 +0x50a fp=0xc420d0ff28 sp=0xc420d0fec8
[19:56:49] istio.io/mixer/pkg/il/interpreter.(*Interpreter).run(0xc420e87350, 0xc420e87320, 0x2198e40, 0xc4204588c0, 0xc420af2c00, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[19:56:49]      pkg/il/interpreter/interpreterRun.go:909 +0x4c0e fp=0xc420d11158 sp=0xc420d0ff28
[19:56:49] istio.io/mixer/pkg/il/interpreter.(*Interpreter).Eval(0xc420e87350, 0x189e866, 0x4, 0x2198e40, 0xc4204588c0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[19:56:49]      pkg/il/interpreter/interpreter.go:70 +0xf7 fp=0xc420d11228 sp=0xc420d11158
[19:56:49] istio.io/mixer/pkg/il/evaluator.(*IL).evalResult(0xc4203d0d50, 0xc420194fe0, 0x20, 0x2198e40, 0xc4204588c0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[...]

PR #1284, opened 2017-09-13
Fix IL panic: concurrent read/write from map

mandarjog:
1. Adds RWLocks around the map that was being concurrently accessed @ozevren 
2. Add a test case to induce this failure.  @mandarjog

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1284)


mandarjog:
Fixes #1282

PR #1290, opened 2017-09-13
Move code coverage to a different prow job

### LINKED ISSUE #1289: Code Coverage is taking 13 minutes out of presubmit
We should be running the code coverage outside of the presubmit job in parallel with the smoke test.

sebastienvas:
Code coverage is taking 13 minutes. Moving it to an after success job. It will run in // with the smoke test.

Need to go before https://github.com/istio/test-infra/pull/496
Fixes #1289

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1290)

### DIFF
--- a/prow/mixer-coverage.sh
+++ b/prow/mixer-coverage.sh
@@ -42,13 +42,8 @@ fi
 echo "=== Bazel Build ==="
 bazel build //...
 
-echo "=== Bazel Tests ==="
-bazel test //...
-
-
 echo "=== go build ./... ==="
 bin/bazel_to_go.py
-go build ./...
 
 echo "=== Code Coverage ==="
 ./bin/codecov.sh | tee codecov.report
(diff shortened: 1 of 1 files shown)


prow/mixer-coverage.sh
```diff
@@ -0,0 +1,62 @@
...
+else
+    # Use the current commit.
+    GIT_SHA="$(git rev-parse --verify HEAD)"
+fi
+
+echo "=== Bazel Build ==="
+bazel build //...
+
+echo "=== Bazel Tests ==="
+bazel test //...
```
mandarjog: coverage does not need to run tests.
It only needs to compile code and and run `bazel_to_go.py`

sebastienvas: Done

PR #1306, opened 2017-09-15
Add example configuration for listchecker 

mandarjog:
1. Add whitelist config example
2. Remove default value for `ProviderUrl`, there is no reasonable default.

This example is used in docs.
**Release note**:
```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1306)

### DIFF
--- a/testdata/config/listcheck.yaml
+++ b/testdata/config/listcheck.yaml
@@ -1,0 +1,30 @@
+apiVersion: "config.istio.io/v1alpha2"
+kind: listchecker
+metadata:
+  name: staticversion
+  namespace: istio-config-default
+spec:
+  # providerUrl: ordinarily black and white lists are maintained
+  # externally and fetched asynchronously using the providerUrl.
+  overrides: ["v1", "v2"]  # overrides provide a static list
+  blacklist: false
+---
+apiVersion: "config.istio.io/v1alpha2"
+kind: listentry
+metadata:
+  name: appversion
+  namespace: istio-config-default
+spec:
+  value: source.labels["version"]
+
+---
+apiVersion: "config.istio.io/v1alpha2"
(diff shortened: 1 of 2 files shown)


testdata/config/listcheck.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: "config.istio.io/v1alpha2"
+kind: listchecker
+metadata:
+  name: staticversion
+  namespace: istio-config-default
+spec:
+  # providerUrl: ordinarily black and white lists are maintained
+  # externally and fetched asynchronously using the providerUrl.
+  overrides: ["v1", "v2"]  # overrides provide a static list
+  blacklist: false
```
douglas-reid: including this just to have the example + docs ?

testdata/config/listcheck.yaml
```diff
@@ -0,0 +1,30 @@
...
+apiVersion: "config.istio.io/v1alpha2"
+kind: rule
+metadata:
+  name: checkwl
+  namespace: istio-config-default
+spec:
+  match: destination.labels["app"] == "ratings"
+  actions:
+  - handler: staticversion.listchecker
+    instances: [ appversion.listentry ]
```
douglas-reid: is this style our style recommendation for instance lists?

testdata/config/listcheck.yaml
mandarjog: Yes, this is used in the docs.

testdata/config/listcheck.yaml
mandarjog: I have been using it, but I think we should switch to the standard way.
It is not as concise, but makes diffs easier.
I will send a PR

PR #1325, opened 2017-09-19
Enable :9093/config endpoint to get status of runtime config resources

mandarjog:
This PR adds `/config/` endpoint to Mixer to query mixer current runtime state for config resources.
It addresses lack of problem determination tools for mixer. If a rule or handler is not taking affect, an operator can use this endpoint to query mixer.

example when an error had occurred.
```
mjog-macbookpro:~ mjog$ curl http://localhost:9093/config/istio-config-default/rule/promtcp  | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   686  100   686    0     0   112k      0 --:--:-- --:--:-- --:--:--  133k
{
  "metadata": {
    "name": "promtcp",
    "namespace": "istio-config-default",
    "labels": {
      "istio-protocol": "tcp"
    },
    "revision": ""
  },
  "spec": {
    "actions": [
      {
        "handler": "handler.prometheus.istio-config-default",
        "instances": [
          "tcpbytesent.metric.istio-config-default",
          "tcpbytereceived.metric.istio-config-default"
        ]
      }
    ]
  },
  "status": {
    "code": "error",
    "errors": [
      "Filtering action from rule istio-config-default/promtcp. Handler handler.prometheus.istio-config-default could not be initialized due to cannot infer type information from params in instance 'requestsize.metric.istio-config-default': OR($request.size, \"abcd\") arg 2 (\"abcd\") typeError got STRING, expected INT64.",
      "Purging rule promtcp with no actions"
    ]
  }
}

```

```release-note
A config endpoint is available on the mixer self monitoring port. 
curl mixer:9093/config/$full_name_of_config_resource

If resource is not present, returns 404
If resource is present, resource.status field contains status.
If status.code == "error", status.errors contains a list of textual errors.

status codes: [ "active", "pendingUpdate", "pendingDelete", "error" ]
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1325)

### DIFF
--- a/cmd/server/cmd/server.go
+++ b/cmd/server/cmd/server.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	_ "expvar" // For /debug/vars registration. Note: temporary, NOT for general use
 	"fmt"
 	"io/ioutil"
@@ -59,6 +60,7 @@ import (
 const (
 	metricsPath = "/metrics"
 	versionPath = "/version"
+	configPath  = "/config/"
 )
 
 type serverArgs struct {
@@ -122,6 +124,11 @@ type ServerContext struct {
 	GP        *pool.GoroutinePool
 	AdapterGP *pool.GoroutinePool
 	Server    *grpc.Server
+}
+
+// ConfigGetter gives access to Mixer config snapshot.
+type ConfigGetter interface {
+	ConfigGet(path string) *store.Resource
 }
 
 func serverCmd(info map[string]template.Info, adapters []adptr.InfoFn, printf, fatalf shared.FormatFn) *cobra.Command {
@@ -272,6 +279,7 @@ func setupServer(sa *serverArgs, info map[string]template.Info, adapters []adptr
 	}
 
 	var dispatcher mixerRuntime.Dispatcher
+	var configGetter ConfigGetter
 
 	if sa.configStore2URL == "" {
 		printf("configStore2URL is not specified, assuming inCluster Kubernetes")
@@ -283,7 +291,7 @@ func setupServer(sa *serverArgs, info map[string]template.Info, adapters []adptr
 	if err != nil {
 		fatalf("Failed to connect to the configuration server. %v", err)
 	}
-	dispatcher, err = mixerRuntime.New(eval, gp, adapterGP,
+	dispatcher, configGetter, err = mixerRuntime.New(eval, gp, adapterGP,
 		sa.configIdentityAttribute, sa.configDefaultNamespace,
 		store2, adapterMap, info,
 	)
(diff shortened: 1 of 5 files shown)


cmd/server/cmd/server.go
```diff
@@ -408,17 +416,39 @@ func setupServer(sa *serverArgs, info map[string]template.Info, adapters []adptr
 		fatalf("Unable to listen on socket: %v", err)
 	}
 
+	sm := http.NewServeMux()
```
jmuk: Please keep existing `http.Handle` things and not defining servemux; actually this implicitly enables net/http/pprof (https://golang.org/pkg/net/http/pprof/).

rshriram:
Is this the last error for this config? or is this a validation error when user submits the config ? I guess the former, since we handle validation errors in line.
If this is the last error for this config, is it just one last error or N such errors?

cmd/server/cmd/server.go
```diff
@@ -408,17 +416,39 @@ func setupServer(sa *serverArgs, info map[string]template.Info, adapters []adptr
...
+
+		var ba []byte
+		var verr error
+
+		if ba, verr = json.Marshal(cfg); err != nil {
+			printf("error converting %v: %v", cfg, verr)
+			out.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		out.Header().Add("content-type", "application/json")
```
jmuk: can we also allow yaml as for output?  I feel like it's more visible and debuggable for human readers.

Maybe it can be dispatched `Accept` request header.

pkg/config/store/store2.go
```diff
@@ -62,8 +62,31 @@ type BackEndResource struct {
 
 // Resource represents a resources with converted spec.
 type Resource struct {
-	Metadata ResourceMeta
-	Spec     proto.Message
+	Metadata ResourceMeta  `json:"metadata"`
+	Spec     proto.Message `json:"spec"`
+	Status   Status        `json:"status"`
```
jmuk: Does this need to belong to Resource?  As far as I see, the status is handled separately as `statusMap` in `pkg/runtime`, so `Resource` doesn't have to hold the status as a field?

kyessenov:
I think you need to add more structure to Status field. Maybe make it a state machine with READY state, and then stuff errors into events instead?

cmd/server/cmd/server.go
mandarjog: One can pipe the json output thru `jq` to make it more readable.

jmuk: It's okay to be JSON by default. I'm saying to allow yaml when `Accept: application/yaml` header is specified.

mandarjog:
@rshriram This is independent of the validation path via webhooks on ingest. 
Mixer watches configuration and validates a snapshot of the config at a time. It is ok for the snapshot to be inconsistent. Errors are recorded during validation and associated with the config resources. 

So those errors are *all* the errors associated with a particular resource.

cmd/server/cmd/server.go
mandarjog: Ok fixed it.

pkg/config/store/store2.go
mandarjog: This is somewhat consistent with k8s, where `spec`, `metadata` and `status` are all at the same level, and status is considered a sub object of the resource.

mandarjog:
@kyessenov It does make sense to have conditions/ events etc. However we can evolve the status schema in next several weeks. This gets in something that is already useful.

kyessenov:
Should we wait until it's ready and we promote Status to an API object then?

douglas-reid:
I'm somewhat concerned that this is a last second addon. It'd be great if we had a full design and API review... and maybe Istio-wide consistency (Pilot).

Do we think this is OK as a hidden feature (that we don't advertise for the time being)?

Should we at least consider moving it to something like `/config/v1alpha1/resources` or something?  Or should it maybe live off the `/debug` subtree?

pkg/config/store/store2.go
```diff
@@ -62,8 +62,31 @@ type BackEndResource struct {
...
+	// PendingDelete - the resource is pending deletion.
+	PendingDelete Code = "pendingDelete"
+	// InError - the resource is in error.
+	InError Code = "error"
+)
+
+// Status is the current status of a resource.
+type Status struct {
+	// Code for this status.
+	Code Code `json:"code"`
```
douglas-reid: this is potentially confusing, with status.code on grpc reflecting a few well-known status codes (which are different than just strings).

douglas-reid (commented):
Can we get this another way, that doesn't expose a new endpoint?

Maybe a prom metric that looks like:

```
name: config_status
labels:
- resource
- status
- message
```

pkg/config/store/store2.go
mandarjog: Changed names to be like `PodStatus`

mandarjog:
Encoding status subresource as metrics does not seem right.

I changed the endpoint to `/debug/config/resources/`

Status as subresource is a standard k8s concept, we are adding it our resources.
Every K8s resource implements `/status` in a different way. `PhasePPP` is a common pattern.
We can have detailed discussion about what goes into our status, however what Is already there is useful and not far off from some other k8s statuses. 

It is fairly easy to replace current ResourceStatus with whatever we evolve it to.

geeknoid:
Should we punt on this PR and address the functionality in the broader context of #1495.

geeknoid:

Thanks.

PR #1326, opened 2017-09-19
Add more unittests for expr/func.go

### LINKED ISSUE #1295: Improve test coverage for //pkg/expr/func.go
https://codecov.io/gh/istio/mixer/src/master/pkg/expr/func.go

Some of the new methods added to the AST implementation need more thorough testing.

xiaolanz:
fix #1295

**Release note**:

```release-note-none
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1326)

### DIFF
--- a/pkg/expr/func.go
+++ b/pkg/expr/func.go
@@ -126,7 +126,7 @@ func (f *eqFunc) call(args0 interface{}, args1 interface{}) bool {
 		}
 		return matchWithWildcards(s0, s1)
 	case []byte:
-		fmt.Printf("IP len %d, v4 %d, v6 %d\n", len(args0.([]byte)) , net.IPv4len, net.IPv6len)
+		fmt.Printf("IP len %d, v4 %d, v6 %d\n", len(args0.([]byte)), net.IPv4len, net.IPv6len)
 		if len(args0.([]byte)) == net.IPv4len || len(args0.([]byte)) == net.IPv6len {
 			// TODO: have the types be net.IP earlier, so this hack isn't necessary
 			ip1 := net.IP(args0.([]byte))
--- a/pkg/expr/func.go
+++ b/pkg/expr/func.go
@@ -126,6 +126,7 @@ func (f *eqFunc) call(args0 interface{}, args1 interface{}) bool {
 		}
 		return matchWithWildcards(s0, s1)
 	case []byte:
+		fmt.Printf("IP len %d, v4 %d, v6 %d\n", len(args0.([]byte)) , net.IPv4len, net.IPv6len)
 		if len(args0.([]byte)) == net.IPv4len || len(args0.([]byte)) == net.IPv6len {
 			// TODO: have the types be net.IP earlier, so this hack isn't necessary
 			ip1 := net.IP(args0.([]byte))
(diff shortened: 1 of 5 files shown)


xiaolanz:
@douglas-reid how would you add unittest for func (f *ipFunc) Call() and func (f *matchFunc) Call()  Do you want me to create a package call() methods just for testing like the others?

mandarjog:
@xiaolanz you can run `bin/linters.sh` locally to avoid failing lints.

pkg/expr/func_test.go
```diff
@@ -52,20 +52,21 @@ func TestEQFunc(tt *testing.T) {
 		{"svc1.ns1.cluster", "*.ns1.cluster1", false},
 		{net.ParseIP("10.3.25.1"), net.ParseIP("10.3.25.1"), true},
```
mandarjog: How was this test case working prior to the fix ?

ZackButcher: Magic

xiaolanz: This test works before because all the IP address test goes through the "default": deepEqual. This explains why that chunk of code shows "not hit" in test coverage.

pkg/expr/func.go
```diff
@@ -126,13 +124,17 @@ func (f *eqFunc) call(args0 interface{}, args1 interface{}) bool {
 		}
 		return matchWithWildcards(s0, s1)
 	case []byte:
-		if len(args0.([]byte)) == net.IPv4len || len(args0.([]byte)) == net.IPv6len {
+		return bytes.Equal(s0, args1.([]byte))
+	case net.IP:
```
douglas-reid: this doesn't seem right. this assumes that things are getting passed in as net.IPs, when they are actually getting passed in []byte. This PR will break existing use cases -- where our expressions pass in []bytes and not IP address.

we need this case to be []byte.

pkg/expr/func.go
```diff
@@ -126,13 +124,17 @@ func (f *eqFunc) call(args0 interface{}, args1 interface{}) bool {
...
+		if len(s0) == net.IPv4len || len(s0) == net.IPv6len {
```
douglas-reid: this is only needed in the []byte case. we don't need it for cases where we already know we have net.IPs.

xiaolanz:
PTAL now func is 99% covered. expr is 97%

xiaolanz:
PTAL @douglas-reid

pkg/expr/func.go
```diff
@@ -126,13 +126,14 @@ func (f *eqFunc) call(args0 interface{}, args1 interface{}) bool {
 		}
 		return matchWithWildcards(s0, s1)
 	case []byte:
-		if len(args0.([]byte)) == net.IPv4len || len(args0.([]byte)) == net.IPv6len {
```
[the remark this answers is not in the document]
xiaolanz: The code change is really a style change and performance enhancement (removing multiple casting). I understand it's code freeze but I think it's quite safe.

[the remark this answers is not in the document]
douglas-reid: `s0` is never cast to []byte. How can the `len(s0)` work?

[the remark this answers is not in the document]
douglas-reid: Sure. The general principle was avoiding any code changes though.

I'll approve, but we should try to resist the urge to clean-up code for tests during the freeze if possible.

xiaolanz:
see type switching https://tour.golang.org/methods/16

douglas-reid:
why did we close this?

xiaolanz:
I thought it was merged. ...

douglas-reid:
ha! ok. let's see if we can get the system to do it properly.

PR #1327, opened 2017-09-20
Replacement for istio-protocol == "tcp" by processing for context.protocol 

mandarjog:
Why do we need this PR:
This PR replaces [Label Filter](https://docs.google.com/document/d/1fGZpgFWJZhNRlQoBlCW815aOFR-ewfxKhla0CcPRsBM/edit#heading=h.bbyxh2ohz3mt) processing with standard `match` expressions.  Label filter was a new feature proposed in the linked document, and it was implemented for 0.2. However K8s labels are not to be used for this purpose. This PR replaces that feature with match selectors. We should get this into 0.2 because we don't want to start writing configs based on label filters.

1. `context.protocol` is extracted from expression so that effectively it will always be evaluated first regardless of where it appears in the expression.

    for example ordinarily 
```yaml
match: connection.connection.sent.bytes > 1000 && context.protocol == "tcp"
``` 
will result in an error because connection.connection.sent.bytes is not available in any other context.

specifying `match: context.protocol == "tcp"` is identical to 
```yaml
labels:
  istio-protocol: tcp
```
2. Add equality check extractor to expr.

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1327)

### DIFF
--- a/pkg/expr/expr.go
+++ b/pkg/expr/expr.go
@@ -408,9 +408,20 @@ func ExtractEQMatches(src string) (map[string]interface{}, error) {
 	return eqMap, nil
 }
 
-func recordEQ(fn *Function, eqMap map[string]interface{}) {
-	if fn.Name == "EQ" && fn.Args[0].Var != nil && fn.Args[1].Const != nil {
+func recordIfEQ(fn *Function, eqMap map[string]interface{}) {
+	if fn.Name != "EQ" {
+		return
+	}
+
+	// x == "y"
+	if fn.Args[0].Var != nil && fn.Args[1].Const != nil {
 		eqMap[fn.Args[0].Var.Name] = fn.Args[1].Const.Value
+		return
+	}
+
+	// yoda style, "y" == x
+	if fn.Args[0].Const != nil && fn.Args[1].Var != nil {
+		eqMap[fn.Args[1].Var.Name] = fn.Args[0].Const.Value
 	}
 }
 
@@ -419,12 +430,15 @@ func extractEQMatches(ex *Expression, eqMap map[string]interface{}) {
 	if ex.Fn == nil {
 		return
 	}
-	// only care about AND and EQ functions.
-	if ex.Fn.Name != "LAND" && ex.Fn.Name != "EQ" {
+
+	recordIfEQ(ex.Fn, eqMap)
+
+	// only recurse on AND function.
+	if ex.Fn.Name != "LAND" {
 		return
 	}
-	//TODO remove collected equality expression from AST
-	recordEQ(ex.Fn, eqMap)
+
+	//TODO remove collected equality expressions from AST
 	for _, arg := range ex.Fn.Args {
 		extractEQMatches(arg, eqMap)
 	}
--- a/pkg/expr/expr.go
+++ b/pkg/expr/expr.go
@@ -390,6 +390,43 @@ func Parse(src string) (ex *Expression, err error) {
 	return ex, nil
 }
 
+// ExtractEQMatches returns a list of attribute <names, values>
+// in literal equality comparisons. The invariant is that if
+// **any* of these comparisons is False, the expression will evaluate to false.
+// For example `destination.service` == "abc"
+// context.protocol == "TCP"
+// can be preprocessed for static equality matches.
+func ExtractEQMatches(src string) (map[string]interface{}, error) {
+	ex, err := Parse(src)
+	if err != nil {
+		return nil, err
+	}
+	eqMap := make(map[string]interface{})
+	extractEQMatches(ex, eqMap)
+	return eqMap, nil
+}
+
+func recordEQ(fn *Function, eqMap map[string]interface{}) {
+	if fn.Name == "EQ" && fn.Args[0].Var != nil && fn.Args[1].Const != nil {
+		eqMap[fn.Args[0].Var.Name] = fn.Args[1].Const.Value
+	}
+}
+
+// parseEQMatches traverse down "LANDS" and record EQs of variable and constants.
+func extractEQMatches(ex *Expression, eqMap map[string]interface{}) {
+	if ex.Fn == nil {
+		return
+	}
+	// only care about AND and EQ functions.
+	if ex.Fn.Name != "LAND" && ex.Fn.Name != "EQ" {
+		return
+	}
+	recordEQ(ex.Fn, eqMap)
+	for _, arg := range ex.Fn.Args {
+		extractEQMatches(arg, eqMap)
+	}
+}
+
 // DefaultCacheSize is the default size for the expression cache.
 const DefaultCacheSize = 1024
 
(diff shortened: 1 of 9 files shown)


testdata/config/tcpMetrics.yaml
```diff
@@ -3,8 +3,6 @@ kind: metric
 metadata:
   name: tcpbytesent
   namespace: istio-config-default
-  labels:
-    istio-protocol: tcp # needed so that mixer will only generate when context.protocol == tcp
```
mandarjog: matches are only relevant for rules.

mandarjog:
@guptasu @douglas-reid we should get this in. PTAL.

pkg/expr/expr.go
```diff
@@ -390,6 +390,46 @@ func Parse(src string) (ex *Expression, err error) {
...
+	if err != nil {
+		return nil, err
+	}
+	eqMap := make(map[string]interface{})
+	extractEQMatches(ex, eqMap)
+	return eqMap, nil
+}
+
+func recordEQ(fn *Function, eqMap map[string]interface{}) {
+	if fn.Name == "EQ" && fn.Args[0].Var != nil && fn.Args[1].Const != nil {
```
jmuk: nit: shouldn't it accept yoda-conditions (i.e. `"abc" == destination.service` style)?

pkg/expr/expr_test.go
```diff
@@ -63,6 +65,81 @@ func TestGoodParse(t *testing.T) {
...
+			src:  `substring(a, 5) == "abc"`,
+			m:    map[string]interface{}{},
+		},
+		{
+			desc: "single EQ check",
+			src:  `origin.host == "9.0.10.1"`,
+			m: map[string]interface{}{
+				"origin.host": "9.0.10.1",
+			},
+		},
```
jmuk: maybe nicer to have a case of the opposite direction, i.e. `"9.0.10.1" == origin.host`

pkg/runtime/controller.go
```diff
@@ -331,11 +331,31 @@ const (
...
+func buildRule(k store.Key, r *cpb.Rule, rt ResourceType) (*Rule, error) {
+	if r.Match != "" {
+		m, err := expr.ExtractEQMatches(r.Match)
+		if err != nil {
+			return nil, err
+		}
+		if v := m[ContextProtocolAttributeName]; v == ContextProtocolTCP {
+			rt.protocol = protocolTCP
+		}
+		// TODO process destination.service and destination.name
```
guptasu: This is a new feature... I think, consider removing the TODO. Someone might think it is denoting incomplete code

rshriram:
In general I am okay with this change as it impacts ux and adds unnecessary backward compatibility burden on us. That said I am concerned that e2e tests with tcp might break - I don't know how it's still passing! Should the mixer config in e2e test be out of date with respect to this pr ? @ZackButcher would like your opinion.

ZackButcher:
We'd discussed this change on Wednesday during the Mixer weekly, we definitely want to get it in for 0.2. Given that this passes our tests, the proxy was already populating `context.*` attributes @mandarjog?

ZackButcher:
I approve from the release manager perspective, will leave to Doug/Sunny to finish the actual code review.

mandarjog:
@rshriram e2e tests pass because I have "added" new code, but not removed the code that processes labels.

Once this gets in, I will update e2e tests and config samples.

pkg/runtime/controller.go
```diff
@@ -331,11 +331,31 @@ const (
 	istioProtocol = "istio-protocol"
 )
 
+// buildRule builds runtime representation of rule based on match condition
+func buildRule(k store.Key, r *cpb.Rule, rt ResourceType) (*Rule, error) {
+	if r.Match != "" {
```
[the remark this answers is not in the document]
douglas-reid: yeah, it'd be nice if we handled the `len(r.Match) == 0` case first with a return and then the body of the this method could lose the indent.

pkg/runtime/controller.go
```diff
@@ -331,11 +331,31 @@ const (
...
+		m, err := expr.ExtractEQMatches(r.Match)
+		if err != nil {
+			return nil, err
+		}
+		if v := m[ContextProtocolAttributeName]; v == ContextProtocolTCP {
```
douglas-reid: just `if m[ContextProtocolAttributeName] == ContextProtocolTCP {` ?

testdata/config/prometheus.yaml
```diff
@@ -92,9 +92,8 @@ kind: rule
 metadata:
   name: promtcp
   namespace: istio-config-default
-  labels:
-    istio-protocol: tcp # needed so that mixer will only execute when context.protocol == TCP 
 spec:
+  match: context.protocol == "tcp"
```
douglas-reid: is our style recommendation that `context.protocol` clauses should come first?  or second, after `destination.service` clauses?   anywhere?

pkg/expr/expr.go
mandarjog: good point, added comparison and tests.

pkg/expr/expr.go
```diff
@@ -390,6 +390,46 @@ func Parse(src string) (ex *Expression, err error) {
...
+	}
+}
+
+// parseEQMatches traverse down "LANDS" and record EQs of variable and constants.
+func extractEQMatches(ex *Expression, eqMap map[string]interface{}) {
+	if ex.Fn == nil {
+		return
+	}
+	// only care about AND and EQ functions.
+	if ex.Fn.Name != "LAND" && ex.Fn.Name != "EQ" {
```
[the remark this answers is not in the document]
mandarjog: refactored code.

pkg/expr/expr_test.go
mandarjog: done.

pkg/runtime/controller.go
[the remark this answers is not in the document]
mandarjog: done.

pkg/runtime/controller.go
mandarjog: done.

testdata/config/prometheus.yaml
mandarjog: Our code does not require anything specific. It would be an unnecessary restriction.
So `anywhere`

douglas-reid: This is a bit surprising, given that we want `destination.service` to always be the first (per previous discussions).

I get that the code doesn't require an ordering, but I would figure that we would want to recommend a "best practice" of sorts here too.

mandarjog: In our examples we can choose destination.service to be first when both are present. But I don't want to state it as an additional best practice. Our config has enough real elements for users to worry about. 

Users can choose to follow our examples.

mandarjog:
@douglas-reid @jmuk @guptasu ping !

PR #1338, opened 2017-09-21
Make StringTable size expandable.

ozevren:
**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1338)

### DIFF
--- a/pkg/il/strings.go
+++ b/pkg/il/strings.go
@@ -19,8 +19,8 @@ import (
 )
 
 const (
-	// maxStrings is the maximum number of strings that can be placed in the strings table.
-	maxStrings = 100
+	// allocSize is the allocation size when extending the string table.
+	allocSize  = 512
 	nullString = "<<DEADBEEF>>"
 )
 
@@ -37,7 +37,7 @@ func newStringTable() *StringTable {
 
 	t := &StringTable{
 		stringToID: make(map[string]uint32),
-		idToString: make([]string, maxStrings),
+		idToString: make([]string, allocSize),
 		nextID:     0,
 	}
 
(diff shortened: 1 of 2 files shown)


ozevren:
Noticed this issue after looking at the code again. This change makes the StringTable size expandable.

guptasu:
@ozevren if this needs to go in for 0.2, please ping on the Mixer slack asking the oncall to take a look and get this in.

ozevren:
fixes #1355

linsun:
what is the user impact of this for 0.2?  if not critical, i'd prefer to defer to 0.3.

ozevren:
The user impact is big. The string table interns strings that are coming in
runtime. A fixed size means, over time the table will be full.

geeknoid:
Yes, looks fine to me.

mandarjog:
@ozevren successive failures, this needs to be investigated.

kyessenov:
please investigate
/do-not-merge

ozevren:
Retesting again.

ozevren:
It looks like the previous failures were due to an issue that was fixed (fix: https://github.com/istio/mixer/commit/c99106271a77d89fa082fb58fa3a4be8a2de3d3f).

rebased on the latest commit, reran all the tests twice, and both passed. Planning to commit soon.

ozevren:
Kuat, Lin, could one of you remove the do-not-merge label?

mandarjog:
@kyessenov @linsun all tests pass. Previously it was failing on the duplicate config error, which has been fixed in another PR.

/lgtm

kyessenov:
prow rebases before running tests. was the fixed not merged prior to running the tests?

linsun:
I don't think i have permission to remove that label.

ldemailly:
@ozevren can you what exactly ends up in idToString  - specifically are there strings in there that come from the request. for instance for an http incoming request, can any of:
- uri path
- header names
- header values
or any other unbound data come into those ?

ozevren:
Yes. idToString currently captures strings that it ends up using in
evaluation of expressions. (It can be header names and or valeus)

ldemailly:
so the whole attribute bag and each key and value in string maps end up in there whether used or not, are you sure ?

if so can we file a followup issue to cleanup as this will be ever growing memory / should show up as a memory leak ?

ozevren:
No, not all of it. Only the ones used when evaluation a particular expression that is needed to resolve the request.

I've filed an issue and also have a potential short-term fix. 
https://github.com/istio/mixer/issues/1369
https://github.com/istio/mixer/pull/1370

PR #1344, opened 2017-09-23
Fix doubleCounting when service namespace is defaultConfigNamespace

mandarjog:
When defaultConfigNamespace was the same as the
service namespace for a request, the same configuration
was processed twice. This resulted in double dispatch of
handlers.

This PR fixes that problem.

PR https://github.com/istio/istio/pull/909 introduces tests that puts defaultConfiguration in the same namespace as istio-system.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1344)

### DIFF
--- a/pkg/runtime/resolver.go
+++ b/pkg/runtime/resolver.go
@@ -174,10 +174,9 @@ func destAndNamespace(attrs attribute.Bag, idAttr string) (dest string, ns strin
 		glog.Warningf(msg)
 		return "", "", errors.New(msg)
 	}
-	var ok bool
 
-	dest, ok = attr.(string)
-	if !ok {
+	var ok bool
+	if dest, ok = attr.(string); !ok {
 		msg := fmt.Sprintf("%s identity must be string: %v", idAttr, attr)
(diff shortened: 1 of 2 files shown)


andraxylia:
Thanks for fixing this @mandarjog

pkg/runtime/resolver.go
```diff
@@ -126,36 +127,26 @@ func (r *resolver) Resolve(attrs attribute.Bag, variety adptTmpl.TemplateVariety
 		resolveActions.With(lbls).Observe(float64(raLen))
 	}()
 
-	attr, _ := attrs.Get(r.identityAttribute)
-	if attr == nil {
-		msg := fmt.Sprintf("%s identity not found in attributes%v", r.identityAttribute, attrs.Names())
-		glog.Warningf(msg)
-		return nil, errors.New(msg)
+	if target, ns, err = destAndNamespace(attrs, r.identityAttribute); err != nil {
```
douglas-reid: nit: should we use `dest` instead of `target` for consistency?

Issue #1347, opened 2017-09-23
Port Mixer to build using CentOS docker-signed container contents


sdake:
Mixer currently builds against debian/ubuntu base images.  This results in some incompatibilities when running on CentOS7.4 kernels as the Xenial kernel and CentOS kernel are not 100% compatible.

Instead permit MIxer to build against a CentOS  target OS.

This is part of a larger port of Istio to CentOS described here: https://github.com/istio/issues/issues/56

geeknoid:
Anything else to do, or can we close this?

sdake:
@geeknoid I've tested master last week on CentOS and it works well.  I don't see any further need to keep this open.  I'm not sure if I originally opened it because of other bugs, or things magically worked themselves out.   We can always reopen a new issue if people run into trouble.

PR #1348, opened 2017-09-24
 Port Mixer to build on CentOS systems

sdake:
Port Mixer to build on CentOS systems
    
Fixes (#1347)
    
This change uses the CentOS 7.4 upstream docker image to
build mixer against.  This naturally requires changes to
the istio repository to work with the way images are tagged
Today          Post Change
                    +-> mixer_ubuntu:SHA
mixer:SHA -|
                    +-> mixer_centos:SHA
    
This naturally requires changes to the istio repository to
work correctly.  I have istio/istio changes ready to go,
but don't know how to submit a prow cross-repository dependency
requirement (to block this patch on the istio changes for that
change to the manifest templates and updateTemplate.sh to all
merge as one unit).

**Release note**:

NONE

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1348)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -121,12 +121,31 @@ new_http_archive(
 
 DEBUG_BASE_IMAGE_SHA = "3f57ae2aceef79e4000fb07ec850bbf4bce811e6f81dc8cfd970e16cdf33e622"
 
+# Official CentOS Cloud Sig Docker images for CentOS 7.4.1708 - Sep 11th, 2017
+new_http_archive(
+    name = "docker_centos",
+    build_file = "BUILD.centos",
+    sha256 = "ae5a842b3bbdfa00e0dda7cb076c9e8040191737e36513dd8fe0a9b7d569abe1",
+    type = "xz",
+    url = "https://github.com/CentOS/sig-cloud-instance-images/raw/CentOS-7.4.1708/docker/centos-7.4.1708-docker.tar.xz",
+)
+
 # See github.com/istio/manager/blob/master/docker/debug/build-and-publish-debug-image.sh
 # for instructions on how to re-build and publish this base image layer.
 http_file(
     name = "ubuntu_xenial_debug",
     sha256 = DEBUG_BASE_IMAGE_SHA,
     url = "https://storage.googleapis.com/istio-build/manager/ubuntu_xenial_debug-" + DEBUG_BASE_IMAGE_SHA + ".tar.gz",
+)
+
+# TODO(sdake) Need to sort out how to get a centos_74_debug image into istio-build
+# Official CentOS Cloud Sig Docker images for CentOS 7.4.1708 - Sep 11th, 2017
+new_http_archive(
+    name = "docker_centos_debug",
+    build_file = "BUILD.centos",
+    sha256 = "ae5a842b3bbdfa00e0dda7cb076c9e8040191737e36513dd8fe0a9b7d569abe1",
+    type = "xz",
+    url = "https://github.com/CentOS/sig-cloud-instance-images/raw/CentOS-7.4.1708/docker/centos-7.4.1708-docker.tar.xz",
 )
 
 go_repository(
--- a/BUILD.centos
+++ b/BUILD.centos
@@ -1,0 +1,9 @@
+# build base docker centos image
+
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
+
+docker_build(
+    name = "centos",
+    tars = ["centos-7.4.1708-docker.tar.xz"],
+    visibility = ["//visibility:public"],
+)
--- a/bin/publish-docker-images.sh
+++ b/bin/publish-docker-images.sh
@@ -25,7 +25,7 @@ function docker_push() {
   fi
 }
 
-BAZEL_IMAGES=('mixer' 'mixer_debug')
+BAZEL_IMAGES=('mixer_ubuntu' 'mixer_ubuntu_debug' 'mixer_centos' 'mixer_centos_debug')
 IMAGES=()
 TAGS=''
 HUBS=''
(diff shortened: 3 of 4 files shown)


douglas-reid:
cool. thanks for tackling this. is there also work to do this for other Istio components? at one point, we were working to having a common base layer for all the components, but I don't quite remember what the current status of that is.

this makes me a bit nervous, only because we are on the verge of a release. changing all the build image stuff could disrupt things -- and, as you note, there is a delicate cross-repo dance that has to be done to update istio/istio.

Would it be a problem to hold off on this until the current release dust settles a bit?

sdake:
@douglas-reid Yes sorry I didn't get a chance to leave a comment for this set of patches (had dinner with son and, barely got this review pushed before dinner).  I have this working for mixer and pilot and istio repos, however, I haven't cleaned up the pilot or istio patches yet enough to submit them.  I'll work on finishing the job on components this week in the event people want to test centos based binaries.

I would have mentioned if I had time before dinner I think it makes total sense not to merge this work until after 0.2.z+ is tagged and rebasing the various non-doc patches (i.e. code change patches); I really don't want to disrupt the release process at all.

I'd really appreciate reviews on these patches (as to the approach) to make sure they are correct.  I sort of think a tagging of mixer:ubuntu-SHA/mixer:centos-SHA would make more sense, however, I didn't have a chance to implement that although would be happy to do so if the reviewers wish.

If you have a link to the common base layer PR or issue, I can work from that.  I think it would make alot of sense to use distroless + static binaries with that, however, I'm not quite sure how to do that work immediately nor if that would work for CentOS and Ubuntu.

One thing I would appreciate is if someone could run this through CI if that is possible.  I'm not sure if prow permits cross-repo dep testing or not.  I'm also not quite sure how to launch the e2e testing in my local environment other then bringing up the changes myself which I have done in my lab.

Cheers,
-steve

douglas-reid:
@sdake no worries. i'll try to get some discussion going on Monday with the rest of team about the approach on tagging, etc.

i don't have a link to any active common base layer PR, but I can point you to a PR (https://github.com/istio/mixer/pull/886) where I tried to switch Mixer to distroless. At the time, it was felt that it was easier/better to stick with xenial. This issue came up again in https://github.com/istio/istio/issues/822.

cc'ing @ldemailly here, as he has spent some time thinking about base images and can probably help drive consensus around an approach.

WORKSPACE
```diff
@@ -121,6 +121,14 @@ new_http_archive(
 
 DEBUG_BASE_IMAGE_SHA = "3f57ae2aceef79e4000fb07ec850bbf4bce811e6f81dc8cfd970e16cdf33e622"
 
+new_http_archive(
+    name = "docker_centos",
+    build_file = "BUILD.centos",
+    sha256 = "9e945e0ea9bf66c73e6695f7d70a9c0bbd3f8f5c67b9a44e3127a4ac495e2d85",
+    type = "xz",
+    url = "https://github.com/CentOS/sig-cloud-instance-images/raw/16dab97b0ce72b1db7a2f9b02c76e452cb0a63cb/docker/centos-7-docker.tar.xz",
```
mandarjog: A sha like `16dab97b0ce72b1db7a2f9b02c76e452cb0a63cb` should have an explanation. 
Does this correspond to a tag? If not what is the date that it was published.

ldemailly:
Quick question: if mixer is running through docker as it is, why does it matter what the base image is ? (for onprem in 0.3 it'd be great to have packages for each component but that wouldn't be docker images)

sdake:
@ldemailly the Ubuntu container images do not work on CentOS base OS + Kubernetes CentOS versions from what I can tell. Some folks want to run Kubernetes on CentOS using the OpenStack (or vsphere or insert cloud provider here) cloud provider model.

From past experience, some containers need to be compiled for the base operating system (e.g. if a container is running on centos, it needs to be built against a centos version of the toolchains/libraries, especially those headers in /usr/include as the kernel headers are slightly different between distributions).  The docker core dev's have acknowledged this problem in docker-dev on IRC in the past.  Over the last four years of container dev I've been involved with, the core devs of Docker have indicated a container is not portable between operating systems in some cases. My personal experience with these cases include the RTNETLINK interface to the kernel, the file capabilities interfaces, the syscall interfaces, and the ioctl interfaces.  There are likely other binary interfaces to kernel.org which are not compatible between various distributions. This is one of a variety of reasons kernel.org has semi-frozen binary interface development (except for syscall, as that is designed into unix/Linux designs).

Let me know if you need more information.

Using FROM scratch is an interesting idea.  This requires all of the various system level shared object dependencies to have static library .a deps such that the binaries can be statically linked.  I'm not convinced static libraries for the various distributions are available for the various distro choices we would choose to make in Istio, but could be wrong (or they could be generated).  I do think that is a cool idea and would shrink the size of the images - although that feels like an orthogonal and separate piece of work (as it doesn't resolve the kernel interface incompatibility problem).

Cheers,
-steve

ldemailly:
I found your overall issue, and like I asked there, can you give a very specific example of what is not working exactly - afaik from an OS point of view mixer is a basic grpc server, nothing arcane there.

Building go `FROM scratch` is pretty easy (see for instance https://github.com/istio/istio/blob/master/devel/fortio/Dockerfile ) and that's where I think we should head rather until proven that it has specific issues

Do you have any link to what you mention and/or specific issues with mixer ?

sdake:
@ldemailly please reference https://github.com/istio/istio/issues/922#issuecomment-332728702.  I think the issue may now only involve building CentOS binaries and running them in the current containers (as they have OS deps in them) using the current bazel docker build & dockerfiles.  I will verify.

ldemailly:
thanks steven, let me know if you encounter issues

mandarjog:
@sdake per your comments in the issue, this is no longer needed. If this is so can you close the PR ?

sdake:
@mandarjog For the moment, I'll close this pull request (but leave the issue open) in case someone else wants to take a crack at a solution to this problem.

Cheers
-steve

PR #1349, opened 2017-09-24
Add flag to control source/origin lookups when destination.service is istio ingress

douglas-reid:
As noted in istio/istio#879, the source information that the Mixer receives (`source.ip` values) are problematic for attempting to lookup (in-cluster). This PR addresses provides a workaround for compounding that issue within Mixer by preventing lookup of source/origin information in the kubernetes adapter when `destination.service == ingress.istio-system.svc.cluster.local`.

This behavior is configurable, through options in the config params passed into the adapter builder.

This PR also cleans up a related bit of the kubernetes adapter behavior related to handling and generation of IP_ADDRESS values. As noted in istio/mixerclient#112 (and also in istio/proxy#483), handling of IP_ADDRESS values is completely inconsistent throughout Istio. The changes here remove any inconsistencies on the part of the kubernetes adapter.

It also fixes the incorrect IP-related config in `testdata/configroot`.

**Release note**:

```release-note-none
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1349)

### DIFF
--- a/adapter/kubernetes/kubernetes.go
+++ b/adapter/kubernetes/kubernetes.go
@@ -86,9 +86,11 @@ const (
 	serviceVal        = "Service"
 
 	// value extraction
-	clusterDomain        = "svc.cluster.local"
-	podServiceLabel      = "app"
-	istioPodServiceLabel = "istio"
+	clusterDomain                      = "svc.cluster.local"
+	podServiceLabel                    = "app"
+	istioPodServiceLabel               = "istio"
+	lookupIngressSourceAndOriginValues = false
+	istioIngressSvc                    = "ingress.istio-system.svc.cluster.local"
 
 	// cache invaliation
 	// TODO: determine a reasonable default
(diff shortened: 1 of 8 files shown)


douglas-reid:
Ping. We need to decide what we want to do for 0.2 here.

adapter/kubernetes/kubernetes.go
```diff
@@ -86,9 +86,11 @@ const (
...
 
 	// value extraction
-	clusterDomain        = "svc.cluster.local"
-	podServiceLabel      = "app"
-	istioPodServiceLabel = "istio"
+	clusterDomain                      = "svc.cluster.local"
+	podServiceLabel                    = "app"
+	istioPodServiceLabel               = "istio"
+	lookupIngressSourceAndOriginValues = false
+	istioIngressSvc                    = "ingress.istio-system.svc.cluster.local"
```
ldemailly: that doesn't seem to match https://github.com/istio/istio/blob/master/install/kubernetes/istio.yaml#L1108 or is this not the name ? 
what test exercises this ?

douglas-reid: It does actually match. You just aren't accounting for the mixer matching of service names. It matches through the "istio" label. And, in this case, the "istio" label value is "ingress".

The only tests for this behavior are in this PR itself.

ldemailly: istio-ingress.istio-system.svc.cluster.local is the service name
if we have some other lookup it shouldn't look like the dns name svcname.namespace.svc.cluster.local is well defined, where/why are we changing this ?

mandarjog: `.svc.cluster.local` is configurable, so it makes sense for the `fully_qualified_istio_ingress_service_name` to be configurable.

douglas-reid: it is configurable in this PR.  the value here is the default value.

PR #1358, opened 2017-09-27
reduce per request log flood

ldemailly:
reduce the log flood / avoids
```
I0927 16:37:27.035598   29029 grpcServer.go:223] AccessLog Quota foo.svc.cluster.local 1/1 
```
for every request

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1358)


mandarjog:
Access log *is* per request so this is removing access log.

How does access log work in envoy? @kyessenov?

kyessenov:
It's on by default, but out of request data path and can be disabled.

ldemailly:
no I'm not touching access log, this is just removing that stderr log about quota in each request path

Issue #1155, opened 2017-08-28
bin/linters.sh is painfully slow


jmuk:
I've been feeling that linters.sh is _very_ slow. On my HP z620 workstation, a run of this script for checking all files takes >7 minutes.

Digging into some of the details, I noticed that `gometalinter.v1` seems the root cause; if we run the head of gometalinter, the performance is improved significantly, it finishes less than a minute (~24 seconds here). I have no idea why / when it's so improved though.

I think the main point of pinning gometalinter through gopkg.in is for pinning linter version (so that it won't report additional errors on existing code), but I believe we should do something better.

douglas-reid:
We pinned to `v1` to attempt to prevent differences in linter results across environments. Any other approach to ensuring consistent linting is also welcome.

I'm skeptical that there was a 14x speed-up across versions. Are you sure it ran the same set of linters on the same set of code?

Gometalinter documentation includes the warning:
> Note that this can be very slow, and you may need to increase the linter --deadline to allow linters to complete.

And I've had to increase the deadline as our codebase has grown for changes that cause the entire repo to be linted.

jmuk:
Checked this a bit more.

The HEAD of `gometalinter` is indeed very fast, with the exactly same set of linters as `v1` has. I skipped `prep_linters` from linters.sh, but still have this result.

Looked into the actual commandlines, I've noticed that the HEAD invokes linters with actual target go files, while `v1` invokes with `.`, I think this is the cause of this huge change.

jmuk:
#1363 fixes this -- at least on my environment. Closing.

PR #1363, opened 2017-09-29
Use recent gometalinter

kyessenov:
Takes the time to run the linter from >20 minutes to 30s

- Eliminates indirection in vendor->genfiles link, since genfiles is also a symbolic link.
- Use docker run to run a frozen linter bunch
- Fix new linter errors

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1363)

### DIFF
--- a/bin/linters.sh
+++ b/bin/linters.sh
@@ -8,29 +8,38 @@ source $SCRIPTPATH/use_bazel_go.sh
 ROOTDIR=$SCRIPTPATH/..
 cd $ROOTDIR
 
-
-PARENT_BRANCH=''
-
-while getopts :c: arg; do
-  case ${arg} in
-    c) PARENT_BRANCH="${OPTARG}";;
-    *) { echo "Unrecognized argument ${OPTARG}"; exit 1; };;
-  esac
-done
-
 prep_linters() {
     bin/bazel_to_go.py
 }
 
 go_metalinter() {
-    NUM_CPU=$(getconf _NPROCESSORS_ONLN)
+    LINTER_SHA="bfcc1d6942136fd86eb6f1a6fb328de8398fbd80"
 
-    # Note: WriteHeaderAndJson excluded because the interface is defined in a 3rd party library.
-    gometalinter="docker run \
-      -v $(bazel info output_base):$(bazel info output_base) \
-      -v $(pwd):/go/src/istio.io/mixer \
-      -w /go/src/istio.io/mixer \
-      gcr.io/istio-testing/linter:bfcc1d6942136fd86eb6f1a6fb328de8398fbd80"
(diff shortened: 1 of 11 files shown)


bin/linters.sh
```diff
@@ -19,82 +19,53 @@ while getopts :c: arg; do
...
         --enable=unparam\
-        --enable=unused\
         --enable=varcheck\
         --enable=vet\
         --enable=vetshadow\
         --skip=testdata\
         --skip=vendor\
         --vendor\
-        $PKGS
+        ./...
```
mandarjog: We do want to maitain he option of running it on specific packages.

pkg/config/validator.go
```diff
@@ -264,14 +264,14 @@ func (a adapterKey) String() string {
 // compatfilterConfig
 // given a yaml file, filter specific keys from it
 // globalConfig contains descriptors and adapters which will be split shortly.
-func compatfilterConfig(cfg string, shouldSelect func(string) bool) ([]byte, map[string]interface{}, error) {
+func compatfilterConfig(cfg string, shouldSelect func(string) bool) ([]byte, error) {
```
mandarjog: Something required you to add a param?

bin/linters.sh
kyessenov: it only takes 30s to run it on all of codebase. I can use PKGS as an imported env variable.

pkg/config/validator.go
kyessenov: unparam complained the parameter is not used.

bin/linters.sh
geeknoid: It takes _way_ more than 30 seconds when running on a Mac. Many many minutes...

jmuk: This patch should make the linter extremely faster than the one on master branch. As I reported in #1155, gometalinter.v1 is 14x slower than the recent version, because of target file selection.

It took ~1min on my environment _including_ fetching docker images. Subsequent calls took ~30 secs.

jmuk:
This should fix #1155. Also it's worth noting that pilot already has the same trick (https://github.com/istio/pilot/pull/1363).

bin/linters.sh
kyessenov: This works great for prow and linux machines, but it's not so great on mac. I tried it on my macbook and docker there is rather slow. I'm thinking of dropping to using native gometalinter on mac? Would that work for you guys?

geeknoid:
That's fine with me.

kyessenov:
PTAL, using native gometalinter is preferable on mac. Takes about 50s instead of 30s on dockerized linux.

PR #1367, opened 2017-09-30
Gazellify the build

kyessenov:
`bin/gazelle` should automatically correct `BUILD` files

**Release note**:

```release-note
NONe
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1367)


kyessenov:
Linters time out since I touched many directories. What's the work-around?

jmuk:
I feel a bit nervous about the deps ordering after `sed` replacement. Please run `bin/fmt.sh` (which invokes `buildifier`) to fix the odering.

Re: linters timeout -- I have no idea. Maybe doing the same trick you made on pilot (using docker image) would help, I guess. Later version of gometalinter should be much faster.

guptasu:
Run the linter with `-c upstream/master`

kyessenov:
@jmuk I see what you mean by dependency ordering. Let's wait for the presubmit to finish. If the ordering is a problem, we might need to adjust the API imports IMHO from github to istio.

kyessenov:
PTAL
1. added mac support
2. started changing import to istio.io/api
3. fixed buildifier lint errors

adapter/inventory.bzl
```diff
@@ -14,15 +14,16 @@ def _inventory_gen(name, packages, out):
...
+    "//adapter/noopLegacy:go_default_library",
 ]
 
 def inventory_library(name, packages, deps):
   _inventory_gen("inventory_file_gen", packages, "inventory.gen.go")
 
   go_library(
       name = name,
-      srcs = ["inventory.gen.go"],      
+      srcs = ["inventory.gen.go", "inventory.go"],
```
mandarjog: @kyessenov we prefer lists to span multiple lines. Did buildifier not catch this ?

adapter/prometheus/BUILD
```diff
@@ -31,6 +31,7 @@ go_test(
         "//adapter/prometheus/config:go_default_library",
         "//pkg/adapter:go_default_library",
         "//pkg/adapter/test:go_default_library",
+        "//template/metric:go_default_library",
```
mandarjog: no sure how this compiled before with a missing dependency?
Was it being transitively included ?

adapter/stackdriver/log/BUILD
```diff
@@ -25,5 +25,9 @@ go_test(
         "//adapter/stackdriver/config:go_default_library",
         "//pkg/adapter/test:go_default_library",
         "//template/logentry:go_default_library",
+        "@com_google_cloud_go//logging:go_default_library",
+        "@org_golang_google_api//option:go_default_library",
+        "@org_golang_google_genproto//googleapis/api/monitoredres:go_default_library",
```
mandarjog: new dependencies without new code ?
Is this a minimal set?

adapter/inventory.bzl
jmuk: this line should be manually written, gazelle / buildifier does not correct a pattern like this (maybe because it's under a `def`)?

kyessenov: Yeah, it's a macro so buildifier does not verify the expansion.

adapter/prometheus/BUILD
kyessenov: That's likely, so it would have caused a build error down the road if the intermediate target is removed.

adapter/stackdriver/log/BUILD
kyessenov: These are probably indirectly imported. It's better to be explicit.

adapter/inventory.bzl
douglas-reid: Isn't this a problem anyway?  Do we want both "inventory.go" and "inventory.gen.go" in this library now?  This seems like a bigger problem than the lists spanning multiple lines.

jmuk: It is not a problem at all. Actually they are already in a single `//adapter:go_default_library` by using `library=` param in `adapter/BUILD` (https://github.com/istio/mixer/blob/master/adapter/BUILD#L11), this does not change anything around the library structure.

Also it's worth noting that the `library=` param in `go_library` is an undocumented hack and will not work anymore with the latest `rules_go`.

douglas-reid: But, having `inventory.go` in the `bzl` file seems wrong. That file is specifically about generation. Putting non-generated code into that script seems incorrect.

It is fine if we want to clean that up when `inventory.go` is completely deleted, but in the meantime, it just seems to be inverted.

kyessenov: `library` is deprecated. I'm simply trying to use bzl/go conventions here, I guess someone else can fix it later who knows the purpose of this macro.

jmuk: wondering who still needs `inventory.go`. Can we remove it now?

kyessenov:
PTAL?

adapter/inventory.bzl
mandarjog: We still use the mixer1 Kubernetes adapter which is included in inventory.

PR #1370, opened 2017-10-02
Implement a purge algorithm for removing entries from the cache/StringTable.

ozevren:
The expression cache holds on to an interned string table that captures runtime string data. This PR evicts the cached expressions when the interned strings count reaches a certain threshold.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1370)


ozevren:
fixes https://github.com/istio/mixer/issues/1369

Issue #1369, opened 2017-10-02
Mixer IL evaluation has slow memory leak.


ozevren:
The string table that interns strings in the Mixer IL is slowly accumulating runtime data. This data gets flushed when the expression cache entries get evicted. In case of steady state of no config change and large cache-size, the interned strings will accumulate.

ldemailly:
I'm still not clear exactly which strings end up in that table

can you give an example of the worst case where it would grow unbounded ? (eg if for instance the URI or a cookie would be in that table)

mandarjog:
The string intern table interns *all* strings. For example this includes *values* of all headers, which can be unbounded.

One of the reasons why this is needed is because the IL uses references  from the table instead of direct string refs. This is an intermediate fix.

Other solutions include having two intern tables. One that is request scoped that is auto purged at the end of a request. The other table holds constants that have longer lives.

ldemailly:
@mandarjog I have been trying to get a precise answer to this question 

@ozevren in #1338 was saying "No, not all of it. Only the ones used when evaluation a particular expression that is needed to resolve the request."

which is it :-) ?

ozevren:
As the code stands today, the IL evaluator will not intern strings eagerly.
It will intern them as it encounters them during string evaluation. For
example, if you have this expression: 'request.headers["a"] |
request.header["b"] | "c"', then the initial interned strings will be
"request.headers", "a", "b", and "c". As an example, assume the first
incoming request has "request_headers = {"a": "aval", "b": "bval", "d":
"dval"}". In this case, the expression will short circuit in the first
expression fragment, after evaluating "request.headers["a"]" and "aval"
will be added to the interned strings table, but not the others.

In the future, we'd like to have two lookup tables, one for constants, the
other for per-request interned string table. We should be able to get some
perf gains using this approach and not leak. But this requires some
substantial changes to the request handling path.

In the interim, my initial check-in will avoid a sudden circuit-break and
my follow-up PR (not submitted yet) should reduce the effects of interning
(i.e. over a steady state, the memory consumption should be bounded even in
the presence of open-set inputs).

Does that make sense?

kyessenov:
Have you considered a session context for the evaluator? Assuming Mixer establishes a session for a proxy, the table can be split into both a global inline table (for global constants) and a local inline table, to reuse strings across subsequent Check calls.

ldemailly:
can you just use an LRU ?

ozevren:
[...]

PR #1371, opened 2017-10-02
Implement a purge algorithm for removing entries from the cache/StringTable

ozevren:
The expression cache holds on to an interned string table that captures runtime string data. This PR evicts the cached expressions when the interned strings count reaches a certain threshold.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1371)

### DIFF
--- a/pkg/il/evaluator/evaluator.go
+++ b/pkg/il/evaluator/evaluator.go
@@ -37,10 +37,11 @@ import (
 // IL is an implementation of expr.Evaluator that also exposes specific methods.
 // Specifically, it can listen to config change events.
 type IL struct {
-	cacheSize   int
-	context     *attrContext
-	contextLock sync.RWMutex
-	fMap        map[string]expr.FuncBase
+	cacheSize                  int
+	maxStringTableSizeForPurge int
+	context                    *attrContext
+	contextLock                sync.RWMutex
+	fMap                       map[string]expr.FuncBase
 }
 
 // attrContext captures the set of fields that needs to be kept & evicted together based on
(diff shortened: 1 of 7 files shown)


ozevren:
fixes #1369

pkg/il/evaluator/evaluator.go
```diff
@@ -183,19 +186,29 @@ func (e *IL) getAttrContext() *attrContext {
...
-	if entry, err = e.getOrCreateCacheEntry(expr); err != nil {
+	if entry, err = ctx.getOrCreateCacheEntry(expr); err != nil {
 		glog.Infof("evaluator.evalResult failed expr:'%s', err: %v", expr, err)
 		return interpreter.Result{}, err
 	}
 
-	return entry.interpreter.Eval("eval", attrs)
+	r, err := entry.interpreter.Eval("eval", attrs)
+	if entry.interpreter.StringTableSize() > maxStringTableSizeForPurge {
+		ctx.cache.Remove(expr)
```
geeknoid: If I understand correctly, this means that once the max size is hit, no new expressions will be added and so new expressions run "slower" for ever more.

Would it be possible to purge an entry from the cache at random instead? This doesn't require any LRU-tracking logic, but would generally behave better in the long run.

ozevren:
With this change, if an expression ends up evaluating with inputs from a
closed set, the behavior will not change (the string table size will stay
constant and will not trigger the new logic).

If the expression evaluates with inputs from an open set, then the table
will grow until the limit and then that particular expression's cache entry
will be removed. The next call would end-up recreating the cache entry by
recompiling and creating a new string table.

This doesn't penalize expressions that use closed set (i.e. fixed set of
values, or same set of clients/adapters returning the same values due to
locality). The more varied inputs an expression has to deal with, the
faster it will be evicted.

Issue #1373, opened 2017-10-03
Mixer rule deletion does not take immediate effect in istio.io demo


myidpt:
Following https://istio.io/docs/tasks/security/secure-access-control.html. After deleting the deny rule, I can not load the "details" until >10 minutes later. However, the deletion takes effect immediately in https://istio.io/docs/tasks/security/basic-access-control.html. According to @mandarjog, Mixer client does about 10min expiration caching on attributes other than labels. That could explain the behavior. But it would be good we can show users immediate result in our demos, and it seems to me caching with 10min expiration is too long.

PR #1378, opened 2017-10-05
change deny ttl to 5 sec

mandarjog:
This PR addresses #1373 .

1. Make all denier options configurable :-) 
2. Change default ttl  to `5s` from `1000 s`

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1378)

### DIFF
--- a/adapter/denier/denier.go
+++ b/adapter/denier/denier.go
@@ -31,25 +31,33 @@ import (
 )
 
 type handler struct {
-	status rpc.Status
+	result adapter.CheckResult
+}
+
+func defaultParam() *config.Params {
+	return &config.Params{
+		Status:        rpc.Status{Code: int32(rpc.FAILED_PRECONDITION)},
+		ValidDuration: 5 * time.Second,
+		ValidUseCount: 1000,
+	}
+}
+
+func newResult(c *config.Params) adapter.CheckResult {
+	return adapter.CheckResult{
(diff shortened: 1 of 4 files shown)


adapter/denier/denier.go
```diff
@@ -39,15 +39,15 @@ type handler struct {
 func (h *handler) HandleCheckNothing(context.Context, *checknothing.Instance) (adapter.CheckResult, error) {
 	return adapter.CheckResult{
 		Status:        h.status,
-		ValidDuration: 1000 * time.Second,
+		ValidDuration: 5 * time.Second,
```
douglas-reid: Should these values be constants, annotated with comments on their setup?  Or maybe even config params for denier handlers?

douglas-reid:
In general, this seems fine to me.

I wonder, however, if we can ever decide universally what the value should be. This feels like it should have a sensible default (and maybe that is: `5s` or `1000` uses), but be configurable *per handler* (so that I can have some deny rules that I don't mind caching for 10 minutes and some that I want to be very responsive).

mandarjog:
Denier `now with more config`  :-)  PTAL

mandarjog:
@myidpt this addresses the cache time issue.

myidpt:
Great fix, Mandar 👍

PR #1381, opened 2017-10-06
stops using library= for template

jmuk:
The `library` param in go_library rule isn't supported anymore,
and soon becomes unavailable. This rule does not have to use that,
simply using proto_compile and specifying the generated files to
srcs would be okay.

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1381)


mandarjog:
@jmuk can you add a link to the deprecation ?

kyessenov:
You can try out latest rules go with latest bazel, and the rule breaks there at `library=`

jmuk:
I don't know the official comment of the deprecation, but https://github.com/bazelbuild/rules_go/issues/744#issuecomment-324490133 says

> the library line on a go_library is not a supported attribute, it's not in the documented interface, and it will cease to exist in the not too distant future...

And indeed, this parameter disappears on the head of `rules_go` as @kyessenov says. Things like this will be necessary anyways when we upgrade to go 1.9 or later.

PR #1383, opened 2017-10-09
Add timestamp() expression support

RoeyMaor:
**Release note**:
```
Add timestamp() expression support.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1383)

### DIFF
--- a/pkg/expr/eval_test.go
+++ b/pkg/expr/eval_test.go
@@ -20,11 +20,13 @@ import (
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"istio.io/mixer/pkg/attribute"
 )
 
 func TestGoodEval(tt *testing.T) {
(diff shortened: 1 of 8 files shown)


douglas-reid:
@RoeyMaor would you mind updating the PR description to have the release note be a single liner. I'm thinking the easiest thing to do would be to move the current release note to the description and replace it with a single "Add timestamp() expression support" (or similar).

douglas-reid:
All: conceptually, do we expect that operators will want to have a static fallback time, or would something like `now()` be more appropriate?  I don't see any reason not to support both, mostly just curious.

mandarjog:
I think most use cases call for `request.now` or similar.
However having a way to create literals for value types is useful.

RoeyMaor:
waiting for https://github.com/istio/mixer/pull/1429 to be merged

RoeyMaor:
@mandarjog

PR #1384, opened 2017-10-09
Update development.md

kyessenov:
Bazel aliases do not create symbolic links for `mixs` and `mixc`.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1384)


douglas-reid:
@kyessenov and others: would it be better to instead change the directory names to `mixs` and `mixc` ?  I don't think having a "server" binary is ideal.

geeknoid:
We don't want this change, we want to continue to have mixs and mixc as our two binary names.

geeknoid:
We don't want to rename our mixs and mixc commands. If the names are problematic due to the gazelle changes, we should rename the directories in istio.io/mixer/cmd accordingly.

PR #1386, opened 2017-10-09
Check gazelle changes

kyessenov:
Ensures that gazelle has been run in every PR.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1386)

### DIFF
--- a/bin/linters.sh
+++ b/bin/linters.sh
@@ -75,6 +75,12 @@ $gometalinter \
     --vendor\
     ./...
 
+echo Check license...
+bin/check_license.sh
+
+echo Check workspace...
+bin/check_workspace.sh
+
 echo Checking gazelle...
 bin/gazelle
 builddiff=$(git diff --name-only -- `find . -type f \( -name BUILD -o -name BUILD.bazel \)`)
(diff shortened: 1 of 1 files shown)


bin/linters.sh
```diff
@@ -75,6 +75,14 @@ $gometalinter \
     --vendor\
     ./...
 
+echo Checking gazelle...
+bin/gazelle
```
mandarjog: This is  does not work the common use case.

1. Update a source file; add a new import
2. Run gazelle, this will update the BUILD
3. check_gazelle will fail because I have not committed the changes yet.

The correct solution is to check before and after image of the file wrt gazelle processing.
Perhaps gazelle supports such an option, if not we will have to do the comparison.

kyessenov:
It is optimized for pre-submit, where the expectation is that the PR is complete. For local development, you can ignore the messages from this script. It runs after linters, so does not prevent useful linter messages.

mandarjog:
It prevents the script from running to completion.
It should either
1. Perform the correct check
2. Exit on error if running in pre-submit.

I prefer # 1

kyessenov:
For #1, we need a PR to gazelle to do file backup/diffing. Let's just put as a last linter, since that makes it a no-op for local development.

PR #1401, opened 2017-10-10
validator server as an independent http server

jmuk:
This is another approach of validator; it takes an http request
for a validation, which receives an operation over multiple
resources. This is different from the approach of external
admission webhook, because webhook only takes an operation of
a resource.

This webserver will be invoked by istioctl to validate the operations,
which typically contains multiple resources and the server
will validate all of them at once.

**Release note**:

```release-note
validator server is added as a custom webserver. Currently structural validations only; referential validation will come in another PR.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1401)

### DIFF
--- a/cmd/server/cmd/BUILD
+++ b/cmd/server/cmd/BUILD
@@ -10,6 +10,7 @@ go_library(
         "root.go",
         "server.go",
         "test_server.go",
+        "validator.go",
     ],
     visibility = [
         "//cmd:__subpackages__",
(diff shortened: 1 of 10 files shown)


mandarjog:
@jmuk I understand why you added batch validation semantics here. Just so that we are on the same page, this is to ensure that given a single mutli-resource file, we won't have to deal with eventual consistency of referred resources, correct? Please update the PR with `why`...

I think there is value in keeping the API in sync with the real webhook.
We just need a guarantee from the client that resources will be submitted serially to the server. The webhook can keep track of uncommitted resources so it does not have to rely on the consistent underlying view.

jmuk:
@mandarjog re: API (i.e. request/response format)
in that case maybe I should revive my older PR rather than this; but even if we can guarantee the order of objects to be operated, there can be timing issues around validation and object creation/deletion in apiserver. Let's talk offline.

jmuk:
@mandarjog please review #1218, closing this.

PR #1402, opened 2017-10-10
change the default config namespace to istio-system

jmuk:
**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1402)


test/e2e/e2e_report_test.go
```diff
@@ -81,12 +81,12 @@ apiVersion: "config.istio.io/v1alpha2"
...
   name: rule1
-  namespace: istio-config-default
+  namespace: istio-system
 spec:
   selector: target.name == "*"
   actions:
-  - handler: fakeHandlerConfig.fakeHandler.istio-config-default
-    instances: [ reportInstance.samplereport.istio-config-default ]
+  - handler: fakeHandlerConfig.fakeHandler.istio-system
+    instances: [ reportInstance.samplereport.istio-system ]
```
douglas-reid: @mandarjog: while we are changing this, should we adopt the style of:

```
instances:
- reportInstance...
```
instead?

mandarjog: good point. yes, we should.

@jmuk 
1. use the extended yaml format.
2. elide namespace from handler and instance names.

jmuk: done here. Regarding 2, do we want to elide namespaces in other parts, especially in `testdata/config` files?

mandarjog:
This will work now that mixer builds work with Basel build 0.6.1 now.
/lgtm

PR #1406, opened 2017-10-11
Build simplification using API build

kyessenov:
Corresponding update for https://github.com/istio/api/pull/193
```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1406)


BUILD
```diff
@@ -6,7 +6,13 @@ load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_prefix")
 
 go_prefix("istio.io/mixer")
 
-gazelle(name = "gazelle")
+gazelle(
+    name = "gazelle",
+    args = [
+        "-build_file_name",
+        "BUILD,BUILD.bazel",
```
jmuk: Now we don't have files named `BUILD.bazel` anymore?  Also, that would be the default value for gazelle. Why should this be specified explicitly?

WORKSPACE
```diff
@@ -20,10 +32,6 @@ load("@io_bazel_rules_docker//docker:docker.bzl", "docker_repositories")
 
 docker_repositories()
 
-load(":adapter_author_deps.bzl", "mixer_adapter_repositories")
-
-mixer_adapter_repositories()
-
```
jmuk: I think this should remain.

adapter/denier/BUILD
```diff
@@ -12,7 +12,7 @@ go_library(
         "//template/checknothing:go_default_library",
         "//template/listentry:go_default_library",
         "//template/quota:go_default_library",
-        "@com_github_googleapis_googleapis//:google/rpc",
+        "@io_istio_api//google/rpc:go_default_library",
```
jmuk: It looks very weird to me having google/rpc proto under istio.io/api. Is there any way to avoid doing this, like using `google.golang.org/genproto` packages?

BUILD
kyessenov: Default is `BUILD.bazel` which is less popular in this repo.

WORKSPACE
kyessenov: We cannot redefine dependencies after load. These are already defined in API dependencies.

adapter/denier/BUILD
kyessenov: Unfortunately, gogo library of istio api is not build-compatible with genproto version. This would not be a problem if you use go library of istio api.

WORKSPACE
jmuk: maybe @douglas-reid should know the answer -- I think it was added for custom (external) adapters, but still I don't know what it should look like with this PR.

kyessenov:
@sebastienvas can we clean the bazel cache please? it's trying to build stale generated sources.

kyessenov:
cc @guptasu for remote adapters knowledge

kyessenov:
blocked by envoy bazel problems

mandarjog:
/hold 
blocked on envoy bazel problems.

geeknoid:

Thanks.

PR #1408, opened 2017-10-11
fix gcloud build script

mandarjog:
**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1408)

### DIFF
--- a/docker/gcloud_build.sh
+++ b/docker/gcloud_build.sh
@@ -2,6 +2,8 @@
 
 set -ex
 
+PROJECT=${PROJECT:?"gcloud project required"}
+
 gcloud docker --authorize-only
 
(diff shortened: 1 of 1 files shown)


docker/gcloud_build.sh
```diff
@@ -2,6 +2,8 @@
 
 set -ex
 
+PROJECT=${PROJECT:?"gcloud project required"}
```
jmuk: If empty, can we use `gcloud config get-value project` to extract the current default project name from the gcloud config?

mandarjog: OK, fixed it. PTAL.

PR #1409, opened 2017-10-11
Fix the (original) cloudbuilder.yaml

costinm:
The automated build is using the new image - but it's building from tagged branches and using a different process.

This builder can be used with a trigger to generate builds on submit, which can be auto-pushed
or tested manually (keel, etc).

The original builder (cloud-builder) is using the new bazel - we can revert to it when mixer build is 
upgraded, but better to use the istio-builder, for consistency with the 'official' builds.

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1409)


docker/cloudbuild.yaml
```diff
@@ -1,7 +1,9 @@
 steps:
-- name: gcr.io/cloud-builders/bazel
-  args: ['run', '//docker:mixer']
+- name: 'gcr.io/istio-io/istio-builder:0.1'
+  args: ['bazel', 'run', '//docker:mixer']
 - name: gcr.io/cloud-builders/docker
   args: ['tag', 'istio/docker:mixer', 'gcr.io/$PROJECT_ID/mixer:$BUILD_ID']
 images:
-- 'gcr.io/$PROJECT_ID/mixer:$BUILD_ID'
+- 'gcr.io/$PROJECT_ID/mixer'
```
douglas-reid: i've forgotten the `images` semantics... any reason to leave off the tag ?

PR #1410, opened 2017-10-12
Add apicheck template

manlinl:
**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1410)


manlinl:
The template is open. New fields for other form of auth attribute can be added as well. Are you suggesting something as generic as  'map<string, ValueType> dimensions' ? I had that idea but @geeknoid prefers something explicit.

geeknoid:
You have to think of this as an object-oriented design pattern. Is there a way you can design the arguments for a check call to be generic such that completely different adapter implementations can be used?

In this case, perhaps all it means is that api_key should in fact be "credential" or something. And maybe api_operation just becomes "operation" so it's not directly associated with the API management thing.

manlinl:
Template is in review, for now, close this PR.

PR #1422, opened 2017-10-13
Add check_bazel_version

mandarjog:
Ensures that bazel version is min <= version <= max.

If bazel version is not correct it produces output like
```
	File "/Users/mjog/GOHOME/src/istio.io/mixer/check_bazel_version.bzl", line 39, in check_version
		fail("\nCurrent Bazel version {} is t...))
Current Bazel version 0.6.0 is too new, expected at most 0.5.4
```
```
	File "/Users/mjog/GOHOME/src/istio.io/mixer/check_bazel_version.bzl", line 39, in check_version
		fail("\nCurrent Bazel version {} is t...))
Current Bazel version 0.5.2 is too old, expected at least 0.5.3
```
We would like to move this to a common repo so version can be checked uniformly.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1422)

### DIFF
--- a/check_bazel_version.bzl
+++ b/check_bazel_version.bzl
@@ -9,7 +9,7 @@ def _parse_bazel_version(bazel_version):
     # Turn "release" into a tuple of strings
     version_tuple = ()
     for number in parts[0].split("."):
-        version_tuple += (str(number),)
+        version_tuple += (int(number),)
     return version_tuple
 
 # acceptable min_version <= version <= max_version
--- a/check_bazel_version.bzl
+++ b/check_bazel_version.bzl
@@ -1,0 +1,39 @@
+def _parse_bazel_version(bazel_version):
+    # Remove commit from version.
+    version = bazel_version.split(" ", 1)[0]
+
+    # Split into (release, date) parts and only return the release
+    # as a tuple of integers.
+    parts = version.split("-", 1)
+
+    # Turn "release" into a tuple of strings
+    version_tuple = ()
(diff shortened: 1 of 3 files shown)


check_bazel_version_test.py
```diff
@@ -0,0 +1,70 @@
...
+    global fail
+    ok = True
+    import shutil
+    import os
+
+    shutil.copyfile("check_bazel_version.bzl", "check_bazel_version.py")
+
+    import check_bazel_version as cv
+
+    ts = [TestCase("0.5.3", "0.5.4", "0.6.0", "too new"),
```
ldemailly: 0.11.0 may be too old instead of too new?

check_bazel_version.bzl
```diff
@@ -0,0 +1,39 @@
...
+
+    # Turn "release" into a tuple of strings
+    version_tuple = ()
+    for number in parts[0].split("."):
+        version_tuple += (int(number),)
+    return version_tuple
+
+# acceptable min_version <= version <= max_version
+def check_version():
+    check_bazel_version("0.5.3", "0.5.4")
```
ldemailly: should in theory be 0.5.3 - 0.5.999 no ? (I mean if 0.5.5 was to come out it should work)

kyessenov:
Can you move this to test-infra instead?

mandarjog:
@kyessenov I will move it to some other place once we settle the repo story.

PR #1366, opened 2017-09-29
upgrade rules_go to work with bazel 0.6.0

jmuk:
fixes #1365 

- WORKSPACE; some dependencies are changed slightly, which forces this change. Especially, `go_repositories()` will import protobuf and x/net libraries if not yet imported, so their imports need to come earlier than go_repositories.
- `library` parameter in `go_library` rule seems gone, adapter and template codegen need to be fixed because of this
- fix test/testenv since otherwise build fails.

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1366)

PR #1423, opened 2017-10-13
BUILD and WORKSPACE fixes for bazel 0.6.1 

### LINKED ISSUE #1365: build broken with bazel 0.6.0
I've gotten build errors as follows:

```
% bazel build //...
ERROR: /usr/local/google/home/mukai/.cache/bazel/_bazel_mukai/23d88ec2f9449b2c9d846b080b7cc444/external/io_bazel_rules_go/go/private/binary.bzl:226:48: name 'lib' is not defined (did you mean 'libs'?)
ERROR: error loading package '': Extension 'go/private/binary.bzl' has errors
ERROR: error loading package '': Extension 'go/private/binary.bzl' has errors
INFO: Elapsed time: 0.132s
FAILED: Build did NOT complete successfully (0 packages loaded)
```

Seems this is because of bazel 0.6.0?

I had another environment which still stays on bazel 0.5.4. Succeeded to build, but started reporting the same error after upgraded to 0.6.0.

lizan:
- Update rules_go to latest
- Cleaning up deps from WORKSPACE which are built-in in rules_go
- Gazelify all BUILD files
- Fix test/repositories.bzl for proxy build

Based on #1366, fixes #1365.

**Release note**:

```release-note
None
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1423)


jmuk:
This looks good, but I'm a bit worried about having bunch of "importpath=" line in every build target.

Actually mixer builds successfully without them, but gazelle can't tell if they are really necessary or not and it adds these lines regardless the necessity (there's a TODO on gazelle side, see https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/rules/generator.go#L155). Considering this situation, my PR added a nasty sed line to exclude them.

I'm not sure which is appealing to other mixer devs, I want to follow others.

jmuk:
ah, by the way, this implicitly upgrades go version to 1.9.1, right? We haven't discussed this much -- personally I'm totally fine with the upgrade since it builds successfully, but reviewers would need to be aware of it (go_version can be specified in `go_register_toolchain` if we really don't want the upgrade).

kyessenov:
Yes, @lizan please pin go version to 1.8.3 so we can explicitly pick a version. rules_go has all the toolchains available. 

I don't mind adding the silly sed to strip importpath to minimize the impact. I think gazelle is likely to remove that in some future form.

adapter_author_deps.bzl
```diff
@@ -4,24 +4,18 @@ def mixer_adapter_repositories():
 
     native.git_repository(
         name = "org_pubref_rules_protobuf",
-        commit = "eafd42ce6471ce3ea265729c85e18e6180dea620",  # Sept 22, 2017 (genfiles path calculation fix)
+        commit = "ff3b7e7963daa7cb3b42f8936bc11eda4b960926",  # Sept 22, 2017 (genfiles path calculation fix)
```
kyessenov: update the date?

lizan:
Sure, golang upgrade shouldn't be implied by this PR, pinned to 1.8.3. Also revert BUILD files and re-gazellified with `importpath` sed.

jmuk:
@lizan, maybe you need to update io_bazel_rules_docker dependency too for the presubmit failure.

jmuk:
/lgtm

The smoketest failure seems happening because `istio/istio` repository still refers to the old rules_go...

mandarjog:
@lizan please rebase the PR.
It will force you to change allowable bazel versions.
0.5.3 - 0.6.1 seems to work.

PR #1460, opened 2017-10-19
Alter Preprocess bag handling to prevent attribute overriding

douglas-reid:
Prevent APAs from overriding attributes sent in the original request.

This adds a new method to `MutableBag` named `PreserveMerge` which takes in bags that possibly contain conflicts and does a merge in which original values are maintained and conflicts are silently dropped on the floor.

This is in response to errors reported on the dev mailing list that stemmed from having *multiple* values for the attribute `destination.service` in the attribute bag that was used to create instances.

**Release note**:

```release-note
Prefer received attribute values to preprocess-derived attribute values.
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1460)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -11,10 +11,13 @@ git_repository(
 )
 
 load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")
+
 go_rules_dependencies()
+
 go_register_toolchains()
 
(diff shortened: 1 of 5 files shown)


douglas-reid:
PTAL. Updated code to address linter issues. This required a small set of changes around error-handling, but I think that is OK given initial feedback.

geeknoid:
Groovy.

/lgtm

PR #1462, opened 2017-10-19
remove the check of max version of bazel

### LINKED ISSUE #1461: bazel 0.7.0 is released, said "too new,  expected at most 0.6.1"
bazel 0.7.0 is now ready, our check script says about at most 0.7.0. It can build actually though.

It is a bit painful to see this message every time a new version of bazel is released; I think the restriction is a bit too strict.
Maybe we can remove the max_version check for now -- the check should be recovered when we find something is found broken on a specific version of bazel.

jmuk:
Currently there have been no reports for newer version of bazel which
can't build mixer since the check_bazel_version.bzl was created.

The max_version restriction is still helpful, but I think we should
allow no-limit for max_version, and that should be the default
currently.

The max_version can be easily added later again, when a new version
of bazel turned out not working with the current mixer code base.

fixes #1461

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1462)

### DIFF
--- a/check_bazel_version.bzl
+++ b/check_bazel_version.bzl
@@ -14,9 +14,10 @@ def _parse_bazel_version(bazel_version):
 
 # acceptable min_version <= version <= max_version
 def check_version():
-    check_bazel_version("0.5.4", "0.6.1")
+    check_bazel_version("0.5.4", None)
 
-# acceptable min_version <= version <= max_version
+# acceptable min_version <= version <= max_version. max_version check will be omitted
+# when None is specified.
 def check_bazel_version(min_version, max_version):
     if "bazel_version" not in dir(native):
         fail("\nCurrent Bazel version is lower than 0.2.1, expected at least %s\n" %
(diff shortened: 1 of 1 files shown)


jmuk:
My corp workstation automatically updates bazel version, I already got 0.7.0 and see no problems except for this version check. The breakage related to rules_go / bazel 0.5.2 / 0.6.0 update was indeed hard, but I believe it rarely happens. I really don't know why we need to pin the bazel version on all of the work environment.

Also, this PR is not completely removing the max-version check; it's still there and can be set easily when something gets broken. I'm saying that we can reasonably say that some versions would be safe.

If you're not convinced, I'd suggest to relax the max version like `<0.8`, assuming patch-level updates won't break existing builds (or `<=1.0`?)

jmuk:
closing, in favor of https://github.com/istio/mixer/pull/1467

Issue #1503, opened 2017-10-23
//pkg/il/evaluator:go_default_test timeouts during pre-push


apicl:
Error message:

//pkg/il/evaluator:go_default_test                                      TIMEOUT in 60.1s

ozevren:
Is there a link to the failing test? evaluator tests run at ~9secs on my laptop. 60 sec timeout implies more than 6x time dilation.

apicl:
https://github.com/istio/mixer/issues/1503

If I "bazel test" locally, it works fine. It failed during the pre-push
phase.



Thanks
Chengliang

apicl:
Fixed in https://github.com/istio/mixer/commit/08f4f364d914684ece832c7d9635b81dba0f702f

PR #1505, opened 2017-10-24
Add auth attributes and change test size for go_default_test

apicl:
#1503 
**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1505)

### DIFF
--- a/pkg/il/evaluator/BUILD
+++ b/pkg/il/evaluator/BUILD
@@ -21,7 +21,7 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    size = "small",
+    size = "medium",
     srcs = ["evaluator_test.go"],
     library = ":go_default_library",
     deps = [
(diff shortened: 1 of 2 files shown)


pkg/il/evaluator/BUILD
```diff
@@ -21,7 +21,7 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    size = "small",
+    size = "medium",
```
geeknoid: Why this change?

geeknoid:
@douglas-reid @sebastienvas 

What happens when we set a test size to medium? Will the lab still run these tests as part of precommit checks?

apicl:
It still runs as part of the pre-push. The only different is that the pre-push will wait a little longer.

apicl:
PTAL

douglas-reid:
@geeknoid `small` -> `medium` shouldn't have any observable impact. moving to `large` or bigger will, based on our configuration, which includes:

```
test --test_size_filters=-large,-enormous

PR #1515, opened 2017-10-26
Update dependency to fix goroutine proliferation in zipkin-go

mandarjog:
This PR fixes goroutine proliferation issue with openzipkin-go.
https://github.com/istio/issues/issues/100#issuecomment-339468537

After the PR is upstreamed, we will update the dependency.

https://github.com/mandarjog/zipkin-go-opentracing/pull/1

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1515)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -465,7 +465,7 @@ go_repository(
 go_repository(
     name = "com_github_apache_thrift",
     build_file_name = "BUILD.bazel",
-    commit = "b2a4d4ae21c789b689dd162deb819665567f481c",  # Jan 6, 2017 (0.10.0)
+    commit = "d4df91709b724174aaf8a957f3edac3573be354e",  # Oct 26, 2017 (HEAD) required by openzipkin dependency.
     importpath = "github.com/apache/thrift",
 )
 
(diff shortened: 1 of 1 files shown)


jmuk:
collected some data points for this PR -- in short, this seems solving the reported problem (istio/issues#100).

My setup is to simply deploy the istio-auth (without zipkin pod), bookinfo sample, and run `wrk` to the productpage for 5 minutes.

The red line in the following graph shows the heap alloc bytes data (tracked by prometheus), with this PR (green is proxy -- unrelated, same in the following graphs):
screenshot 2017-10-26 at 14 17 39

Without this PR, typically heap allocation looks like
screenshot 2017-10-26 at 14 18 00

note that the spiky pattern happens after wrk run finishes -- still using memory for zipkin, we guess.

The number of goroutines, 
screenshot 2017-10-26 at 14 06 38

Still have some increases since gRPC creates a goroutine for every request, but actually, the goroutine count looks like as follows without this PR
screenshot 2017-10-26 at 14 18 56

and note that, with the scale of the latter graph, the former goroutine "increase" looks like a flat line.

PR #1523, opened 2017-10-27
Make Result.Interface() return an empty string, as is.

ozevren:
+ Removing the old handling code that converts an empty string to
nil.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/istio/mixer/1523)

### DIFF
--- a/pkg/il/compiler/compiler_test.go
+++ b/pkg/il/compiler/compiler_test.go
@@ -806,6 +806,13 @@ end`,
 	{
 		expr: `ar["c"] | "foo"`,
 		input: map[string]interface{}{
+			"ar": map[string]string{},
+		},
+		result: "foo",
+	},
+	{
+		expr: `ar["c"] | "foo"`,
+		input: map[string]interface{}{
 			"ar": map[string]string{"c": "b"},
 		},
 		result: "b",
--- a/pkg/il/compiler/compiler_test.go
+++ b/pkg/il/compiler/compiler_test.go
@@ -388,6 +388,20 @@ end
 `,
 	},
 	{
+		expr: `ar["b"]`,
+		input: map[string]interface{}{
+			"ar": map[string]string{},
+		},
+		result: "",
+		code: `
+fn eval() string
+  resolve_f "ar"
+  anlookup "b"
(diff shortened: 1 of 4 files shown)


ozevren:
Fixes https://github.com/istio/mixer/issues/934

ozevren:
/test mixer-presubmit

Test passes when run locally.

ozevren:
friendly ping

pkg/il/compiler/compiler_test.go
```diff
@@ -388,6 +388,20 @@ end
 `,
 	},
 	{
+		expr: `ar["b"]`,
```
douglas-reid, line 391: won't this lead to the issue that originally caused us to make `""` => `nil` ?  what happens for: `ar["b"] | "other"`?

ozevren: Depending on the bag contents, ar["b"] | "other" will evaluate to:
bag {} => "other"
bag { ar:{} } => "other"
bag { ar:{ "b": ""} } => ""

guptasu (approved):
LGTM

pkg/il/compiler/compiler_test.go
douglas-reid: Do we have tests for those cases? If not, can we add them?

douglas-reid:
with the repo migration in-flight, we shouldn't merge any PRs at the moment.

pkg/il/compiler/compiler_test.go
ozevren: Yes we do. I've also added one more, just in case, in an additional changelist within the same PR.

ozevren:
What is the ETA for repo migration completion?

douglas-reid:
@ozevren i've been led to believe that it will be Wed. Nov. 1.  Any PRs submitted now would be lost in the shuffle. Unfortunately, this means you will have to move these changes over and resubmit (with hopefully a successful transfer of test infra, etc., already done and working). :(.

geeknoid:

Thanks.

